PR #8, opened 2014-12-12
SSHD-388 - SSH server fails during bundle update

janstey:
Simple fix is the catch the NPE coming from the updating felix framework. The server seems to be able to gracefully handle this condition.


janstey:
@gnodet is this change OK to merge?

PR #10, opened 2014-12-18
[SSHD-393] Fix frozen ssh output stream on window adjust

marosmars:
When server output stream delayed write it did not resume it after the window was adjusted
Resume writes the same way client ssh channel does

maros.mars@gmail.com


tonydamage:
marosmars could you also attach unit test?

marosmars:
The unit tests are already there. I can see them within the commit

gnodet:
Could you please close this PR now ?

marosmars:
Done. Thx
When are you releasing new version of mina-sshd with this fix included ?

gnodet:
Very soon, I'm mainly waiting for a confirmation that SSHD-348 is fixed.

PR #12, opened 2015-09-10
Candidate fix for SSHD-557

sngardner:
Fixes whole-server hang by removing blocking waitForSpace() call and replacing it with getSize(). In the case that there is no space available, flush() will not block any more and will simply try again - a zero-length buffer will be passed to waitAndConsume(), but that method will immediately return.


lgoldstein:
See my comment for the SSHD-557 issue

gnodet:
Raised https://issues.apache.org/jira/browse/SSHD-1003

PR #13, opened 2015-09-17
Fix configuring TCP_NODELAY

avthart:
The wrong property is currently used for configuring TCP nodelay. See diff.


lgoldstein:
Well caught - merged in (can't close this pull request because I don't have write access...)

PR #15, opened 2015-11-18
[SSHD-589] Automatic detect dh key size on top of [SSHD-589] initial

alonbl:


alonbl:
@lgoldstein I added automatic detection on top of your patch, please consider to accept.

sshd-core/src/main/java/org/apache/sshd/common/util/SecurityUtils.java
```diff
@@ -89,6 +110,60 @@ public static boolean hasEcc() {
...
+                for (int s = 1024; s <= 8192 && isDHGroupExchangeSupported(s); s += 1024) {
+                    maxSupportedKeySize = s;
+                }
+            }
+
+            MAX_DHG_KEY_SIZE_HOLDER.set(maxSupportedKeySize);
+        }
+
+        return maxSupportedKeySize;
+    }
```
lgoldstein: I like it, but here is how I would write it - please let me know if you agree:
```java
    public static final int MIN_DHGEX_KEY_SIZE = 1024;
    public static final int MAX_DHGEX_KEY_SIZE = 8192;

    public static int getMaxDHGroupExchangeKeySize() {
        int maxSupportedKeySize;
        synchronized (MAX_DHG_KEY_SIZE_HOLDER) {
            maxSupportedKeySize = MAX_DHG_KEY_SIZE_HOLDER.get();
            if (maxSupportedKeySize != 0) { // 1st time we are called ?
                return maxSupportedKeySize;
            }

            String propValue = System.getProperty(MAX_DHGEX_KEY_SIZE_PROP);
            if (GenericUtils.isEmpty(propValue)) {
                // Go down from max. to min. to ensure we stop at 1st maximum value success
                for (int testKeySize = MAX_DHGEX_KEY_SIZE; testKeySize >= MIN_DHGEX_KEY_SIZE; testKeySize -= 1024) {
                    if (isDHGroupExchangeSupported(testKeySize)) {
                        maxSupportedKeySize = testKeySize;
                        break;
                    }
                }
            } else {
                maxSupportedKeySize = Integer.parseInt(propValue);
                ValidateUtils.checkTrue(maxSupportedKeySize != 0,   // -1 is OK - means user wants to disable DH group exchange
                        "Configured " + MAX_DHGEX_KEY_SIZE_PROP + " value must be non-zero: %d", maxSupportedKeySize);
            }

            MAX_DHG_KEY_SIZE_HOLDER.set(maxSupportedKeySize);
        }

        return maxSupportedKeySize;
    }
```

alonbl: sure I agree! :)

lgoldstein: Then I will make the change, run it through the builder and if it is successful will merge into the master branch and let you know so you can close the pull request (we do not merge from them only use them for code discussions)

alonbl:
Last question... BuiltinDHFactories#dhg14, still has:
```
        @Override
        public boolean isSupported() {
            return SecurityUtils.isBouncyCastleRegistered();
        }
```
Is this required now, or can also use SecurityUtils.isDHGroupExchangeSupported()?

alonbl:
While we at it... BuiltinSignatures has several of:
```
        @Override
        public boolean isSupported() {
            return SecurityUtils.isBouncyCastleRegistered() || SecurityUtils.hasEcc();
        }
```
Won't SecurityUtils.hasEcc() will return true if BouncyCastle is registered anyway?

These are the last BouncyCastle specific code that relates to crypto.

lgoldstein:
I am aware of that already for some time - there are some scenarios that I am looking into where it may make a subtle difference

alonbl:
Merged

PR #17, opened 2015-11-19
[SSHD-584] Closer to OpenSSH file permissions checks

alonbl:
1. Owner can be either running user or root in *NIX.
2. Config file have less restrictive requirements than keys.

I can probably remove duplicate code, but thought first to get some feedback.

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/client/config/hosts/DefaultConfigFileHostEntryResolver.java
+++ b/sshd-core/src/main/java/org/apache/sshd/client/config/hosts/DefaultConfigFileHostEntryResolver.java
@@ -23,10 +23,9 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.file.LinkOption;
 import java.nio.file.Path;
-import java.nio.file.attribute.PosixFilePermission;
 import java.util.List;
 
-import org.apache.sshd.common.config.keys.KeyUtils;
+import org.apache.sshd.common.util.Pair;
 import org.apache.sshd.common.util.ValidateUtils;
 import org.apache.sshd.common.util.io.IoUtils;
 
@@ -75,12 +74,13 @@ public class DefaultConfigFileHostEntryResolver extends ConfigFileHostEntryResol
                 log.debug("reloadHostConfigEntries({}@{}:{}) check permissions of {}", username, host, port, path);
             }
 
-            PosixFilePermission violation = KeyUtils.validateStrictKeyFilePermissions(path);
+            Pair<String, Object> violation = validateStrictConfigFilePermissions(path);
             if (violation != null) {
-                throw new IOException("String permission violation (" + violation + ") for " + path);
+                throw new IOException(String.format("%s for %s", violation.getFirst(), path));
             }
         }
 
         return super.reloadHostConfigEntries(path, host, port, username);
     }
+
 }
--- a/sshd-core/src/main/java/org/apache/sshd/client/config/keys/ClientIdentity.java
+++ b/sshd-core/src/main/java/org/apache/sshd/client/config/keys/ClientIdentity.java
@@ -23,7 +23,6 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.LinkOption;
 import java.nio.file.Path;
-import java.nio.file.attribute.PosixFilePermission;
 import java.security.GeneralSecurityException;
 import java.security.KeyPair;
 import java.util.Collection;
@@ -293,6 +292,7 @@ public final class ClientIdentity {
      * <U>insensitive</U>), value=the {@link Path} of the file holding the key
      * @throws IOException If failed to access the file system
      * @see KeyUtils#validateStrictKeyFilePermissions(Path, LinkOption...)
+     * @see KeyUtils#validateStrictKeyFileOwner(Path, LinkOption...)
      */
     public static Map<String, Path> scanIdentitiesFolder(
             Path dir, boolean strict, Collection<String> types, Transformer<String, String> idGenerator, LinkOption... options)
@@ -316,8 +316,7 @@ public final class ClientIdentity {
             }
 
             if (strict) {
-                PosixFilePermission perm = KeyUtils.validateStrictKeyFilePermissions(p, options);
-                if (perm != null) {
+                if (KeyUtils.validateStrictKeyFilePermissions(p, options) != null) {
                     continue;
                 }
             }
--- a/sshd-core/src/main/java/org/apache/sshd/client/config/keys/ClientIdentityFileWatcher.java
+++ b/sshd-core/src/main/java/org/apache/sshd/client/config/keys/ClientIdentityFileWatcher.java
@@ -21,13 +21,13 @@ package org.apache.sshd.client.config.keys;
 
 import java.io.IOException;
 import java.nio.file.Path;
-import java.nio.file.attribute.PosixFilePermission;
 import java.security.GeneralSecurityException;
 import java.security.KeyPair;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.sshd.common.config.keys.FilePasswordProvider;
 import org.apache.sshd.common.config.keys.KeyUtils;
+import org.apache.sshd.common.util.Pair;
 import org.apache.sshd.common.util.ValidateUtils;
 import org.apache.sshd.common.util.io.IoUtils;
 import org.apache.sshd.common.util.io.ModifiableFileWatcher;
@@ -91,10 +91,10 @@ public class ClientIdentityFileWatcher extends ModifiableFileWatcher implements 
 
     protected KeyPair reloadClientIdentity(Path path) throws IOException, GeneralSecurityException {
         if (isStrict()) {
-            PosixFilePermission violation = KeyUtils.validateStrictKeyFilePermissions(path, IoUtils.EMPTY_LINK_OPTIONS);
+            Pair<String, Object> violation = KeyUtils.validateStrictKeyFilePermissions(path, IoUtils.EMPTY_LINK_OPTIONS);
             if (violation != null) {
                 if (log.isDebugEnabled()) {
-                    log.debug("reloadClientIdentity({}) ignore due to permission violation: {}", path, violation);
+                    log.debug("reloadClientIdentity({}) ignore due to {}", path, violation.getFirst());
                 }
                 return null;
             }
--- a/sshd-core/src/main/java/org/apache/sshd/common/config/keys/KeyUtils.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/config/keys/KeyUtils.java
@@ -47,6 +47,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
@@ -61,6 +62,7 @@ import org.apache.sshd.common.digest.DigestUtils;
 import org.apache.sshd.common.keyprovider.KeyPairProvider;
 import org.apache.sshd.common.util.GenericUtils;
 import org.apache.sshd.common.util.OsUtils;
+import org.apache.sshd.common.util.Pair;
 import org.apache.sshd.common.util.ValidateUtils;
 import org.apache.sshd.common.util.buffer.Buffer;
 import org.apache.sshd.common.util.buffer.ByteArrayBuffer;
@@ -123,17 +125,32 @@ public final class KeyUtils {
      * (For {@code Unix}) If the path is a file, then its folder may not have
      * group or others permissions
      * </P></LI>
+     *
+     * <LI><P>
+     * The path must be owned by current user.
+     * </P></LI>
+     *
+     * <LI><P>
+     * (For {@code Unix}) The path may be owned by root.
+     * </P></LI>
+     *
+     * <LI><P>
+     * (For {@code Unix}) If the path is a file, then its folder must also
+     * have valid owner.
+     * </P></LI>
+     *
      * </UL>
      *
      * @param path    The {@link Path} to be checked - ignored if {@code null}
      *                or does not exist
      * @param options The {@link LinkOption}s to use to query the file's permissions
-     * @return The violated {@link PosixFilePermission} - {@code null} if
+     * @return The violated permission as {@link Pair} first is a message second is
+     * offending object {@link PosixFilePermission} or {@link String} for owner - {@code null} if
      * no violations detected
      * @throws IOException If failed to retrieve the permissions
      * @see #STRICTLY_PROHIBITED_FILE_PERMISSION
      */
-    public static PosixFilePermission validateStrictKeyFilePermissions(Path path, LinkOption... options) throws IOException {
+    public static Pair<String, Object> validateStrictKeyFilePermissions(Path path, LinkOption... options) throws IOException {
         if ((path == null) || (!Files.exists(path, options))) {
             return null;
         }
@@ -144,20 +161,52 @@ public final class KeyUtils {
         }
 
         if (perms.contains(PosixFilePermission.OTHERS_EXECUTE)) {
-            return PosixFilePermission.OTHERS_EXECUTE;
+            PosixFilePermission p = PosixFilePermission.OTHERS_EXECUTE;
+            return new Pair(String.format("Permissions violation (%s)", p), p);
         }
 
         if (OsUtils.isUNIX()) {
             PosixFilePermission p = IoUtils.validateExcludedPermissions(perms, STRICTLY_PROHIBITED_FILE_PERMISSION);
             if (p != null) {
-                return p;
+                return new Pair(String.format("Permissions violation (%s)", p), p);
             }
 
             if (Files.isRegularFile(path, options)) {
                 Path parent = path.getParent();
                 p = IoUtils.validateExcludedPermissions(IoUtils.getPermissions(parent, options), STRICTLY_PROHIBITED_FILE_PERMISSION);
                 if (p != null) {
-                    return p;
+                    return new Pair(String.format("Parent permissions violation (%s)", p), p);
+                }
+            }
+        }
+
+        String current = OsUtils.getCurrentUser();
+        String owner = IoUtils.getFileOwner(path, options);
+
+        if (GenericUtils.isEmpty(owner)) {
+            // we cannot get owner
+            // general issue: jvm does not support permissions
+            // security issue: specific filesystem does not support permissions
+            return null;
+        }
+
+        Set<String> expected = new HashSet<>();
+        expected.add(current);
+        if (OsUtils.isUNIX()) {
+            // Windows Administrator was considered however
+            // in Windows most liklely a group is in used.
+            expected.add(OsUtils.ROOT_USER);
+        }
+
+        if (!expected.contains(owner)) {
+            return new Pair(String.format("Owner violation (%s)", owner), owner);
+        }
+
+        if (OsUtils.isUNIX()) {
+            if (Files.isRegularFile(path, options)) {
+                String parentOwner = IoUtils.getFileOwner(path.getParent(), options);
+                if (!expected.contains(parentOwner)) {
+                    return new Pair(String.format("Parent owner violation (%s)", parentOwner), parentOwner);
                 }
             }
         }
--- a/sshd-core/src/main/java/org/apache/sshd/common/util/OsUtils.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/util/OsUtils.java
@@ -21,6 +21,7 @@ package org.apache.sshd.common.util;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * Operating system dependent utility methods.
@@ -28,14 +29,24 @@ import java.util.List;
  * @author <a href="mailto:dev@mina.apache.org">Apache MINA SSHD Project</a>
  */
 public final class OsUtils {
+
+    /**
+     * Property that can be used to override the reported value from {@link #getCurrentUser()}
+     * if not set then &quot;user.name&quot; system property is used
+     */
+    public static final String CURRENT_USER_OVERRIDE_PROP = "org.apache.sshd.currentUser";
+
     public static final String WINDOWS_SHELL_COMMAND_NAME = "cmd.exe";
     public static final String LINUX_SHELL_COMMAND_NAME = "/bin/sh";
+
+    public static final String ROOT_USER = "root";
 
     public static final List<String> LINUX_COMMAND =
             Collections.unmodifiableList(Arrays.asList(LINUX_SHELL_COMMAND_NAME, "-i", "-l"));
     public static final List<String> WINDOWS_COMMAND =
             Collections.unmodifiableList(Collections.singletonList(WINDOWS_SHELL_COMMAND_NAME));
 
+    private static final AtomicReference<String> CURRENT_USER_HOLDER = new AtomicReference<>(null);
     private static final boolean WIN32 = GenericUtils.trimToEmpty(System.getProperty("os.name")).toLowerCase().contains("windows");
 
     private OsUtils() {
@@ -65,6 +76,35 @@ public final class OsUtils {
             return WINDOWS_COMMAND;
         } else {
             return LINUX_COMMAND;
+        }
+    }
+
+    /**
+     * Get current user name.
+     * @return current user.
+     */
+    public static String getCurrentUser() {
+        String username = null;
+        synchronized (CURRENT_USER_HOLDER) {
+            username = CURRENT_USER_HOLDER.get();
+            if (username != null) {  // have we already resolved it ?
+                return username;
+            }
+
+            username = System.getProperty(CURRENT_USER_OVERRIDE_PROP, System.getProperty("user.name"));
+            ValidateUtils.checkNotNullAndNotEmpty(username, "No username available");
+            CURRENT_USER_HOLDER.set(username);
+        }
+
+        return username;
+    }
+
+    /**
+     * Can be used to programmatically set the username reported by {@link #getCurrentUser()}
+     */
+    public static void setCurrentUser(String username) {
+        synchronized (CURRENT_USER_HOLDER) {    // not really necessary, but let's be consistent
+            CURRENT_USER_HOLDER.set(username);
         }
     }
 
--- a/sshd-core/src/main/java/org/apache/sshd/common/util/io/IoUtils.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/util/io/IoUtils.java
@@ -32,6 +32,7 @@ import java.nio.file.LinkOption;
 import java.nio.file.OpenOption;
 import java.nio.file.Path;
 import java.nio.file.attribute.PosixFilePermission;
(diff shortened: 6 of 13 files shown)


sshd-core/src/main/java/org/apache/sshd/client/config/hosts/DefaultConfigFileHostEntryResolver.java
```diff
@@ -75,12 +90,105 @@ public final boolean isStrict() {
...
                 throw new IOException("String permission violation (" + violation + ") for " + path);
             }
+
+            String ownerViolation = validateStrictConfigFileOwner(path);
+            if (ownerViolation != null) {
+                throw new IOException("String owner violation (" + ownerViolation + ") for " + path);
+            }
         }
 
         return super.reloadHostConfigEntries(path, host, port, username);
```
lgoldstein: I prefer:
```java
if (GenericUtils.length(owner) > 0) {
    throw ...
}
```
The assumption is that there is not such thing as "empty" owner - and if there is, then we assume no violation since we cannot determine the user.

sshd-core/src/main/java/org/apache/sshd/client/config/hosts/DefaultConfigFileHostEntryResolver.java
```diff
@@ -75,12 +90,105 @@ public final boolean isStrict() {
...
+        }
+
+        Set<String> expected = new HashSet<>();
+        if (current != null) {
+            expected.add(current);
+        }
+        if (OsUtils.isUNIX()) {
+            expected.add(OsUtils.ROOT_USER);
+        }
+
```
lgoldstein: Maybe for _Windows_ we should also allow _administrator_

sshd-core/src/main/java/org/apache/sshd/client/config/hosts/DefaultConfigFileHostEntryResolver.java
```diff
@@ -75,12 +90,105 @@ public final boolean isStrict() {
...
+        String current = IoUtils.getCurrentUser();
+        String owner = IoUtils.getFileOwner(path, options);
+
+        if (current == null &&  owner == null) {
+            // we cannot detect permissions
+            return null;
+        }
+
+        Set<String> expected = new HashSet<>();
```
lgoldstein: Use:
```java
Set<String> expected = new TreeSet<>(String.CASE_INSENSITITVE_ORDER);
```
Personally, I think that anyone using "ROOT" and "root" as distinct users (for Linux) or "administrator" and "ADMINISTRATOR" (Windows) deserve all they get. However, perhaps this may be better:
```java
Set<String> expected = OsUtils.isWin32() ? new TreeSet<>(String.CASE_INSENSITITVE_ORDER) : new HashSet<>();
```

sshd-core/src/main/java/org/apache/sshd/client/config/keys/ClientIdentity.java
```diff
@@ -320,6 +321,10 @@ public static KeyPairProvider loadDefaultKeyPairProvider(
                 if (perm != null) {
                     continue;
                 }
+                String owner = KeyUtils.validateStrictKeyFileOwner(p, options);
+                if (owner != null) {
+                    continue;
+                }
             }
```
lgoldstein: I prefer:
```java
if (GenericUtils.length(owner) > 0) {
    continue;
}
```
I don't think "empty" owner are expected, and even if they are IMO they are the same as "no user".

sshd-core/src/main/java/org/apache/sshd/client/config/keys/ClientIdentityFileWatcher.java
```diff
@@ -98,6 +98,13 @@ protected KeyPair reloadClientIdentity(Path path) throws IOException, GeneralSec
                 }
                 return null;
             }
+            String ownerViolation = KeyUtils.validateStrictKeyFileOwner(path, IoUtils.EMPTY_LINK_OPTIONS);
+            if (ownerViolation != null) {
+                if (log.isDebugEnabled()) {
+                    log.debug("reloadClientIdentity({}) ignore due to owner violation: {}", path, ownerViolation);
+                }
```
lgoldstein: I prefer:
```java
if (GenericUtils.length(ownerViolation) > 0) {
    ...
}
```
I don't think "empty" owner are expected, and even if they are IMO they are the same as "no user".

sshd-core/src/main/java/org/apache/sshd/common/config/keys/KeyUtils.java
```diff
@@ -166,6 +167,69 @@ public static PosixFilePermission validateStrictKeyFilePermissions(Path path, Li
...
+        }
+
+        String current = IoUtils.getCurrentUser();
+        String owner = IoUtils.getFileOwner(path, options);
+
+        if (current == null &&  owner == null) {
+            // we cannot detect permissions
+            return null;
+        }
+
```
lgoldstein: See my previous remark regarding using _GenericUtils.isEmpty_

sshd-core/src/main/java/org/apache/sshd/common/config/keys/KeyUtils.java
```diff
@@ -166,6 +167,69 @@ public static PosixFilePermission validateStrictKeyFilePermissions(Path path, Li
...
+        Set<String> expected = new HashSet<>();
```
lgoldstein: See my previous remark regarding using a _TreeSet_

sshd-core/src/main/java/org/apache/sshd/common/config/keys/KeyUtils.java
```diff
@@ -166,6 +167,69 @@ public static PosixFilePermission validateStrictKeyFilePermissions(Path path, Li
...
+        if (current != null) {
+            expected.add(current);
+        }
+        if (OsUtils.isUNIX()) {
+            expected.add(OsUtils.ROOT_USER);
+        }
```
lgoldstein: See my previous remark about the _Windows_ "administrator_

sshd-core/src/main/java/org/apache/sshd/common/config/keys/KeyUtils.java
```diff
@@ -166,6 +167,69 @@ public static PosixFilePermission validateStrictKeyFilePermissions(Path path, Li
...
+            return owner;
+        }
+
+        if (OsUtils.isUNIX()) {
+            if (Files.isRegularFile(path, options)) {
+                String parentOwner = IoUtils.getFileOwner(path.getParent(), options);
+                if (!expected.contains(parentOwner)) {
+                    return parentOwner;
+                }
+            }
```
lgoldstein: I feel better with:
```java
String parentOwner = IoUtils.getFileOwner(path.getParent(), options);
if ((GenericUtils.length(parentOwner) > 0) && (!expected.contains(parentOwner))) {
    return parentOwner;
}
```

sshd-core/src/main/java/org/apache/sshd/common/util/OsUtils.java
```diff
@@ -31,6 +31,8 @@
     public static final String WINDOWS_SHELL_COMMAND_NAME = "cmd.exe";
     public static final String LINUX_SHELL_COMMAND_NAME = "/bin/sh";
 
+    public static final String ROOT_USER = "root";
+
     public static final List<String> LINUX_COMMAND =
```
lgoldstein: Perhaps should also add:
```java
public static final String WINDOWS_ADMIN_USER = "administrator";
```

sshd-core/src/main/java/org/apache/sshd/common/util/io/IoUtils.java
```diff
@@ -257,6 +260,56 @@ public static void setPermissionsToFile(File f, Collection<PosixFilePermission>
...
+        try {
+            UserPrincipal principal = Files.getOwner(path, options);
+            if (principal != null) {
+                owner = principal.getName();
+            }
+        } catch (UnsupportedOperationException e) {
+            // ignore
+        }
+        return owner;
+    }
```
lgoldstein: Simpler code:
```java
try {
    UserPrincipal principal = Files.getOwner(path, options}
    return (principal != null) ? principal.getName() : null;
} catch (UnsupportedOperationException  e) {
    return null;  // ignore
}

File: sshd-core/src/main/java/org/apache/sshd/common/util/io/IoUtils.java [hunk 12]
```diff
@@ -257,6 +260,56 @@ public static void setPermissionsToFile(File f, Collection<PosixFilePermission>
...
+                        // mark as attempted
+                        owner = "";
+                    }
+                } finally {
+                    if (t != null) {
+                        t.delete();
+                    }
+                }
+                CURRENT_USER_HOLDER.set(owner);
+            }
```
lgoldstein: Why not simply _System.getProperty("user.name")_ ?

sshd-core/src/main/java/org/apache/sshd/server/config/keys/DefaultAuthorizedKeysAuthenticator.java
```diff
@@ -112,6 +112,11 @@ protected boolean isValidUsername(String username, ServerSession session) {
             if (violation != null) {
                 throw new IOException("String permission violation (" + violation + ") for " + path);
             }
+
+            String ownerViolation = KeyUtils.validateStrictKeyFileOwner(path);
+            if (ownerViolation != null) {
+                throw new IOException("String owner violation (" + ownerViolation + ") for " + path);
+            }
         }
```
lgoldstein: See my previous remark regarding _GenericUtils.length(...) > 0_

sshd-core/src/test/java/org/apache/sshd/common/util/io/IoUtilsTest.java
```diff
@@ -54,4 +57,12 @@ public void testGetEOLBytes() {
...
         }
     }
+
+    @Test
+    public void testGetCurrentUser() throws IOException {
+        // system may be unsupportable, hence the null
+        Collection expected = Arrays.asList(null, System.getProperty("user.name"));
+        String user = IoUtils.getCurrentUser();
+        assertTrue(String.format("Incorrect user got from file owner, expected '%s' got %s", expected, user), expected.contains(user));
+    }
```
lgoldstein: No need if we agree that _IoUtils.getCurrentUser()_ returns _System.getProperty("user.name")_

BTW what do you mean by 'system may be unsupportable'

sshd-core/src/main/java/org/apache/sshd/client/config/hosts/DefaultConfigFileHostEntryResolver.java
```diff
@@ -75,12 +90,105 @@ public final boolean isStrict() {
...
+            return null;
+        }
+
+        String current = IoUtils.getCurrentUser();
+        String owner = IoUtils.getFileOwner(path, options);
+
+        if (current == null &&  owner == null) {
+            // we cannot detect permissions
+            return null;
+        }
```
lgoldstein: I prefer:
```java
if (GenericUtils.isEmpty(current) && GenericUser.isEmpty(owner)) {
    return null;   // we cannot detect owners
}
```

sshd-core/src/main/java/org/apache/sshd/client/config/hosts/DefaultConfigFileHostEntryResolver.java
```diff
@@ -75,12 +90,105 @@ public final boolean isStrict() {
...
+
+        Set<String> expected = new HashSet<>();
+        if (current != null) {
+            expected.add(current);
+        }
```
lgoldstein: See my previous remark regarding _GenericUtils.length(...) > 0_

sshd-core/src/main/java/org/apache/sshd/common/util/io/IoUtils.java
```diff
@@ -67,6 +69,7 @@
     private static final byte[] EOL_BYTES = EOL.getBytes(StandardCharsets.UTF_8);
 
     private static final LinkOption[] NO_FOLLOW_OPTIONS = new LinkOption[]{LinkOption.NOFOLLOW_LINKS};
+    private static final AtomicReference<String> CURRENT_USER_HOLDER = new AtomicReference<>();
```
lgoldstein: No need for it if we agree that _System.getProperty("user.name")_ is the value we seek

sshd-core/src/main/java/org/apache/sshd/server/config/keys/DefaultAuthorizedKeysAuthenticator.java
```diff
@@ -112,6 +112,11 @@ protected boolean isValidUsername(String username, ServerSession session) {
             if (violation != null) {
                 throw new IOException("String permission violation (" + violation + ") for " + path);
             }
+
+            String ownerViolation = KeyUtils.validateStrictKeyFileOwner(path);
+            if (ownerViolation != null) {
+                throw new IOException("String owner violation (" + ownerViolation + ") for " + path);
+            }
```
lgoldstein: The messages should be simply "Permission violation(...) for ..." and "Owner violation (...) for ..." - not need for the _String_ work

lgoldstein:
In general looking good (some minor issues) except for:

* Seems like some of the code is duplicate and could be converged into single methods
* The issue of the "current user" value needs to be resolved
* Need to think about the _Windows_ "administrator" user as the _Linux_ "root" equivalent (yes / no, for / against ?).

lgoldstein:
I also think that the new static methods are more natural to reside in _ModifiableFileWatcher_ (which is an ancestor of _DefaultConfigFileHostEntryResolver_) as it usually serves as the base class for the various configuration files watchers, which may also benefit from the same permissions validation code as the _~/.ssh/config_ file

sshd-core/src/main/java/org/apache/sshd/client/config/hosts/DefaultConfigFileHostEntryResolver.java
alonbl: done

sshd-core/src/main/java/org/apache/sshd/client/config/hosts/DefaultConfigFileHostEntryResolver.java
alonbl: I thought about this, but then rejected, as in windows the best practice is to use the administrators group in the most minimalistic configuration, or other groups. So in windows it has no actual sense to check these in java.

lgoldstein: I can live with that - maybe though a remark to this end should be added at this line to let others know that the _Windows_ "administrator" issue was considered here and why we chose not to include it.

sshd-core/src/main/java/org/apache/sshd/client/config/hosts/DefaultConfigFileHostEntryResolver.java
alonbl: I replied about regarding Administrator, I do not think this should be supported.

I also do not think that in *NIX root is case insensitive.

sshd-core/src/main/java/org/apache/sshd/client/config/keys/ClientIdentity.java
alonbl: done

sshd-core/src/main/java/org/apache/sshd/client/config/keys/ClientIdentityFileWatcher.java
alonbl: done

sshd-core/src/main/java/org/apache/sshd/client/config/hosts/DefaultConfigFileHostEntryResolver.java
lgoldstein: Then we are good - I can live with a _HashSet_

sshd-core/src/main/java/org/apache/sshd/common/config/keys/KeyUtils.java
alonbl: done, but I am unsure why null is not the only value to mark this state, but not that important.

sshd-core/src/main/java/org/apache/sshd/common/config/keys/KeyUtils.java
alonbl: did not do this as root is case sensitive and this should not be supported on Windows.

sshd-core/src/main/java/org/apache/sshd/common/config/keys/KeyUtils.java
alonbl: same.

sshd-core/src/main/java/org/apache/sshd/common/config/keys/KeyUtils.java
alonbl: Done.

sshd-core/src/main/java/org/apache/sshd/common/util/OsUtils.java
alonbl: See above notes.

sshd-core/src/main/java/org/apache/sshd/common/util/io/IoUtils.java
alonbl: Done

sshd-core/src/main/java/org/apache/sshd/common/util/io/IoUtils.java
alonbl: delete does not throw IOException.

and I considered using user.name but then there is an issue with mismatch between java way of resolving user via environment and such and how resolve back id may look like.

so I decided to play safe.

if you think that user.name is sufficient I will use it, but I feel like it is insufficient.

sshd-core/src/main/java/org/apache/sshd/server/config/keys/DefaultAuthorizedKeysAuthenticator.java
alonbl: Done

sshd-core/src/test/java/org/apache/sshd/common/util/io/IoUtilsTest.java
alonbl: null is accepted response if system does not support file ownership detection.

sshd-core/src/main/java/org/apache/sshd/client/config/hosts/DefaultConfigFileHostEntryResolver.java
alonbl: so why not use !isEmpty in all other cases to be symmetric?

anyway, done.

sshd-core/src/main/java/org/apache/sshd/client/config/hosts/DefaultConfigFileHostEntryResolver.java
alonbl: Done

sshd-core/src/main/java/org/apache/sshd/common/util/io/IoUtils.java
alonbl: See previous comments.

sshd-core/src/main/java/org/apache/sshd/server/config/keys/DefaultAuthorizedKeysAuthenticator.java
alonbl: I also wondered why there is a string, please push a fix to master and I will rebase.

sshd-core/src/main/java/org/apache/sshd/client/config/hosts/DefaultConfigFileHostEntryResolver.java
alonbl: Done

sshd-core/src/main/java/org/apache/sshd/common/util/io/IoUtils.java
lgoldstein: The let me suggest a compromise:
```java
/**
  * Property that can be used to override the reported value from {@link #getCurrentUser()}
  * if not set then &quot;user.name&quot; system property is used
  */
public static final String CURRENT_USER_OVERRIDE_PROP = "org.apache.sshd.currentUser";
private static final AtomicReference<String> CURRENT_USER_HOLDER = new ...(null);

public static String getCurrentUser() {
    String username = null;
    synchronized (CURRENT_USER_HOLDER) {
        username = CURRENT_USER_HOLDER.get();
        if (username == null) {  // have we already resolved it ?
           return username;
       }

       username = System.getProperty(CURRENT_USER_OVERRIDE_PROP, System.getProperty("user.name"));
       ??? ValidateUtils.checkNotEmpty(username, "No username available"); ???
       CURRENT_USER_HOLDER.set(username);
    }

    return username;
}

// can be used to programmatically set the username reported by {@link #getCurrentUser()}
public static void setCurrentUser(String username) {
    synchronized (CURRENT_USER_HOLDER) {    // not really necessary, but let's be consistent
          CURRENT_USER_HOLDER.set(username);
    }
```

If we agree on this, then you need locate *all* the other places in the code (and there are very few) where "user.name" is used and replace it with "IoUtils.getCurrentUser()". Also, thinking about it, the "get/setCurrentUser" methods should reside in _OsUtils_

alonbl: ok, if you accept that, done.

sshd-core/src/main/java/org/apache/sshd/client/config/hosts/DefaultConfigFileHostEntryResolver.java
lgoldstein: I would have accepted _!isEmpty_ - however, as a matter of style, I find it easier to ask:
```java
if (length(a) > 0) {
    return ...;
}
```
rather than
```java
if (!isEmpty(a)) {
}
```

alonbl:
will you accept the validate method return a list of strings instead of permissions? this way I can write single function instead of two, there is no reason to return permissions as far as I can see.

sshd-core/src/main/java/org/apache/sshd/server/config/keys/DefaultAuthorizedKeysAuthenticator.java
lgoldstein: Will be faster if you simply correct these 2 strings...

sshd-core/src/main/java/org/apache/sshd/client/config/hosts/DefaultConfigFileHostEntryResolver.java
alonbl: I think that symmetric code is more important, either you ask for length or for empty in all cases.

lgoldstein:
I don't understand what you mean by "will you accept the validate method return a list of strings instead of permissions? this way I can write single function instead of two, there is no reason to return permissions as far as I can see." can you include a skeleton code snippet that demonstrates what you mean and how you intend to use it ?

sshd-core/src/main/java/org/apache/sshd/server/config/keys/DefaultAuthorizedKeysAuthenticator.java
alonbl: ok, usually each patch should touch its task, done.

sshd-core/src/main/java/org/apache/sshd/client/config/hosts/DefaultConfigFileHostEntryResolver.java
lgoldstein: Let's agree to disagree

sshd-core/src/main/java/org/apache/sshd/common/util/io/IoUtils.java
lgoldstein: Then we are fine - make this change and re-submit (including moving the code to _OsUtils_ and replacing the "user.name" usage in the code with calls to _OsUtils.getCurrentUser()_).

alonbl:
> can you include a skeleton code snippet that demonstrates what you mean and how you intend to use it ?

For example:
```
public static Collection<String> validateStrictConfigFilePermissions(Path path, LinkOption... options) throws IOException
```
This will return a string messages about permissions issues/owner issues in single function.

lgoldstein:
I thought as much but wanted to be sure. You are *technically* right, it would be more efficient, but this is not a big issue for this specific code which invoked only first time or whenever the file changes (which is very rarely). The prevailing consideration here is debug-ability - I want the exception or log message to be clear about the reason for the failure - permissions or ownership. If we use the function you propose way we just print some strings without context.

alonbl:
I do not understand, the string will contain the context and proper message. Much safer to use one function to validate instead of trying to remember to call both every place.

alonbl:
we can also return single string if we are interested in only 1st error.

lgoldstein:
If I remember correctly we are interested only in the 1st error - i.e., if permissions fail we do not check ownership (which is OK). If this is indeed **always** the case, I prefer the following:
```java
public static Pair<String, Object> validateStrictConfigFilePermissions(Path path, LinkOption... options) throws IOException {
```
Where if _null_ then validation successful, otherwise _getFirst_=text explaining the failure reason (e.g., "permission", "owner") and _getSecond()_ contains the offending object - which may be a _PosixFilePermission_ or a string of the offending owner or group.

alonbl:
please review, I need to do more testing, but in the meantime check if you have any more comments.

sshd-core/src/main/java/org/apache/sshd/common/config/keys/KeyUtils.java
```diff
@@ -144,20 +161,52 @@ public static PosixFilePermission validateStrictKeyFilePermissions(Path path, Li
...
+        if (!expected.contains(owner)) {
+            return new Pair(String.format("Owner violation (%s)", owner), owner);
+        }
+
+        if (OsUtils.isUNIX()) {
+            if (Files.isRegularFile(path, options)) {
+                String parentOwner = IoUtils.getFileOwner(path.getParent(), options);
+                if (!expected.contains(parentOwner)) {
+                    return new Pair(String.format("Parent owner violation (%s)", parentOwner), parentOwner);
                 }
```
lgoldstein:
```java
if ((!GenericUtils.isEmpty(parentOwner)) && (!expected.contains(parentOwner))) {
     return ...
```

lgoldstein:
Other than the minor issue I indicated seems fine.

sshd-core/src/main/java/org/apache/sshd/common/config/keys/KeyUtils.java
alonbl: there is no need, if we do have owner file system support, then it must exist, if return null then it won't be in the collection and we fail.

alonbl:
ok, looks like working, can you please double check this at your side?

alonbl:
Can you please consider DefaultConfigFileHostEntryResolver#reloadHostConfigEntries to not throw exception if config file has incorrect permissions and just ignore that file? There is no reason why a program should not continue working in this case.

lgoldstein:
You are right - will do.

alonbl:
Merged

PR #19, opened 2015-11-22
[SSHD-586] OpenSSH compliant public key fingerprint

alonbl:


sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
```diff
@@ -0,0 +1,150 @@
...
+        d.update(buffer.array(), 0, buffer.wpos());
+
+        byte[] data = d.digest();
+
+        if ("MD5".equals(d.getAlgorithm())) {
+            return "MD5:" + BufferUtils.printHex(':', data);
+        } else {
+            return d.getAlgorithm().replace("-", "").toUpperCase() + ":" + Base64.encodeToString(data).replaceAll("=", "");
+        }
+    }
```
lgoldstein: I recommend the following (D.R.Y ...):
```java
String algo = d.getAlgorithm();
if (BuiltinDigests.md5.getAlgorithm().equals(algo)) {
    return algo + ":" + BufferUtils.printHex(':', data);
} else {
    return algo.replace("-", "").toUpperCase() + ":" + Base64.encodeToString(data).replaceAll("=", "");
}
```

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * @param expected The expected fingerprint.
+     * @param key the public key - ignored if {@code null}
```
lgoldstein: If {@code null} / empty then returns a failure indication with the default fingerprint. If no digest specified the the default digest is used.

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
```diff
@@ -0,0 +1,150 @@
...
+     * @return Pair<Boolean, String> - first is success indicator, second is actual fingerprint.
+     */
+    public static Pair<Boolean, String> checkFingerPrint(String expected, PublicKey key) throws Exception {
+        if (key == null) {
+            return null;
+        }
+
+        if (expected == null) {
+            return new Pair<>(false, getFingerPrint(key));
+        }
```
lgoldstein:
```java
if (GenericUtils.isEmpty(expected)) {
    return new Pair<>(false, getFingerPrint(key));
}
```

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
```diff
@@ -0,0 +1,150 @@
...
+
+        String comps[] = expected.split(":", 2);
+        DigestFactory factory;
```
lgoldstein: if (GenericUtils.length(comps) < 2) {
   return new Pair<>(false, getFingerPrint(key));
}

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
```diff
@@ -0,0 +1,150 @@
...
+        if (comps[0].length() > 2) {
+            factory = BuiltinDigests.fromString(comps[0]);
+            if (factory == null) {
+                return new Pair<>(false, getFingerPrint(key));
+            }
```
lgoldstein: Add a remark explaining why this is true:
```java
// We know that all digests have a length > 2 - if 2 (or less) then assume a pure HEX value
```

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
```diff
@@ -0,0 +1,150 @@
...
+        } else {
+            factory = BuiltinDigests.md5;
+            expected = "MD5:" + expected;
+        }
```
lgoldstein: D.R.Y. (Do not Repeat Yourself):
```java
factory = BuiltinDigests.md5;
expected = factory.getAlgorithm() + ":" + expected;
```

sshd-core/src/main/java/org/apache/sshd/common/util/Pair.java
```diff
@@ -45,6 +45,29 @@ public S getSecond() {
...
+        }
+
+        if (obj == this) {
+            return true;
+        }
+
+        if (!(obj instanceof Pair)) {
+            return false;
+        }
+
```
lgoldstein: Unless a good reason for this, the following is considered more correct:
```java
if (getClass() != obj.getClass()) {
    return false;
```
}
```

File: sshd-core/src/test/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtilsTest.java [hunk 8]
```diff
@@ -0,0 +1,130 @@
...
+        public KeyEntry(String keyString, Pair<DigestFactory, String>... hashes) {
+            this.keyString = keyString;
+            this.hashes = hashes;
+        }
+    }
+
+    private static KeyEntry[] KEY_ENTRIES = new KeyEntry[] {
+        new KeyEntry(
+            "ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAxr3N5fkt966xJINl0hH7Q6lLDRR1D0yMjcXCE5roE9VFut2ctGFuo90TCOxkPOMnwzwConeyScVF4ConZeWsxbG9VtRh61IeZ6R5P5ZTvE9xPdZBgIEWvU1bRfrrOfSMihqF98pODspE6NoTtND2eglwSGwxcYFmpdTAmu+8qgxgGxlEaaCjqwdiNPZhygrH81Mv2ruolNeZkn4Bj+wFFmZTD/waN1pQaMf+SO1+kEYIYFNl5+8JRGuUcr8MhHHJB+gwqMTF2BSBVITJzZUiQR0TMtkK6Vbs7yt1F9hhzDzAFDwhV+rsfNQaOHpl3zP07qH+/99A0XG1CVcEdHqVMw== lgoldstein@LGOLDSTEIN-WIN7",
+            new Pair<>(
```
lgoldstein: I prefer:
```java
private static final List<KeyEntry> KEY_ENTRIES =
    Collections.unmodifiableList(Arrays.asList(
          new KeyEntry(...),
          new KeyEntry(...),
          ...));
```
This ensures that the test code does not inadvertently modify the tested data.

sshd-core/src/test/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtilsTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+    private static class KeyEntry {
+        public final String keyString;
+        public final Pair<DigestFactory, String>[] hashes;
+        public PublicKey key;
+        public KeyEntry(String keyString, Pair<DigestFactory, String>... hashes) {
+            this.keyString = keyString;
+            this.hashes = hashes;
+        }
+    }
```
lgoldstein: I prefer:
```java
private static class KeyEntry {
        public final String keyString;
        public final List<Pair<DigestFactory, String>> hashes;
        public PublicKey key;
        public KeyEntry(String keyString, Pair<DigestFactory, String>... hashes) {
            this.keyString = keyString;
            this.hashes = Collections.unmodifiableList(Arrays.asList(hashes));
        }
    }
```
This ensure that the test code does not inadvertently change the tested data

sshd-core/src/test/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtilsTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+    @BeforeClass
+    public static void beforeClass() throws IOException, GeneralSecurityException {
+        for (KeyEntry entry : KEY_ENTRIES) {
+            try {
+                entry.key = PublicKeyEntry.parsePublicKeyEntry(entry.keyString).resolvePublicKey();
+            } catch (NoSuchProviderException e) {
+                // ignore key
+            }
+        }
```
lgoldstein: Add an output message indicating this so we'll know it happened when we examine the tests output
```java
} catch (NoSuchProviderException e) { // ignored
       System.out.println("Skip unsupported key: " + entry.keyString);
}
```

sshd-core/src/test/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtilsTest.java
```diff
@@ -0,0 +1,130 @@
...
+                    );
+                    assertEquals(
+                        String.format("Fingerprint check succeeded for key %s invalid digest %s", kentry.key.getAlgorithm(), dentry.getFirst().getName()),
+                        new Pair<Boolean, String>(false, dentry.getSecond()),
+                        OpenSSHKeyUtils.checkFingerPrint(dentry.getSecond() + "A", kentry.key)
+                    );
+                }
+            }
+        }
+    }
```
lgoldstein: I am not a big fan of repeated "getXXX" calls to the same value. Preferably:
```java
PublicKey key = kentry.key;
if (key != null) {
    String algo = key.getAlgorithm();
    for (Pair<DigestFactory, String> dentry : kentry.hashes) {
        DigestFactory factory = dentry.getFirst();
        String name = factory.getName();
        String expected = dentry.getSecond();
        assertEquals(
                  String.format("Fingerprint does not match for key %s digest %s", algo, name),
                  expected, OpenSSHKeyUtils.getFingerPrint(factory, key));
         ...the other asserts...
    }
}
```
Easier to read and understand instead of having to remember what "getFirst" and "getSecond" means

lgoldstein:
> not sure what is better, to return a pair or to have an appendable parameter for the actual digest

Definitely a _Pair_, but maybe instead of returning _true / false_ as the first value we should return a _FingerprintCheckResult_:
```java
public static class FingerprintCheckResult {
    private final boolean matching;
    private final String reason;   // if match failed then some reason - e.g., "unknown factory", "malformed data", etc... null/empty if matching
    private final String expected;

    public FingerprintCheckResult(boolean matching, String reason, String expected) {
       ...this.xxx = xxx...
    }

   public boolean isMatching() { ... }
   public String getReason() { ... }
   public String getExpectedFingerPrint() { ... }

   @Override
   public String toString() { ... }

   @Override
   public int hashCode() { ... }

   @Override
   public int equals() { ... }
}
```

lgoldstein:
Instead of running a loop inside the test consider writing a _Parameterized_ test - see [Parameterized tests](https://github.com/junit-team/junit/wiki/Parameterized-tests). There are also some existing tests in SSHD already so you can see how this is done - e.g., _CopyDataExtensionImplTest_ - especially how to choose a **meaningful** _name_ value. You can replace the _@BeforeClass_ with the _parameters()_ method and create the actual list of keys to be tested then and there.

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
```diff
@@ -0,0 +1,150 @@
...
+
+        String fingerprint = getFingerPrint(factory, key);
+        return new Pair<>(expected.equals(fingerprint), fingerprint);
```
lgoldstein: For MD5 you must use _expected.equalsIgnoreCase(fingerprint)_ in case HEX is uppercase vs. lowercase.

sshd-core/src/main/java/org/apache/sshd/common/util/Pair.java
```diff
@@ -45,6 +45,29 @@ public S getSecond() {
...
+        Pair<?, ?> other = (Pair<?, ?>)obj;
+        return Objects.equals(first, other.first) && Objects.equals(second, other.second);
+    }
```
lgoldstein: Since _Pair_ is not _final_ (and neither are its _get_-ters) I prefer:
```java
return Objects.equals(getFirst(), other.getFirst()) && Objects.equals(getSecond(), other.getSecond());
```

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
alonbl: Done

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
alonbl: done, I hope.

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
alonbl: Done

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
alonbl: Done.

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
alonbl: Done

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
alonbl: Done.

sshd-core/src/main/java/org/apache/sshd/common/util/Pair.java
alonbl: OK, I just copied this from other place in code :)
Done.

sshd-core/src/test/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtilsTest.java
alonbl: Done

sshd-core/src/test/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtilsTest.java
alonbl: Done

sshd-core/src/test/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtilsTest.java
alonbl: Doe

sshd-core/src/test/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtilsTest.java
alonbl: Done

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
alonbl: I solved it by converting the expected to lower case and make sure we return lower case signature.

sshd-core/src/main/java/org/apache/sshd/common/util/Pair.java
alonbl: Done

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * @return The default {@link Factory} of {@link Digest}s used
+     * by the {@link #getFingerPrint(PublicKey)} and {@link #getFingerPrint(String)}
+     * methods
+     * @see #setDefaultFingerPrintFactory(Factory)
+     */
+    public static Factory<? extends Digest> getDefaultFingerPrintFactory() {
+        return DEFAULT_DIGEST_HOLDER.get();
+    }
```
lgoldstein: If you really want to be flexible you can make this configurable (I am willing to accept the code even without this)
```java
/**
  * System property that can be used to control the default fingerprint factory used for keys.
  * If not set the {@link #DEFAULT_FINGERPRINT_DIGEST_FACTORY} is used
  */
public static final String OPENSSH_KEY_FINGERPRINT_FACTORY_PROP = "org.apache.sshd.opensshFingerprintFactory";

public static Factory<? extends Digest> getDefaultFingerPrintFactory() {
    Factory<? extends Digest> factory = null;
    synchronized(DEFAULT_DIGEST_HOLDER) {
        factory = DEFAULT_DIGEST_HOLDER.get();
        if (factory != null) {
            return factory;
        }

        String propVal = System.getProperty(OPENSSH_KEY_FINGERPRINT_FACTORY_PROP);
        if (GenericUtils.isEmpty(propVal)) {
            factory = DEFAULT_FINGERPRINT_DIGEST_FACTORY;
        } else {
            factory = ValidateUtils.checkNotNull(BuiltinDigests.fromFactoryName(propVal), "Unknown digest factory: %s", propVal);
        }

        DEFAULT_DIGEST_HOLDER.set(factory);
    }

    return factory
}

```

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
lgoldstein: Not good - for SHA the fingerprint is BASE64 where case matters - only for MD5 the case does not matter...

sshd-core/src/main/java/org/apache/sshd/common/util/Pair.java
lgoldstein: What place ? In SSHD ? Show me so I can look and see if it is warranted...

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
alonbl: please wait for me to push before comment....

alonbl:
> see Parameterized tests

Done.

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
alonbl: Will do next push, not sure I understand what is the difference from KeyUtils.

lgoldstein: Agreed - let's do the same for KeyUtils while we are at it...

sshd-core/src/test/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtilsTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+    public final PublicKey key;
+    public final DigestFactory digestFactory;
+    public final String expected;
+
+    public OpenSSHKeyUtilsTest(PublicKey key, DigestFactory digestFactory, String expected) {
+        super();
+        this.key = key;
+        this.digestFactory = digestFactory;
+        this.expected = expected;
```
lgoldstein: I only use explicit super call only when it is the only statement in the constructor so I can have an easy debug breakpoint. Otherwise it should be removed.

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
```diff
@@ -0,0 +1,159 @@
...
+                return new Pair<>(false, getFingerPrint(key));
+            }
+        } else {
+            factory = BuiltinDigests.md5;
+            expected = factory.getName() + expected.toLowerCase();
+        }
+
+        String fingerprint = getFingerPrint(factory, key);
+        return new Pair<>(expected.equals(fingerprint), fingerprint);
+    }
```
lgoldstein: There is still a bug if the input already contains "MD5:0A:B3:D5:F3...." - then your _expected.equals(...)_ will fail. Therefore the code should be:
```java
if (comps[0].length() > 2) {
   ...same as before...
} else {  // we know that all digest names are more than 2 letters so we assume pure HEX MD5 here
    factory = BuiltinDigests.md5;
    expected = factory.getName().toUpperCase() + ":" + expected; 
}

String fingerprint = getFingerPrint(factory, key);
boolean matches = BuiltinDigests.md5.getFactoryName().equals(factory.getName()) ? expected.equalsIgnoreCase(fingerprint) : expected.equals(fingerprint);
return return new Pair<>(matches, fingerpint);
```
Note that you also forgot to add the ':' separator and make the factory uppercase in case of implied MD5... In this context, your code is willing to accept "sha1:..." as well as "SHA1:...." since _BuiltinDigests.fromString(...)_ is case *insensitive* - is it OK or not ? Need to decide

lgoldstein:
I have been looking at the code and my remarks some more and realized that the issue of MD5 fingerprint requires some more tests to the _OpenSSHKeyUtilsTest_:
```java
@Test
public void testMD5PureHexFingerprint() {
    Assume.assumeTrue("Skip non-MD5 fingerprint", BuiltinDigests.md5.equals(digestFactory));
    // Deconstruct MD5 fingerprint
    int pos = expected.indexOf(':');
    String pureHex = expected.substring(pos + 1);

    assertEquals("Fingerprint check failed for " + pureHex, new Pair<Boolean, String>(true, expected), OpenSSHKeyUtils.checkFingerPrint(pureHex, key));
}

@Test
public void testMD5CaseInsensitiveFingerprint() {
    Assume.assumeTrue("Skip non-MD5 fingerprint", BuiltinDigests.md5.equals(digestFactory));
    // Deconstruct MD5 fingerprint
    int pos = expected.indexOf(':');
    String name = expected.substring(0, pos);
    String pureHex = expected.substring(pos + 1);

    Pair<Boolean, String> expCheck = new Pair<Boolean, String>(true, expected);
    for (int index = 0; index < pureHex.length(); index++) {
         pureHex = shuffleCase(pureHex);
         String fingerprint = ((index & 0x01) == 0) ? pureHex : name + ":" + pureHex;
         assertEquals("Fingerprint check failed for" + fingerprint, expCheck, OpenSSHKeyUtils.checkFingerPrint(fingerprint, key));
    }
}
```

sshd-core/src/test/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtilsTest.java
alonbl: Done.

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
alonbl: Done.

alonbl:
> I have been looking at the code and my remarks some more and realized that the issue of MD5 fingerprint requires some more tests to the OpenSSHKeyUtilsTest:

yes, I thought as well to add these, should be in own class.

lgoldstein:
I will wait then for the next push...

alonbl:
OK, tell  me what you think, need to go for a while.

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
lgoldstein: Where is the code for configurable fingerprint factory we agreed on ?

sshd-core/src/test/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtilsCheckTest.java
```diff
@@ -0,0 +1,99 @@
...
+    @BeforeClass
+    public static void beforeClass() throws GeneralSecurityException, IOException {
+        key = PublicKeyEntry.parsePublicKeyEntry(KEY_STRING).resolvePublicKey(PublicKeyEntryResolver.FAILING);
+    }
+
+    @Parameters(name = "expected={0}, test={1}")
+    public static Collection<Object[]> parameters() {
+        return CASES;
+    }
+
```
lgoldstein: Move all the code for the CASES here - simply return the _Collections.unmodifiableList(...)_ - no need to declare it _private static final_

alonbl: Done, I hope.

sshd-core/src/main/java/org/apache/sshd/common/config/keys/OpenSSHKeyUtils.java
alonbl: I will post both at separate patch, ok?

lgoldstein: Fine - please do so. Merging the code...

lgoldstein:
BTW, the more I think about it, the more I am convinced that we should replace the _KeyUtils#getFingerprint_ with your code - i.e., merge _OpenSSHKeyUtils_ into _KeyUtils_. If you are working on the above mentioned patch, let's do that - should be pretty straightforward.

lgoldstein:
I'll do the merging of _OpenSSHKeyUtils _ into _KeyUtils_ - including the default digest configuration...

alonbl:
ok, thanks!

PR #20, opened 2015-11-23
[SSHD-596] Support dhg14 by runtime DH key size detection

alonbl:


lgoldstein:
Looks fine - I prefer to wait on an OK from someone else to feel confident about it...

lgoldstein:
Here is the source of my concern:
```java
public static boolean isDHGroupExchangeSupported(int maxKeySize) {
    BigInteger r = new BigInteger("0").setBit(maxKeySize - 1);
    DHParameterSpec dhSkipParamSpec = new DHParameterSpec(r, r);
    try {
        KeyPairGenerator kpg = KeyPairGenerator.getInstance("DH");
        kpg.initialize(dhSkipParamSpec);
        return true;
    } catch(Throwable t) {
        return false;
    }
}

// now let's test it
for (int keySize : new int[]{512, 1024, 2048, 3072, 4096, 8192}) {
    boolean supported = isDHGroupExchangeSupported(keySize);
    System.out.println("Key size " + keySize + " supported=" + supported);
}
```
On an _Oracle_ JVM 1.8.0_66-b18 (without any _Bouncycastle_) the output is:

>Key size 512 supported=true
>Key size 1024 supported=true
>Key size 2048 supported=true
>Key size 3072 supported=false
>Key size 4096 supported=false
>Key size 8192 supported=false

As you can see, it seems that 2048 is supported out-of-the-box - even without the extra cryptograhic strength policy. It therefore begs the question, why whoever wrote the initial code decided to ask about _Bouncycastle_ for 2048 bits but not for 1024 ? It may have been an oversight - I don't know...

alonbl:
without max strength jce? also in 1.7?

alonbl:
> without max strength jce? also in 1.7?

Indeed, java-7 supports DH 1024 no more see Keysize Restrictions[1]

[1] https://docs.oracle.com/javase/7/docs/technotes/guides/security/SunProviders.html

lgoldstein:
You are right - when I run the same code on Java 1.7.0_67 I get:
>Key size 512 supported=true
>Key size 1024 supported=true
>Key size 2048 supported=false
>Key size 3072 supported=false
>Key size 4096 supported=false
>Key size 8192 supported=false

I feel more confident now and will merge the request

alonbl:
Merged, thanks!

PR #21, opened 2015-11-23
Builder: supress heap pollution vararg warning

alonbl:
This warning was always annoying.
I hope this is the right solution.


lgoldstein:
Done - merged

PR #24, opened 2016-02-13
SSHD-642

jonnyzzz:
Check expected signature length and add missing zeroes if needed.
Tests is also added
It turned out Bouncy Castle does not fail if signature size is less than public key modulus size


oleg-nenashev:
@jonnyzzz From what I see the change has been integrated externally, so no need to keep this PR.

jonnyzzz:
@oleg-nenashev looks like that. Will close it, thanks

PR #27, opened 2016-10-30
[SSHD-706] AbstractSession#checkForTimeouts: Avoid eager log4j invoca…

davido:
…tions

Without Logger.isDebugEnabled() guard, the log message was always
constructed even though the log level was less than debug.


davido:
Thanks for merging.

PR #29, opened 2017-02-08
Use RSA for generated host keys by default

jglick:
As described [here](http://stackoverflow.com/a/33692432/12916), the default DSS algorithm is rejected by modern OpenSSH clients for security reasons. To make the server work smoothly by default it is better to use the RSA algorithm.


lgoldstein:
Merged - thanks...

jglick:
Merged in c01a6c4f918416d4fc1875fb37d8bc97ab6623f1 it seems.

PR #30, opened 2017-03-28
[SSHD-734] Attach session before starting sshd exchanges in ClientSes…

grgrzybek:
…sionImpl constructor


grgrzybek:
I see this is already fixed

PR #31, opened 2017-05-26
[SSHD-750] Allow SSH 1.99 in addition to 2.0 for SSHD

nu11ptr:
Several versions ago, SSH 1.99 support was added to the client to allow it to connect to Cisco devices. I'm now proposing the same be done to allow Cisco devices still using 1.99 to connect to SSHD.

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/server/session/AbstractServerSession.java
+++ b/sshd-core/src/main/java/org/apache/sshd/server/session/AbstractServerSession.java
@@ -347,7 +347,7 @@ public abstract class AbstractServerSession extends AbstractSession implements S
         }
 
         String errorMessage = null;
-        if ((errorMessage == null) && (!clientVersion.startsWith(DEFAULT_SSH_VERSION_PREFIX))) {
+        if (!(clientVersion.startsWith(DEFAULT_SSH_VERSION_PREFIX) || clientVersion.startsWith("SSH-1.99-"))) {
             errorMessage = "Unsupported protocol version: " + clientVersion;
         }
 


lgoldstein:
Before we can address this, please open a proper JIRA issue. That being said let's look at What does 1.99 mean  - it involves more than just willing to accept the version.

>> SSH servers that advertise themselves as 1.99 (e.g. SSH-1.99-OpenSSH_4.1) are really SSH-2 servers that are configured to fall back to using SSH-1 if requested.

or

>> Server implementations MAY support a configurable compatibility flag that enables compatibility with old versions. When this flag is on, the server SHOULD identify its 'protoversion' as "1.99". Clients using protocol 2.0 MUST be able to identify this as identical to "2.0". In this mode, the server SHOULD NOT send the Carriage Return character (ASCII 13) after the identification string.

>> In the compatibility mode, the server SHOULD NOT send any further data after sending its identification string until it has received an identification string from the client. The server can then determine whether the client is using an old protocol, and can revert to the old protocol if required. In the compatibility mode, the server MUST NOT send additional data before the identification string.

and from [RFC-4253 - section 5](https://tools.ietf.org/html/rfc4253#section-5):

>> Earlier versions of this protocol have not been formally documented, but it is widely known that they use 'protoversion' of   "1.x" (e.g., "1.5" or "1.3"). 

So by willing to accept it, we also undertake responsibility to be SSH-1 compliant, which I am not sure that we want to, especially since it is considered effectively obsolete. With all the sympathy to the issue that you have come across, the "fix" does not involve just accepting the version but also a willingness to be SSH-1 compliant. Therefore, while it may fix the immediate proble it make SSHD more complex than we want to.

nu11ptr:
Thanks for your comments and consideration. I would agree about not talking SSH 1.x and that isn't what the patch does or suggests. I think what you list from the RFC pertains to the opposite situation (a server advertising 1.99 and a client advertising 2.0). It is worth noting that the MINA SSH client ALREADY handles talking to a 1.99 server via a PR made a while back, so it is only the server code that doesn't handle 1.99 clients currently. As long as both only talk 2.x, I'd expect the client and server capabilities to be in sync, and based on what I can tell, after my patch, the client and server version compat. check look identical.

This situation is about a client advertising 1.99 to Apache SSHD (which is advertising 2.0 only). As such, by NOT handling 1.99 clients, SSHD is taking a stance of not talking to an SSH 2.0 client...simply because it also supports 1.0 fallback. Perhaps I'm simply not familiar enough with the RFC, but I'm not seeing a security issue at play here (if this opened a hole where the server would now talk 1.x that would be a different story, but that isn't the request).  (NOTE: Apparently the RFC is ambiguous on this: https://community.hpe.com/t5/Networking/SSH-1-99-sent-by-client-not-RFC-compliant/td-p/4688885)

Also, note this probably never came up before because I suspect most proper SSH implementations long ago just dropped 1.x support. However, there is "another world" to consider and that is the embedded world where code is not often updated (for better or worse), and specifically for me, Cisco devices. While it is easy to disable 1.x support via the proper command, there are MANY devices out in the wild where this command is not issued. Worst of all (and this is a Cisco issue, not MINA), the error you receive when you try to connect is not helpful at all, so the typical user would have no idea how to fix the issue. I had to do a packet capture just to find the problem (and could then see MINA SSHD was helpful in saying "unsupported version").

I don't have an issue opening a JIRA, but overall I do think it makes sense to accept this PR or something like it based on what I understand it does. I would also agree before coming to final conclusion it would be helpful to fully understand the ramifications, however. I've already patched it in my version, and it works fine, but I thought it might be helpful to others so that is why I propose.

lgoldstein:
You make some valid points - I will look into it - in any case, this patch will be available only as of version 1.6 whose release date is not set.

PR #32, opened 2017-05-31
[SSHD-710] fix ed25519

pottedplant:
fix eddsa public key recovery
fix ed25519 openssh key parser


pottedplant:
merged via 02f1d57

PR #35, opened 2017-08-01
[SSHD-758] Add OpenSSHGeneratorFileKeyProvider

cschneider:


lgoldstein:
As indicated in the issue we don't intend to use this code in the foreseeable future. Even if we were, it is a long way from being up to the project's standards - especially lacking full coverage of **all** possible keys - not just RSA - and all possible PEM formats (there are quite a few of them...)

tomaswolf:
Closing old outdated PRs.

PR #38, opened 2017-10-11
[SSHD-776] SSHD local port forwarding close session unexpectedly

fcav1973:
This patch contains the following modifications:

1. _AbstractConnectionService.channelWindowAdjust_: we did the same modifications as for channelEof; checking RFC there is the same level of ambiguity
2. _Nio2Session.exceptionCaught_: the session should be closed anyway (the close method is already resilient to subsequent calls). This was necessary because we found some messages in write queue after the exception management.
3. _TcpipServerChannel.close_: this have been changed to let the connector to close immediatly os gracefully, accordingly to the way the API is called. Most of the problem we have found where related write messages still pending during a gracefull close
4. _TcpipServerChannel.handleWriteDataFailure_: this have been necessary because in case the remote server had closed the connection, some data could be still waiting in the channel to be written to the outgoing session (even if channel close have been already sent) that could be already closed. In that case we should simply have to discard those error.


fcav1973:
Merged with https://github.com/apache/mina-sshd/commit/0e99597aceb035d271f0140348df23d32b5e3bcd

PR #42, opened 2018-01-08
SSHD-746 add minimal support so IPv6 addresses do not throw exceptions.

daik:
This is the bare minimum required to make it possible to connect to hosts identified by an IPv6 address. I have tested this to confirm it works. The aim here is to make it possible (i.e no exceptions thrown) to connect to hosts defined by only an IPv6 address.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -327,6 +327,11 @@
                 </exclusions>
             </dependency>
             <dependency>
+                <groupId>commons-validator</groupId>
+                <artifactId>commons-validator</artifactId>
+                <version>1.6</version>
+            </dependency>
+            <dependency>
                 <groupId>org.apache.httpcomponents</groupId>
                 <artifactId>httpclient</artifactId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -326,6 +326,11 @@
                     </exclusion>
                 </exclusions>
             </dependency>
+            <dependency>
+                <groupId>commons-validator</groupId>
+                <artifactId>commons-validator</artifactId>
+                <version>1.6</version>
+            </dependency>
```
lgoldstein: Sorry, but we try to keep the code as independent as possible - the only mandatory dependency is _slf4j_. If you want this code, in then copy the relevant methods into our code (with the necessary adjustments) - I suggest placing them in the `SshdSocketAddress` class (which already has some IPv4 related code...)

daik:
lgoldstein kindly did the work needed, and thus this PR is now obsolete.

PR #43, opened 2018-01-11
[SSHD-793] Fix the memory leak issue of sshd server

lifangning:
[SSHD-793] Fix the memory leak issue of continuing to connect and disconnect to sshd.


sshd-core/src/main/java/org/apache/sshd/common/io/nio2/Nio2Acceptor.java
```diff
@@ -189,8 +189,12 @@ protected void onCompleted(AsynchronousSocketChannel result, SocketAddress addre
...
                 session = Objects.requireNonNull(createSession(Nio2Acceptor.this, address, result, handler), "No NIO2 session created");
                 handler.sessionCreated(session);
-                sessions.put(session.getId(), session);
-                session.startReading();
+                if (!session.isClosing()) {
+                    sessions.put(session.getId(), session);
+                    session.startReading();
+                } else {
+                    handler.sessionClosed(session);
+                }
```
lgoldstein: Perhaps in order to avoid race conditions the following is better:
```java
Long sessionId = session.getId();
sessions.put(sessionId, session);
if (session.isClosing()) {
    try {
        handler.sessionClose(session);
    } finally {
        sessions.remove(sessionId);
    }
} else {
    session.startReading();
}
````

PR #44, opened 2018-02-01
[SSHD-799] SSHD port forwarding. Missing SSH_MSG_CHANNEL_OPEN_FAILURE

fcav1973:
message


sshd-core/src/main/java/org/apache/sshd/common/session/helpers/DefaultUnknownChannelReferenceHandler.java
```diff
@@ -74,8 +74,8 @@ public Channel handleUnknownChannelCommand(
 
             case SshConstants.SSH_MSG_CHANNEL_DATA:
             case SshConstants.SSH_MSG_CHANNEL_EXTENDED_DATA:
-                // Not sure if entirely compliant with RFC4254, but try to stem the flood
-                wantReply = true;
+                // Not sure if entirely compliant with RFC4254, but try to stem the flood	
+                wantReply = false;
                 break;
```
lgoldstein: This is part of SSHD-800 and should not be part of this patch

fcav1973:
I reverted the commits related to [SSHD-800] and [SSHD-801] for which I openeted https://github.com/apache/mina-sshd/pull/45 PR

PR #45, opened 2018-02-02
[SSHD-801] SSHD port forwarding. Channel closed abruptly on remote

fcav1973:
socket error

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/server/forward/TcpipServerChannel.java
+++ b/sshd-core/src/main/java/org/apache/sshd/server/forward/TcpipServerChannel.java
@@ -180,7 +180,18 @@ public class TcpipServerChannel extends AbstractServerChannel {
 
             @Override
             public void exceptionCaught(IoSession session, Throwable cause) throws Exception {
-                close(true);
+            	if( isOpen() ) {
+                    if (log.isDebugEnabled()) {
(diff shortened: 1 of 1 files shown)


sshd-core/src/main/java/org/apache/sshd/server/forward/TcpipServerChannel.java
```diff
@@ -328,7 +339,12 @@ protected void handleWriteDataFailure(byte cmd, byte[] data, int off, int len, T
         }
 
         if (ioSession.isOpen()) {
-            session.exceptionCaught(t);
+			// SSHD-795 IOException (Broken pipe) on a socket local forwarding channel
+			// causes SSH client-server connection down...
+        	if(log.isDebugEnabled()) {
+        		log.debug("handleWriteDataFailure Ignoring {}", t.getMessage());
+        	}
+			// session.exceptionCaught(t);
```
lgoldstein: This effectively means that any failure to write data keeps the channel open. If the _ioSession_ still open, let's at least abort the channel - i..e, close it forcefully somehow - perhaps even send an `SSH_MSG_CHANNEL_FAILURE` with a proper reason code and disconnect message

fcav1973:
Closing this PR because modification have been already integrated, thanks to @lgoldstein

PR #46, opened 2018-02-02
[SSHD-799] SSHD port forwarding. Missing SSH_MSG_CHANNEL_OPEN_FAILURE

fcav1973:
New PR containing only commit for the issue in subject

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractConnectionService.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractConnectionService.java
@@ -639,19 +639,21 @@ public abstract class AbstractConnectionService<S extends AbstractSession>
                     buf.putInt(window.getPacketSize());
                     session.writePacket(buf);
                 } else {
+                    int reasonCode = 0;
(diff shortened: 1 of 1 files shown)


fcav1973:
I've seen that `isOpened() == true` implies `isDone() == true` (at least in current `DefaultOpenFuture` implementation), so it probably should not be a problem.
This probably happen because in method

```
    protected void handleChannelOpenFailure(OpenFuture f, Throwable problem) {
        signalChannelOpenFailure(problem);
        notifyStateChanged(problem.getClass().getSimpleName());
        close(true);

        if (problem instanceof ConnectException) {
            f.setException(new SshChannelOpenException(getId(), SshConstants.SSH_OPEN_CONNECT_FAILED, problem.getMessage(), problem));
        } else {
            f.setException(problem);
        }

    }

```

`close(true)` is called before `f.setException` causing the `connectFeature` to be closed before the exception is set

lgoldstein:
The let's change this and call `close` after setting the exception...

fcav1973:
Closing this PR because modification have been already integrated, thanks to @lgoldstein

PR #47, opened 2018-02-07
[SSHD-803] SSHD port forwarding. Missing SSH_MSG_CHANNEL_EOF

fcav1973:


sshd-core/src/main/java/org/apache/sshd/server/forward/TcpipServerChannel.java
```diff
@@ -260,6 +260,20 @@ protected void handleChannelOpenFailure(OpenFuture f, Throwable problem) {
 
     @Override
     public CloseFuture close(boolean immediately) {
+
+        if (!immediately && (out != null)) {
+            try {
+                if (log.isDebugEnabled()) {
+                    log.debug("Closing ChannelOutputStream: {}", TcpipServerChannel.this);
+                }
+                out.close();
```
fcav1973: No completely sure this is the correct way (`TcpipServerChannel.close(false)` is also called in case of `handleWriteDataFailure` and `ioHandler.exceptionCaught`) but it seems to be the only  way for the `SSH_MSG_CHANNEL_EOF` message to be send back to the ssh client

sshd-core/src/main/java/org/apache/sshd/server/forward/TcpipServerChannel.java
```diff
@@ -260,6 +260,20 @@ protected void handleChannelOpenFailure(OpenFuture f, Throwable problem) {
...
+            } catch (IOException ignored) {
+                if (log.isErrorEnabled()) {
+                    log.error("Error closing ChannelOutputStream: {}", TcpipServerChannel.this, ignored);
+                }
+            }
+        }
```
lgoldstein: 1. We never ask if ERROR enabled - only DEBUG and TRACE
2. If this is an ignored exception why log it as ERROR ? DEBUG or TRACE would seem more appropriate

sshd-core/src/main/java/org/apache/sshd/common/io/nio2/Nio2Session.java
```diff
@@ -312,7 +312,7 @@ protected void handleReadCycleCompletion(
                 if (log.isDebugEnabled()) {
                     log.debug("handleReadCycleCompletion({}) Socket has been disconnected (result={}), closing IoSession now", this, result);
                 }
-                close(true);
+                close(false);
```
lgoldstein: Why `close(false)` ? Let's add some comment explaining so in the future it is not changed back inadvertently

sshd-core/src/main/java/org/apache/sshd/server/forward/TcpipServerChannel.java
lgoldstein: 1. Style  - parentheses around `(!immediately)`
2. Code seems a bit complex and not well documented to explain why this is done - why is closing what seems to be an `OutputStream` same as sending `SSH_MSG_CHANNEL_EOF`
3. If indeed `SSH_MSG_CHANNEL_EOF`  is required only in a certain scenario why not send it explicitly ? We do have a `sendOf()` method that can be invoked by the channel on itself (and it is even idempotent...)

fcav1973: 1. Ok, I'll change it
2. 3. You're right It's a little bit wired and I'll try to add some comments to explain it better. Anyway, after having reviewed the code, to me it seems that out.close() is more approriate than simply sendEof() because close() will also call flush(), avoiding some data not beign sent back to client.

sshd-core/src/main/java/org/apache/sshd/common/io/nio2/Nio2Session.java
```diff
@@ -309,10 +309,16 @@ protected void handleReadCycleCompletion(
...
+                // gracefully (eg. TcpipServerChannel need to send and ordered
+                // SSH_MSG_CHANNEL_EOF and SSH_MSG_CHANNEL_CLOSE sequence).
+                boolean immediately = false;
+
                 if (log.isDebugEnabled()) {
                     log.debug("handleReadCycleCompletion({}) Socket has been disconnected (result={}), closing IoSession now", this, result);
                 }
-                close(true);
+
+                close(immediately);
```
lgoldstein: This decision seems too early - `Nio2Session` does not know it is invoked as part of the channel or session or whatever. Is there some way we can limit this to just the `TcpipServerChannel` ? Somehow detect it is being invoked in that context ?

fcav1973: At the moment it's not clear to me what happens using `MINASession`, which looks like a complete different approach. I understand you comments on releation between `Nio2Session` and `TcpipServerChannel`. 
Anyway, I think that channel EOF should be consided a 'normal' condition so that a 'graceful' shoutdown should be trigered.
As far as I understood, the relation between `Nio2Session` and 'TcpipServerChannel` seems to be 'embedded' inside the `ioHandler`. May be we could change 'IoHandler' interface and transform `sessionClosed(IoSession session)` into `sessionClosed(IoSession session, boolean immediatly)`, so to let specific `ioHandler` do the right job inside `handleReadCycleCompletion`.

Like:

```
            } else {
                if (log.isDebugEnabled()) {
                    log.debug("handleReadCycleCompletion({}) Socket has been disconnected (result={}), closing IoSession now", this, result);
                }
                IoHandler handler = getIoHandler();
                handler.sessionClosed(this, false);

                close(true);
            }
```

PR #52, opened 2018-04-13
[SSHD-640] Streams are not flushed from SSH Command

gnodet:


sshd-core/src/main/java/org/apache/sshd/server/channel/ChannelSession.java
```diff
@@ -801,6 +801,9 @@ protected void closeShell(int exitValue) throws IOException {
         }
 
         if (!isClosing()) {
+            if (out != null) {
+                out.flush();
+            }
```
lgoldstein: Recommend try-catch block around the `flush()` with warning log message if exception on flushing. We want the EOF + exitValue to be sent even if flushing fails

PR #55, opened 2018-04-18
Sshd 817

gnodet:


sshd-core/src/test/java/org/apache/sshd/server/auth/AsyncAuthInteractiveTest.java
```diff
@@ -0,0 +1,105 @@
...
+            }
+        }
+        channel = (ChannelShell) session.openChannel("shell");
+        channel.connect();
+
+        if (channel != null) {
+            try {
+                channel.disconnect();
+            } catch (Exception ignore) {
+            }
```
lgoldstein: Missing `// ignored` comment - I believe our _checkstyle_ should fail on this...

sshd-core/src/test/java/org/apache/sshd/server/auth/AsyncAuthInteractiveTest.java
```diff
@@ -0,0 +1,105 @@
...
+        }
+        if (session != null) {
+            try {
+                session.disconnect();
+            } catch (Exception ignore) {
+            }
```
lgoldstein: Missing `// ignored` comment - I believe our _checkstyle_ should fail on this...

sshd-core/src/test/java/org/apache/sshd/server/auth/AsyncAuthTest.java
```diff
@@ -0,0 +1,104 @@
...
+import org.junit.runners.MethodSorters;
+
+/**
+ * @author <a href="mailto:dev@mina.apache.org">Apache MINA SSHD Project</a>
+ */
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
+public class AsyncAuthTest extends AsyncAuthTestBase {
+
+    public AsyncAuthTest() {
+    }
```
lgoldstein: Missing `super()` call

sshd-core/src/test/java/org/apache/sshd/server/auth/AsyncAuthTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+            @Override
+            public boolean promptYesNo(String s) {
+                return true;
+            } // Accept all server keys
+
+            @Override
+            public void showMessage(String s) {
+            }
+        });
```
lgoldstein: Document empty block with `// do nothing` comment

sshd-core/src/test/java/org/apache/sshd/server/auth/AsyncAuthTestBase.java
```diff
@@ -0,0 +1,142 @@
...
+import org.apache.sshd.util.test.EchoShellFactory;
+import org.junit.After;
+import org.junit.Test;
+
+/**
+ * @author <a href="mailto:dev@mina.apache.org">Apache MINA SSHD Project</a>
+ */
+public abstract class AsyncAuthTestBase extends BaseTestSupport {
+
+    SshServer server;
```
lgoldstein: Should be _private_ as well

sshd-core/src/test/java/org/apache/sshd/server/auth/AsyncAuthTestBase.java
```diff
@@ -0,0 +1,142 @@
...
+
+    private PasswordAuthenticator authenticator;
+
+    public AsyncAuthTestBase() {
+    }
```
lgoldstein: Missing `super();` call

sshd-core/src/test/java/org/apache/sshd/server/auth/AsyncAuthTestBase.java
```diff
@@ -0,0 +1,142 @@
...
+
+    public void startServer(Integer timeout) throws Exception {
+        if (server != null) {
+            fail("Server already started");
+        }
+        server = SshServer.setUpDefaultServer();
+        if (timeout != null) {
+            server.getProperties().put(FactoryManager.AUTH_TIMEOUT, timeout.toString());
+        }
+        server.setPort(5000);
```
lgoldstein: Why do we have to use an explicit port and not let the server the 1st available one ? What if 5000 is used by the user's machine for something else ?

sshd-core/src/test/java/org/apache/sshd/server/auth/AsyncAuthTestBase.java
```diff
@@ -0,0 +1,142 @@
...
+        AsyncAuth auth = new AsyncAuth();
+        new Thread(() -> {
+            try {
+                Thread.sleep(delay);
+            } catch (InterruptedException ignore) {
+            } finally {
+                auth.setAuthed(result);
+            }
+        }).start();
+        throw auth;
```
lgoldstein: 1. Let's make it a daemon thread
2. Add `// ignored` comment in the empty _catch_ clause

sshd-netty/pom.xml
```diff
@@ -0,0 +1,214 @@
...
+            <version>${project.version}</version>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.sshd</groupId>
+            <artifactId>sshd-sftp</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
```
lgoldstein: I think the dependency should be inverted - i.e., add a profile to _sshd-sftp_ to test it with _Netty_ (see the example I did with MINA)

sshd-netty/pom.xml
```diff
@@ -0,0 +1,214 @@
...
+    </dependencies>
+
+    <build>
+        <testSourceDirectory>${build.directory}/test-sources</testSourceDirectory>
+        <testResources>
+            <testResource>
+                <directory>${build.directory}/test-resources</directory>
+            </testResource>
+        </testResources>
+        <plugins>
```
lgoldstein: Please use `${project.build.directory}` to avoid _Maven_ warnings about deprecation

sshd-netty/src/main/java/org/apache/sshd/netty/NettyIoHandlerBridge.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.sshd.common.util.Readable;
+
+/**
+ * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
+ * @author <a href="mailto:dev@mina.apache.org">Apache MINA SSHD Project</a>
+ */
+public class NettyIoHandlerBridge {
+
+    public NettyIoHandlerBridge() {
+    }
```
lgoldstein: Missing `super();` call

sshd-netty/src/main/java/org/apache/sshd/netty/NettyIoService.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public class NettyIoService extends AbstractCloseable implements IoService {
+
+    static final AttributeKey<NettyIoSession> SESSION_KEY = AttributeKey.valueOf(NettyIoSession.class.getName());
+    static final AttributeKey<IoConnectFuture> CONNECT_FUTURE_KEY = AttributeKey.valueOf(IoConnectFuture.class.getName());
+
+    final AtomicLong sessionSeq = new AtomicLong();
+    final Map<Long, IoSession> sessions = new ConcurrentHashMap<>();
+    NettyIoServiceFactory factory;
+    ChannelGroup channelGroup;
```
lgoldstein: _private_ or at least _protected_

sshd-netty/src/main/java/org/apache/sshd/netty/NettyIoService.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public NettyIoService() {
+    }
+
```
lgoldstein: Missing `super();` call

sshd-netty/src/main/java/org/apache/sshd/netty/NettyIoServiceFactory.java
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
+ * @author <a href="mailto:dev@mina.apache.org">Apache MINA SSHD Project</a>
+ */
+public class NettyIoServiceFactory extends AbstractCloseable implements IoServiceFactory {
+
+    final EventLoopGroup eventLoopGroup;
+    final boolean closeEventLoopGroup;
+
```
lgoldstein: _private_ or at least _protected_

sshd-netty/src/main/java/org/apache/sshd/netty/NettyIoServiceFactoryFactory.java
```diff
@@ -0,0 +1,49 @@
...
+
+import io.netty.channel.EventLoopGroup;
+
+/**
+ * @author <a href="mailto:julien@julienviet.com">Julien Viet</a>
+ * @author <a href="mailto:dev@mina.apache.org">Apache MINA SSHD Project</a>
+ */
+public class NettyIoServiceFactoryFactory implements IoServiceFactoryFactory {
+
+    final EventLoopGroup eventLoopGroup;
```
lgoldstein: _private_ or at least _protected_

sshd-netty/src/main/java/org/apache/sshd/netty/NettyIoSession.java
```diff
@@ -0,0 +1,195 @@
...
+    private final IoHandler handler;
+    private final DefaultCloseFuture closeFuture = new DefaultCloseFuture(toString(), lock);
+    private final long id;
+    private ChannelHandlerContext context;
+    private SocketAddress remoteAddr;
+    private ChannelFuture prev;
+    final ChannelInboundHandlerAdapter adapter = new ChannelInboundHandlerAdapter() {
+        @Override
+        public void channelActive(ChannelHandlerContext ctx) throws Exception {
+            context = ctx;
```
lgoldstein: _private_ (like the rest)

gnodet:
While I somewhat agree with the _private_ things, some of the rules (which are not enforced because the files do pass all the current rules) like adding a "do nothing" to an empty block (that's kinda intuitive that nothing happens in the empty block) or redundant constructs (like calling super()) tend to obfuscate rather than enhance the code....
Anyway, if you want some rules to be followed, they should be enforced, else they're just wishes.
I'll fix those anyway...

lgoldstein:
I am still looking for PMD and or _Checkstyle_ rules implementations for these issues (and more) - anyway, I believe we do have rules that say the empty block should be documented - that is why I was surprised it passed them (I'll look into it). On my wish list:

- Mandate `super()` call in constructor - provided no other statements in it
- Forbid indirect access to static methods/fields

gnodet:
What do you mean by _indirect access to static methods/fields_ ?

lgoldstein:
>> redundant constructs (like calling super()) tend to obfuscate rather than enhance the code

Specifically for `super()` call if no other statement exists in the constructor - the reason for this preference is **debugging** - if there is no statement in the constructor, it is sometimes very difficult if not next to impossible to place a breakpoint in the constructor and see who calls it and why/where. The same applies for the rule that is being enforced to mandates the existence of at least one constructor for a class - **debugging**. While it is true that Java rules say that a default public no-args constructor is automatically created if no other constructor is defined, I believe we need to maintain the capability to place a breakpoint anywhere of interest in the code - and I believe we can agree that a constructor can be such a location. If we have no explicit constructor or the constructor does nothing (i.e., does no execute any statement) it may be very difficult to debug the code.

lgoldstein:
>> What do you mean by indirect access to static methods/fields ?

```java
class Foo {
     public static final int MAX_FOO = 5;

    public static void throwFooException(...) {
    }
}

class Bar extends Foo {
   ...
}

class Baz {
    void doSth(int x) {
       if (x > Bar.MAX_FOO) {     // <<<<==== Foo.MAX_FOO should be accessed directly
           throw Bar.throwFooException(...);      // <<<<==== Foo.throwFooException should be accessed directly
       }
    }
}
```

I believe [this article](https://stackoverflow.com/questions/18867978/why-is-it-discouraged-to-access-static-members-indirectly) and [this one](https://books.google.co.il/books?id=DY5VDQAAQBAJ&pg=PT27&lpg=PT27&dq=java+indirect+access+to+static+member&source=bl&ots=rY4CLcuOWg&sig=Zj0_9QL766Wmo7-UgtYunUVsR44&hl=en&sa=X&ved=0ahUKEwilqcnDjcTaAhUHy6QKHTStCIoQ6AEISzAE#v=onepage&q=java%20indirect%20access%20to%20static%20member&f=false) provide some reasons why this is not a good idea.

gnodet:
Retest this please

PR #58, opened 2018-05-25
do not replace username with identity

DmitryLitvintsev:
This patch addresses the following problem:

In the existing code, once the identity is not null it replaces username. 
In case of GSS authentication the identity looks like 

   joe@EXAMPLE.ORG

It is separate and not equivalent to username. 

The issue came up while I was developing kerberized sshd that is used in admin shell interface in our 
project - dCache (dcache.org). We use apache MINA ssh server. Multiple users, each having their 
specific identities, need to login to account having one username. ACLs are defined for the username and multiple user identies map to this username (something similar to what is provided by .k5login file allowing multiple users login to the same account). The actual shell implementation picks username from Environment.ENV_USER and since exiting code replaces user w/ identity I am pretty much stuck.  

w/ this patch I am a happy customer :) 

Thank you, 
Dmitry

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/server/auth/gss/UserAuthGSS.java
+++ b/sshd-core/src/main/java/org/apache/sshd/server/auth/gss/UserAuthGSS.java
@@ -130,7 +130,7 @@ public class UserAuthGSS extends AbstractUserAuth {
                 Buffer msgbuf = new ByteArrayBuffer();
                 msgbuf.putBytes(ValidateUtils.checkNotNullAndNotEmpty(session.getSessionId(), "No current session ID"));
                 msgbuf.putByte(SshConstants.SSH_MSG_USERAUTH_REQUEST);
-                msgbuf.putString(super.getUsername());
+                msgbuf.putString(getUsername());
                 msgbuf.putString(getService());
                 msgbuf.putString(getName());
 
@@ -178,11 +178,6 @@ public class UserAuthGSS extends AbstractUserAuth {
                 }
             }
         }
-    }
-
-    @Override
-    public String getUsername() {
-        return identity != null ? identity : super.getUsername();
     }
 
     /**


lgoldstein:
I have no problem with this patch, but what if the identity is not provided ? What should be placed in the message ? Should an exception be thrown ?

DmitryLitvintsev:
I don't think so. No identity provided seem to be handled earlier and a message is clear. 

here is an example w/ this patch:

   1. w/ valid kerberos ticket:

```[root@fnisd1 opt]# klist
Ticket cache: FILE:/tmp/krb5cc_0_aziiG13005
Default principal: litvinse@FNAL.GOV
....

[root@fnisd1 opt]# ssh  -p 22224 -l admin fnisd1
dCache (4.2.0-SNAPSHOT)
Type "\?" for help.

[fnisd1] (local) admin > 
```
(note, the name 'admin' in command prompt that is the one extracted from Environment.ENV_USER. W/o this patch it would have been litvinse@FNAL.GOV which was a source of problem for me)

   2. w/o kerberos ticket. That is identity is not provided to the server

```
[root@fnisd1 opt]# kdestroy 
[root@fnisd1 opt]# ssh  -p 22224 -l admin fnisd1
Permission denied (keyboard-interactive,gssapi-with-mic).
[root@fnisd1 opt]# echo $?
255
```

lgoldstein:
Patch [merged in](https://github.com/apache/mina-sshd/commit/1a0409ee887907c07082170d5af294d8b2183d71).

DmitryLitvintsev:
Thank you! Just to make sure - am I supposed to close this request or you do it?

lgoldstein:
You should close it since you opened it...

DmitryLitvintsev:
OK, sorry then for leaving it open for so long. Closing.

PR #59, opened 2018-06-11
[SSHD-827] Close immediately channels that were not opened due to exception

janwaros:
Fix for SSHD-827


janwaros:
Merged on commit 0f75d959d61d42017f09a944e1d97eeef17b2a92.

PR #60, opened 2018-06-13
Gc socks timing

gacamp:
There is a race condition when using SOCKS proxies with the SSHD server.  After the initial SOCKS negotiation the SOCKS proxy creates a channel.  When notified that the SOCKS channel is open the SOCKS proxy then sends the final accept SOCK packet.  However there is a timing issue where very fast hosts could have already sent a packet over the now open channel - which will arrive at the client before the final SOCKS proxy packet.  This confuses the SOCKS client connected to the server as its expected a SOCKS packet and instead gets a packet from the underlying stream.

I'm uncertain if you will want to use this code or not.  The way I have fixed the race condition is a little odd in this pull request.  In the patch I couldn't find an easy way to delay the underlying protocol packets until after the connection was created, so instead I send the socks response before opening the channel (which is technically wrong) and then I delay any SOCKS responses until the channel is open.  The reason I did this was because this confines all of the required changes to the proxy, so it was a quicker fix for me.


sshd-core/src/main/java/org/apache/sshd/common/forward/SocksProxy.java
```diff
@@ -278,23 +286,25 @@ protected void onMessage(Buffer buffer) throws IOException {
...
             response.wpos(1);
-            Throwable t = future.getException();
-            if (t != null) {
-                service.unregisterChannel(channel);
-                channel.close(true);
-                response.putByte((byte) 0x01);
-            } else {
-                response.putByte((byte) 0x00);
-            }
+            response.putByte((byte) 0x00);
```
lgoldstein: Seems that channel success is sent back to the peer **before** `onChannelOpened` is invoked - i.e., if the channel open fails, the peer is simply brutally disconnected after (erroneously) thinking the the channel has been opened. I am not familiar with SOCKS protocol, but can the peer send messages before receiving an OK that channel has been opened.

gacamp: Not sure if this was written before or after my other reply, pasting in here for clarity.

HostA (SOCKS) -> SSHD -> HostB

HostA requests connection via SOCKS/SSHD to Host B - SocksInitRequest
SSHD replies SocksInitReply
HostA requests command connect via SOCKS/SSHD
SSHD connects to HostB
HostB sends first packets to SSHD 
SSHD sends these packets to HostA
SSHD replies SOCKS command connection open to HostA

As I said the patch could be confusing because I don't fix the underlying issue.  I work around it using a different flow.

 

The underlying issue is that the packets from HostB are sent to HostA, before SHHD sends the socks command response to HostA.

So the better fix would be to delay the packets from HostB until SSH has sent the socks command response to HostA.

However I couldn't see an easy way to do this.

 

So instead of this in the patch I immediately send the socks command response to HostA, to make sure its always sent before any packets from HostB.

The issue is now HostA thinks that the socks connection is active when actually we could be still opening the connection to HostB.

So HostA could start sending packets to HostB at this point.

So I then have to buffer these packets until the connection to HostB is actually open, at which point i then send on the packets.


> Seems that channel success is sent back to the peer before onChannelOpened is invoked

Yes I did this to ensure that the the channel success message is always received by the client before any packets from the channel.

> if the channel open fails, the peer is simply brutally disconnected after (erroneously) thinking the the channel has been opened. 

yes, as I said in the original comment "Note: as we are just using SSHD for testing I wasn't overly concerned with error checking or avoiding synchronization - I'm just providing it to help illustrate the issue"

> I am not familiar with SOCKS protocol, but can the peer send messages before receiving an OK that channel has been opened.

No it can't.  But that isn't the issue.
The issues is that the server is sending packets from the channel before the channel OK message is sent.

gnodet:
I can't reproduce the problem, and I'm skeptical how it can happen.  The reason is that the `onChannelOpened` is called when the `SSH_MSG_CHANNEL_OPEN_CONFIRMATION` message is received for the underlying ssh channel, and afaik, it always happen before `SSH_MSG_CHANNEL_DATA`. Please provide a unit test or any way to reproduce the problem.

gnodet:
> I can't reproduce the problem, and I'm skeptical how it can happen. The reason is that the `onChannelOpened` is called when the `SSH_MSG_CHANNEL_OPEN_CONFIRMATION` message is received for the underlying ssh channel, and afaik, it always happen before `SSH_MSG_CHANNEL_DATA`. Please provide a unit test or any way to reproduce the problem.

Actually, I found a way to reproduce the problem : https://github.com/gnodet/mina-sshd/commit/26c0f090bc7a1935c19724241184f021eba6afcd

The problem is indeed that the channel does not synchronise channel requests and the channel opening answer, so that messages can be sent on the channel before the `SSH_MSG_CHANNEL_OPEN_CONFIRMATION`.

tomaswolf:
Closing old outdated PRs.

PR #61, opened 2018-06-20
New methods requested for ssh port forwarding

robertodeandrea:
These are the minor changes required for integrate actual port forwarding mechanism in my code.

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/client/session/AbstractClientSession.java
+++ b/sshd-core/src/main/java/org/apache/sshd/client/session/AbstractClientSession.java
@@ -330,6 +330,18 @@ public abstract class AbstractClientSession extends AbstractSession implements C
         filter.stopDynamicPortForwarding(local);
     }
 
+    @Override
+    public boolean hasLocalPortForwardingStarted(SshdSocketAddress local) {
+        ForwardingFilter filter = getForwardingFilter();
+        return filter.hasLocalPortForwardingStarted(local);
+    }
+
+    @Override
+    public boolean hasRemotePortForwardingStarted(SshdSocketAddress remote) {
+        ForwardingFilter filter = getForwardingFilter();
+        return filter.hasRemotePortForwardingStarted(remote);
+    }
+
     protected ForwardingFilter getForwardingFilter() {
         ConnectionService service = Objects.requireNonNull(getConnectionService(), "No connection service");
         return Objects.requireNonNull(service.getForwardingFilter(), "No forwarder");
--- a/sshd-core/src/main/java/org/apache/sshd/client/session/AbstractClientSession.java
+++ b/sshd-core/src/main/java/org/apache/sshd/client/session/AbstractClientSession.java
@@ -331,15 +331,15 @@ public abstract class AbstractClientSession extends AbstractSession implements C
     }
 
     @Override
-    public boolean hasLocalPortForwardingStarted(SshdSocketAddress local) {
+    public boolean hasLocalPortForwardingStartedForPort(int port) {
         ForwardingFilter filter = getForwardingFilter();
-        return filter.hasLocalPortForwardingStarted(local);
+        return (filter != null) && filter.hasLocalPortForwardingStartedForPort(port);
     }
 
     @Override
-    public boolean hasRemotePortForwardingStarted(SshdSocketAddress remote) {
+    public boolean hasRemotePortForwardingStartedForPort(int port) {
         ForwardingFilter filter = getForwardingFilter();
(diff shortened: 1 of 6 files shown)


sshd-core/src/main/java/org/apache/sshd/client/session/AbstractClientSession.java
```diff
@@ -330,6 +330,18 @@ public void stopDynamicPortForwarding(SshdSocketAddress local) throws IOExceptio
...
+        ForwardingFilter filter = getForwardingFilter();
+        return filter.hasLocalPortForwardingStarted(local);
+    }
+
+    @Override
+    public boolean hasRemotePortForwardingStarted(SshdSocketAddress remote) {
+        ForwardingFilter filter = getForwardingFilter();
+        return filter.hasRemotePortForwardingStarted(remote);
+    }
+
```
lgoldstein: Please be _null_-tolerant:
```java
ForwardingFilter filter = getForwardingFilter();
return (filter != null) && filter.hasXXXX();
```

sshd-core/src/main/java/org/apache/sshd/common/forward/DefaultForwardingFilter.java
```diff
@@ -1017,4 +1017,34 @@ public void exceptionCaught(IoSession session, Throwable cause) throws Exception
...
+        ValidateUtils.checkTrue(remote.getPort() >= 0, "Invalid remote port: %s", remote);
+    
+        synchronized (remoteToLocal) {
+            SshdSocketAddress local = remoteToLocal.get(Integer.valueOf(remote.getPort()));
+            if (local != null) {
+                started = true;  
+            }
+        }      
+        return started;
+    }
```
lgoldstein: No need to use a separate variable for the return - do it a.s.a.p:
```java
synchronized (xxx) {
    return xxx.containsKey(...port...);
}
```

sshd-core/src/main/java/org/apache/sshd/common/forward/DefaultForwardingFilter.java
```diff
@@ -1017,4 +1017,34 @@ public void exceptionCaught(IoSession session, Throwable cause) throws Exception
             session.close(true);
         }
     }
+
+    @Override
+    public boolean hasLocalPortForwardingStarted(SshdSocketAddress local) {
+        boolean started = false;
+        Objects.requireNonNull(local, "Local address is null");
+        ValidateUtils.checkTrue(local.getPort() >= 0, "Invalid local port: %s", local);
```
lgoldstein: Shouldn't this be `local.getPort() > 0` - what is the meaning of port=_0_ ?

sshd-core/src/main/java/org/apache/sshd/common/forward/DefaultForwardingFilter.java
```diff
@@ -1017,4 +1017,34 @@ public void exceptionCaught(IoSession session, Throwable cause) throws Exception
...
+            }
+        }      
+        return started;
+    }
+
+    @Override
+    public boolean hasRemotePortForwardingStarted(SshdSocketAddress remote) {
+        boolean started = false;
+        Objects.requireNonNull(remote, "Remote address is null");
+        ValidateUtils.checkTrue(remote.getPort() >= 0, "Invalid remote port: %s", remote);
```
lgoldstein: Shouldn't this be `local.getPort() > 0` - what is the meaning of port=_0_ ?

sshd-core/src/main/java/org/apache/sshd/client/session/AbstractClientSession.java
robertodeandrea: ok, I agree

sshd-core/src/main/java/org/apache/sshd/common/forward/DefaultForwardingFilter.java
robertodeandrea: ok; I agree

sshd-core/src/main/java/org/apache/sshd/common/forward/DefaultForwardingFilter.java
robertodeandrea: ok, I agree

sshd-core/src/main/java/org/apache/sshd/common/forward/DefaultForwardingFilter.java
robertodeandrea: ok,I agree

gnodet:
Merged

PR #62, opened 2018-07-03
SSH831 Request new getter methods on port forwarding channel classes

robertodeandrea:


sshd-core/src/main/java/org/apache/sshd/common/forward/TcpipClientChannel.java
```diff
@@ -94,20 +96,26 @@ public Type getTcpipChannelType() {
...
         InetSocketAddress dst;
+        InetSocketAddress loc;
         Type openType = getTcpipChannelType();
         switch (openType) {
             case Direct:
                 src = (InetSocketAddress) serverSession.getRemoteAddress();
                 dst = this.remote.toInetSocketAddress();
+                loc = (InetSocketAddress) serverSession.getLocalAddress();
+                tunnelEntrance = new SshdSocketAddress(loc.getHostString(), loc.getPort());
+                tunnelExit = new SshdSocketAddress(dst.getHostString(), dst.getPort());
```
lgoldstein: _loc_ can be defined locally only when needed:
```java
case Direct: {
    src = ...
    dst = ...
    InetSocketAddress loc = ....
    break;
}
```

sshd-core/src/main/java/org/apache/sshd/common/forward/TcpipClientChannel.java
```diff
@@ -166,4 +174,13 @@ protected synchronized void doWriteData(byte[] data, int off, long len) throws I
...
+    
+    public SshdSocketAddress getTunnelEntrance() {
+        return tunnelEntrance;
+    }
+  
+    public SshdSocketAddress getTunnelExit() {
+        return tunnelExit;
+    }
+
 }
```
lgoldstein: Superfluous empty line before end of class

sshd-core/src/main/java/org/apache/sshd/server/forward/TcpipServerChannel.java
```diff
@@ -107,6 +110,22 @@ public TcpipServerChannel(ForwardingFilter.Type type) {
         return type;
     }
 
+    public SshdSocketAddress getTunnelEntrance() {
+        return tunnelEntrance;
+    }
+
+    public SshdSocketAddress getTunnelExit() {
+        return tunnelExit;
+    }
```
lgoldstein: Should be made as some sort of "marker" interface since implemented by several entities:
```java
public interface TunnelingChannelEndpointsProvider {
    SshdSocketAddress getTunnelEntrance();

    SshdSocketAddress getTunnelExit();
}
```

sshd-core/src/main/java/org/apache/sshd/server/forward/TcpipServerChannel.java
```diff
@@ -133,6 +152,10 @@ protected OpenFuture doInit(Buffer buffer) {
             default:
                 throw new IllegalStateException("Unknown server channel type: " + channelType);
         }
+        
+        origIpAddress = originatorIpAddress;
+        tunnelEntrance = new SshdSocketAddress(hostToConnect, portToConnect);
+        tunnelExit = address;
```
lgoldstein: Why expose only originator address and not the port as well ?
```java

....
originatorAddress = new SshdSocketAddress(originatorIpAddress, originatorPort);
....

public SshdSocketAddress getOriginatorAddress() {
    return originatorAddress;
}
```

sshd-core/src/main/java/org/apache/sshd/server/forward/TcpipServerChannel.java
robertodeandrea: I agree, I will shortly commit the new code with your right suggestion

lgoldstein:
No need to make any more changes - I am merging the code shortly...

gnodet:
Merged

PR #65, opened 2018-07-31
SSHD-838 Remove LoggingUtils.logExceptionStackTrace.

bturner:
- Removed LoggingUtils.logExceptionStackTrace and its related helpers
- Updated call sites in Nio2Acceptor and Nio2Session to log exceptions by passing them directly to SLF4J

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/common/util/logging/LoggingUtils.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/util/logging/LoggingUtils.java
@@ -19,7 +19,6 @@
 
 package org.apache.sshd.common.util.logging;
 
-import java.io.IOException;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


sshd-core/src/main/java/org/apache/sshd/common/util/logging/LoggingUtils.java
```diff
@@ -555,108 +545,4 @@ public String toString() {
...
-     */
-    public static int getDefaultStackTraceLoggingDepth() {
-        return Math.max(DEFAULT_STACK_TRACE_LOGGING_DEPTH_VALUE_HOLDER.get(), DEFAULT_STACK_TRACE_LOGGING_DEPTH_VALUE);
-    }
-
-    /**
-     * @param value The value to set - <B>Note:</B> the effective value is the <U>maximum</U>
-     * between it and {@value #DEFAULT_STACK_TRACE_LOGGING_DEPTH_VALUE}
-     */
-    public static void setDefaultStackTraceLoggingDepth(int value) {
```
bturner: Perhaps all these utility methods can't actually be removed. I'm not sure how "API" they are. If they're public API, I'm happy to restore them all (and maybe mark them deprecated?) if you prefer. Just let me know.

lgoldstein:
Sorry, I don't think we will do this (see the issue comment) - better close the PR...

lgoldstein:
Merged - please close the PR

PR #67, opened 2018-10-01
[SSHD-846] Allow KeyPairGenerators to be garbage-collected

rovarga:
Since we do not use the KeyPairGenerator once we have generated the keypair, make sure we remove our reference so it can be garbage-collected. With Bouncy Castle that translates to around ~34KiB memory savings per session.


rovarga:
Closing as per @lgoldstein 's request.

PR #68, opened 2018-10-09
SSHD-848 Problem opening local port forwarding channel

robertodeandrea:
This is our proposed code change to handle problem reported in SSHD-848.
I'll wait your comments.

Regards
Roberto

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/common/forward/DefaultForwardingFilter.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/forward/DefaultForwardingFilter.java
@@ -971,7 +971,6 @@ public class DefaultForwardingFilter
                     OpenFuture openFuture = channel.getOpenFuture();
                     if (!openFuture.isDone()) {
                         openFuture.addListener(f -> {
-                            // If exception signaled then close channel immediately
                             channel.close(false);        
                         });
                     } else {
--- a/sshd-core/src/main/java/org/apache/sshd/common/forward/DefaultForwardingFilter.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/forward/DefaultForwardingFilter.java
@@ -957,15 +957,27 @@ public class DefaultForwardingFilter
 
         @Override
         public void sessionClosed(IoSession session) throws Exception {
-            TcpipClientChannel channel = (TcpipClientChannel) session.removeAttribute(TcpipClientChannel.class);
+            TcpipClientChannel channel = (TcpipClientChannel) session.removeAttribute(TcpipClientChannel.class);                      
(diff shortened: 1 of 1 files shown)


sshd-core/src/main/java/org/apache/sshd/common/forward/DefaultForwardingFilter.java
```diff
@@ -957,15 +957,27 @@ public void sessionCreated(IoSession session) throws Exception {
...
                     log.debug("sessionClosed({}) closing channel={} after {} messages - cause={}",
-                            session, channel, messagesCounter, (cause == null) ? null : cause.getClass().getSimpleName());
+                        session, channel, messagesCounter, (cause == null) ? null : cause.getClass().getSimpleName());
+                }
+              
+                if (cause != null) {
+                    channel.close(true);
+                } else {
+                    OpenFuture openFuture = channel.getOpenFuture();
+                    if (!openFuture.isDone()) {
```
lgoldstein: 1.  The remark is misplaced - should be over `channel.close(true)`
2.  I am not comfortable with the idea of delaying the channel closure until the open is finished. This is `sessionClosed` so why wait for the channel open to finish ?

robertodeandrea: 1. I agree. I removed the comment and made a new commit.
2. I do not agree. If you don't delay closing the channel until the channel is completely opened, you risk loosing all the data bytes sent by the client to the local port forwarding listening socket/tunnel. This is what happened in our test case.

lgoldstein: I am inclined to agree with you on (2) -  reviewing and testing...

lgoldstein:
I have reviewed the code and I think it should behave as follows:
```java
if (cause != null) {
    // If exception occurred close the channel immediately
    channel.close(true);
} else {
    /*
     *  Make sure channel is fully open in case the client was very fast
     *  and sent data + closed the connection before channel open was completed.
     */
     OpenFuture openFuture = channel.getOpenFuture();
      // If channel is established then listener is invoked immediately
     openFuture.addListener(f -> {
           Throwable err = f.getException();
           if (err != null) {
                 log.warn("sessionClosed({}) closing incompletely open channel={} after {} messages due to {}[{}]",
                      session, channel, messagesCounter, err.getClass().getSimpleName(), err.getMessage());
           }
          channel.close(err != null);
       });
 }
```
Can you test it and see if it works for your ?

robertodeandrea:
ok, I let you know if it works for us.
Thank you very much for your support.

robertodeandrea:
Your code works fine.
Can you merge it into Apache SSHD master git repository ?

Regards
Roberto

PR #69, opened 2018-10-10
[SSHD-847] Add sshd-osgi artifact shading sshd-common and sshd-core

rovarga:


rovarga:
This is a stab in the dark and does not quite work because it generates superfluous imports.

PR #70, opened 2018-10-15
SSHD-849

robertodeandrea:
This is our code change related to issue SSHD-849.
Can you give us your feedback ?

Thanks in advance
Roberto


sshd-core/src/main/java/org/apache/sshd/common/future/DefaultSshFuture.java
```diff
@@ -106,10 +106,9 @@ public void setValue(Object newValue) {
             }
 
             result = (newValue != null) ? newValue : GenericUtils.NULL;
+            notifyListeners();
```
lgoldstein: Notifying the listener while under lock is extremely dangerous. I also fail to see the problem and why this solves it...

sshd-core/src/main/java/org/apache/sshd/common/forward/DefaultForwardingFilter.java
```diff
@@ -957,15 +957,26 @@ public void sessionCreated(IoSession session) throws Exception {
...
+                } else {
+                    OpenFuture openFuture = channel.getOpenFuture();
+                    if (!openFuture.isDone()) {
+                        openFuture.addListener(f -> {
+                            channel.close(false);        
+                        });
+                    } else {
+                        channel.close(false);                            
+                    }                  
                 }
```
lgoldstein: This code is not rebased on latest version...

sshd-core/src/main/java/org/apache/sshd/common/future/DefaultSshFuture.java
robertodeandrea: The problem happens after the issue SSHD-848 was fixed. When a client session is closed listeners are added when the OpenFuture is done. If the listeners are NOT notified, the local port forwarding channel is closed whilst bytes are already flowing from client to the target server. This causes some bytes to be lost in transit. In the current code notifyListener() is called after a the synchronized block. Only with our code change we fixed the problem.
Do you understand what I mean ?

sshd-core/src/main/java/org/apache/sshd/common/forward/DefaultForwardingFilter.java
robertodeandrea: I agree, I changed only DefaultSshFuture.java for this issue.

sshd-core/src/main/java/org/apache/sshd/common/future/DefaultSshFuture.java
fcav1973: In order to better clarify the issue, what happens is that during execution of method `DefaultSshFuture.setValue`, after the synchronized block is executed _but before or during_ `notifyListeners`  is called, the method `addListener` is called (see `DefaultForwardingFilter.sessionClosed`) by another thread so that `notifyNow` is `true` and the listener is executed immediately
`
...
        if (notifyNow) {
            notifyListener(listener);
        }
...
`
Not respecting the ordering the listeners are added. In our case

`                        openFuture.addListener(f -> {
                            channel.close(false);        
                        });
`
is executed while the existing listeners are called inside `setValue` so that all packets after `channel.close(false)` are not sent

lgoldstein:
I don't understand why notifications are made under the lock - it is extremely dangerous, since we don't know what the listeners are doing. They might lock/wait for something and thus cause a domino effect where the entire future is locked. Please provide an explanation what it the race condition and why calling the listeners under the lock is mandatory (or are there alternatives).

lgoldstein:
I believe I see the issue - the `notifyListeners` method might work on the listeners field while it is being modified. This can be fixed in another way (working on it and will publish it soon for your review).

lgoldstein:
Further analysis does not reveal any problem - if result has been already set then the `listeners` field does not change any longer. Furthermore `notifyListeners` is called only when value is set for the **first** time. Therefore, I don't see the race condition.

fcav1973:
The race condition occurs if, during setValue inside notifyListeners, a
addListener is called.
In that case, while notifyListeners is looping through the listenerList,
addListener, since the is result is already set, calls notifyListener by
itself, which is executed immediatly.
So, the listeners order is not respected: while the listeners list is
looped for execution (setValue), the last one is executed (addListener),
instead of being appended at the end of the listeners list

Il giorno mer 17 ott 2018 alle ore 07:04 Lyor Goldstein <
notifications@github.com> ha scritto:

lgoldstein:
>>> So, the listeners order is not respected: while the listeners list is looped for execution ('setValue'), the last one is executed ('addListener'), instead of being appended at the end of the listeners list

This is by design - doing it the other way around it more complicated. Please note that relying on the **order** of the listeners is "brittle" - so unless there is a compelling reason to do so I would rather avoid it as it entails non-trivial changes to the code - way beyond what is in this PR.

fcav1973:
If the listener mechanism should not be considered reliable, than modification in https://github.com/apache/mina-sshd/pull/70/files/851b22fc431e4ed6ba440874f236d93a6aca9827#diff-30af04f3ecf0db9373c6e029a900f550

`                    OpenFuture openFuture = channel.getOpenFuture();
                    if (!openFuture.isDone()) {
                        openFuture.addListener(f -> {
                            channel.close(false);        
                        });
                    } else {
                        channel.close(false);                            
                    }                  `

does not work. In fact what we have seen is that while the 'notifyListeners' method `setValue` is executed, the 'channel.close(false);' is executed at the same time, with the result that the channel is closed and all bytes that should be transferred by the remaining listeners (those in the 'notifyListeners' at a certain point of the loop) are lost

lgoldstein:
I understand now the issue - we need to figure out a better mechanism - one that does not involve locking the entire future while listeners are being executed.

lgoldstein:
>> In fact what we have seen is that while the 'notifyListeners' method setValue is executed, the 'channel.close(false);' is executed at the same time

I don't see how this is possible, since `setValue` locks the future and only afterwards runs the notifications. During the lock, `addListener` cannot be invoked...

lgoldstein:
I have been looking at the code of `DefaultForwardingFilter` (lines 1078-1009) and have a better grasp of the problem:
```java
OpenFuture future = channel.getOpenFuture();
            if (future.isOpened()) {
                OutputStream outputStream = channel.getInvertedIn();
                outputStream.write(buffer.array(), buffer.rpos(), buffer.available());
                outputStream.flush();
            } else {
                future.addListener(f -> {
                    try {
                        OutputStream outputStream = channel.getInvertedIn();
                        outputStream.write(buffer.array(), buffer.rpos(), buffer.available());
                        outputStream.flush();
                    } catch (IOException e) {
                        try {
                            exceptionCaught(session, e);
                        } catch (Exception err) {
                            log.warn("messageReceived({}) failed ({}) to signal {}[{}] on channel={}: {}",
                                session, err.getClass().getSimpleName(), e.getClass().getSimpleName(),
                                e.getMessage(), channel, err.getMessage());
                        }
                    }
                });
            }
```
Perhaps we can somehow use register as listener a "pending messages queue" and have the future registered at `sessionClosed` wait until it has been emptied. I was wondering perhaps you could also think about a solution along these lines.

lgoldstein:
I have tentatively written https://github.com/apache/mina-sshd/pull/72 - please review (and test if you feel it is adequate). I am also interested in some stress testing if you can manage it since the code is more complex (more overhead per incoming message).

PR #72, opened 2018-10-18
[SSHD-849] Using a ClientChannelPendingMessagesQueue to regulate messages order while forwarding channel is being set up

lgoldstein:


lgoldstein:
The code can also be checked out and tested from https://github.com/lgoldstein/mina-sshd/tree/SSHD-849.

fcav1973:
We have tested you modifications, but we have found some problems, I thinks related to `queue.waitForPendingMessagesFlush(TimeUnit.SECONDS, maxWaitSec);` inside `DefaultForwardingFilter` which is blocking the thread pool. We have worked out a patch for this (something like adding another feature field to `ClientChannelPendingMessagesQueue` so to append the listener executing `channel.close`) and we are trying it (with good results till now). We will give you feedback in the next few days.

lgoldstein:
Good news - will await your input

robertodeandrea:
Hi Lyor, 
we made definitive changes to our code that is working fine. 
Our last commits are at https://github.com/robertodeandrea/mina-sshd/commit/c381e522dc254387009d82612dc15f0c584dece6

lgoldstein:
There are quite a few issues with this code - however, I think you are on to something with the usage of the extra future (the completion one). I have made the changes based on your idea into my original code and pushed it into the same branch (https://github.com/lgoldstein/mina-sshd/tree/SSHD-849) + PR. Can you review + test it an see ?

fcav1973:
Hi,
your modifications seems to be good. We're still testing with previous one
since we're now trying to overcome problems with SSH_MSG_CHANNEL_WINDOW_ADJUST
management that, since it's blocking, it's forcing us to use a thread poll
equal to max level of concurrent channels.
We'll test your version of the modification in the next days and give you a
feedback.


Il giorno mar 23 ott 2018 alle ore 17:37 Lyor Goldstein <
notifications@github.com> ha scritto:

lgoldstein:
Great - hope you can overcome the `SSH_MSG_CHANNEL_WINDOW_ADJUST` issue...

PR #73, opened 2018-10-23
[SSHD-851] Last chunk of data lost into local port forwarding channel

robertodeandrea:
This is my proposed code change to fix [SSHD-851](http://issues.apache.org/jira/browse/SSHD-851).


sshd-core/src/main/java/org/apache/sshd/server/forward/TcpipServerChannel.java
```diff
@@ -320,8 +322,14 @@ protected Closeable getInnerCloseable() {
...
                     protected CloseFuture doCloseGracefully() {
-                        executor.submit(() -> connector.close(false));
-                        return null;
+                        Future<CloseFuture> closeFeature = executor.submit(() -> connector.close(false));
+
+                        try {
+                            return closeFeature.get();
+                        } catch (InterruptedException | ExecutionException e) {
+                            log.error("CloseFuture execution completed with error ", e);
+                            return null;
```
lgoldstein: Does not seem to honor the **asynchronous** contract of this method - the whole idea of the `submit` call is **not** to wait for the connector to close. This is in contrast with `doCloseImmediately` which is **synchronous**.

tomaswolf:
Closing old outdated PRs.

PR #74, opened 2018-10-28
Add continuous integration

wh0:
This PR adds a continuous integration configuration, which I think would make it easier for contributors to test their changes.

About this PR:
- would use Travis CI. See here https://travis-ci.org/wh0/mina-sshd for example build output on my fork.
- Travis CI doesn't have IPv6 set up on their runners, so I've made it skip the IPv6-related client tests.
- One test is failing: `testLocalForwardingPayload(org.apache.sshd.common.forward.PortForwardingLoadTest)`. Could someone confirm that this test passes for them?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,19 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements. See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership. The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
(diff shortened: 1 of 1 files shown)


lgoldstein:
The initiative has merit, but the configuration is incomplete:

* Must make sure Java *8* is used
* Must test this on both _Windows_ and _Linux_
* The `PortForwardingLoadTest` passes for local builds (verified on _Windows_ and _Linux_) so must investigate why it fails on _Travis_

Furthermore, the project is not hosted on _Github_ - just mirrored on it, so we don't really have _Travis_ access...

wh0:
Thanks for bringing up these points. I didn't know about these requirements. I can't reply and simply fix all of those, but here's some more info if anyone's wondering:

* Project not on GitHub: https://github.com/travis-ci/travis-ci/issues/3087 Travis doesn't intend to support.
* Tests on Windows: https://github.com/travis-ci/travis-ci/issues/2104 Travis hasn't declared its intent to support, but is tracking technical issues that are currently blocking them from doing so.
* Java 8: https://docs.travis-ci.com/user/languages/java/#testing-against-multiple-jdks although it already looks like there are enough non-starters about using Travis, it happens that one can specify the version used to test.

lgoldstein:
>> I can't reply and simply fix all of those

You do get points for the initiative... - I agree with you that we do need some kind of continuous integration mechanism, though due to lack of time and resources we never seem to get around to it. I have created [SSHD-856](https://issues.apache.org/jira/browse/SSHD-856) to help us track this issue, but I have major doubts we will get to it without some outside contribution.

PR #75, opened 2018-10-28
Core: Make checkConfig public

wh0:
`checkConfig` is documented in the README as a reference for how to configure an SshClient or SshServer, but we can't call it because it's protected.

This method would be helpful to have for setting up some sensible defaults when you want to use this library for its protocol handling combined with custom connection management. You would be able to call `checkConfig` without, for example, the entire `start` method in SshServer.

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/client/SshClient.java
+++ b/sshd-core/src/main/java/org/apache/sshd/client/SshClient.java
@@ -321,7 +321,7 @@ public class SshClient extends AbstractFactoryManager implements ClientFactoryMa
     }
 
     @Override
-    protected void checkConfig() {
+    public void checkConfig() {
         super.checkConfig();
 
         Objects.requireNonNull(getForwarderFactory(), "ForwarderFactory not set");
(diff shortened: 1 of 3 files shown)


lgoldstein:
>>> This method would be helpful to have for setting up some sensible defaults when you want to use this library for its protocol handling combined with custom connection management. You would be able to call checkConfig without, for example, the entire start method in SshServer.

`checkConfig` simply completes missing configurations that the user may have chosen not to call explicitly (i.e., chose the defaults). The correct sequence (as demonstrated in the README file) is:

```java
SshServer sshd = SshServer.setupDefaultServer();
sshd.set...
sshd.set...
sshd.set...
sshd.start();l
```
i.e., the user should set up the client/server in whatever way is required and then call _start_ - bearing in mind the defaults. These defaults are described in the documentation + can be seen in the `checkConfig` code. I therefore, do not see the benefit in having `checkConfig` be public.

lgoldstein:
I am sorry, but I don't see enough value for this change - especially since `checkConfig` is a misnomer (it should be called `autoConfigureMissingConfigutations`). One may ask - why not do these in `setupDefaultServer` ? And the reason is clear - whatever the user did not override with `setXXX` calls will be auto-completed by the code when the client/server is started. I.e., we need to distinguish between non-initialized configuration and initialized one. By allowing external calls to `checkConfig` we would not eliminate the need to call it internally since we do need the configuration "auto-completion" feature it provides. Furthermore, the `checkConfig` code relies on some call sequence - which can be easily violated if we allow it to be called out-of-sequence.

wh0:
Ok, thanks for the timely feedback. It is a very niche use case not to call `start()` anyway, and it's easy enough to work around. I agree that it's not worth introducing the possibility of call sequence errors.

PR #78, opened 2018-10-28
Contrib: Add a security provider registrar for AndroidOpenSSL

wh0:
On Android 9, you can't use the built in Bouncy Castle security provider because Android blocks attempts to use Bouncy Castle for operations that AndroidOpenSSL supports https://developer.android.com/guide/topics/security/cryptography#bc-algorithms.

So here's a security provider registrar that hooks up AndroidOpenSSL.

### DIFF
--- a/sshd-contrib/src/main/java/org/apache/sshd/common/util/security/androidopenssl/AndroidOpenSSLSecurityProviderRegistrar.java
+++ b/sshd-contrib/src/main/java/org/apache/sshd/common/util/security/androidopenssl/AndroidOpenSSLSecurityProviderRegistrar.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


sshd-contrib/src/main/java/org/apache/sshd/common/util/security/androidopenssl/AndroidOpenSSLSecurityProviderRegistrar.java
```diff
@@ -0,0 +1,45 @@
...
+
+public class AndroidOpenSSLSecurityProviderRegistrar extends AbstractSecurityProviderRegistrar {
+    public AndroidOpenSSLSecurityProviderRegistrar() {
+        super("AndroidOpenSSL");
+    }
+
+    @Override
+    public boolean isSupported() {
+        return true;
+    }
```
lgoldstein: I would rather have `isSupported` check something that really shows in the security provider is supported - e.g., the existence of some class, system property, etc... - something that would prevent it from being "supported" if it is not running in an Android environment.

wh0: In fact there is such a system property. I've changed it to check for that. Thanks for the suggestion.

lgoldstein:
Merged in (see [this commit](https://github.com/apache/mina-sshd/commit/ec8ebfb6e284e432b701068b13466b1a781f39ce)) - pity you did not use some identity name/mail so that your contribution could be recognized. Please close the PR.

PR #81, opened 2018-12-10
[SSHD-708] Support reading encrypted OpenSSH private keys

tomaswolf:
Include the BCrypt pbkdf implementation from org.connectbot.jbcrypt
and use it to decrypt password-protected OpenSSH private keys.

Add test cases.


sshd-common/src/main/java/org/apache/sshd/common/config/keys/loader/openssh/OpenSSHKeyPairResourceParser.java
```diff
@@ -18,11 +18,14 @@
  */
 package org.apache.sshd.common.config.keys.loader.openssh;
 
+import static java.text.MessageFormat.format;
```
lgoldstein: We do not allow static imports - the code does not seem to be compiled (which would have made `checkstyle` complain).

sshd-common/src/main/java/org/apache/sshd/common/config/keys/loader/openssh/OpenSSHKeyPairResourceParser.java
```diff
@@ -170,6 +170,66 @@ protected PublicKey readPublicKey(
...
+                return readPrivateKeys(session, resourceKey, context, publicKeys, passwordProvider, stream);
+            }
+        }
+        // Encrypted keys
+        if (passwordProvider == null) {
+            throw new StreamCorruptedException(format("No password provider for encrypted key {0}", resourceKey.getName()));
+        }
+        if (!context.validate(data)) {
+            throw new StreamCorruptedException(format("Invalid data in encrypted key {0}", resourceKey.getName()));
+        }
```
lgoldstein: Please don't use `MessageTextFormat#format` - there is no upside in this case...

sshd-common/src/main/java/org/apache/sshd/common/config/keys/loader/openssh/OpenSSHKeyPairResourceParser.java
```diff
@@ -182,9 +242,8 @@ protected PublicKey readPublicKey(
         boolean traceEnabled = log.isTraceEnabled();
         int check1 = KeyEntryResolver.decodeInt(stream);
         int check2 = KeyEntryResolver.decodeInt(stream);
-        if (traceEnabled) {
-            log.trace("readPrivateKeys({}) check1=0x{}, check2=0x{}",
-                resourceKey, Integer.toHexString(check1), Integer.toHexString(check2));
+        if (check1 != check2) {
+            throw new StreamCorruptedException(format("Check number mismatch in key {0}", resourceKey.getName()));
         }
```
lgoldstein: Please don't use `MessageTextFormat#format` - there is no upside in this case...

sshd-common/src/main/java/org/apache/sshd/common/config/keys/loader/openssh/OpenSSHParserContext.java
```diff
@@ -19,10 +19,27 @@
 
 package org.apache.sshd.common.config.keys.loader.openssh;
 
+import static java.text.MessageFormat.format;
```
lgoldstein: See previous remark regarding static imports

sshd-common/src/main/java/org/apache/sshd/common/config/keys/loader/openssh/OpenSSHKeyPairResourceParser.java
tomaswolf: The code _is_ compiled (and run -- of course I tested that the new tests work). But I don't see any checkstyle warnings in Eclipse. Probably would need to install some checkstyle plugin...

PR #82, opened 2018-12-12
[SSHD-708] Fixed support for OpenSSH encrypted private keys decoding when 'bcrypt' KDF is used

lgoldstein:
* Increased the limit on `MAX_ROUNDS`
* Made the limitation configurable via system property
* Updated `README.md` file about the `bcrypt` settings and behavior regarding the `MAX_ROUNDS`
* Removed dangerous constructor/initializer of the KDF options

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,22 +247,51 @@ detect and use the identity files residing in the user's *~/.ssh* folder (e.g., 
 **Note:** if the identity files are encrypted via a password, one must configure a `FilePasswordProvider` so that the code can decrypt them before using
 and presenting them to the server as part of the authentication process. Reading key files in PEM format (including encrypted ones) is supported by default
 for the standard keys and formats. Using additional non-standard special features requires that the [Bouncy Castle](https://www.bouncycastle.org/) supporting
-artifacts be available in the code's classpath. One can also read files in
-[OpenSSH](http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/ssh/PROTOCOL.key?rev=1.1&content-type=text/x-cvsweb-markup) format without any specific extra
-artifacts (although for reading _ed25519_ keys one needs to add the _EdDSA_ support artifacts). **Note:** for the time being, password
-encrypted _ed25519_ private key files are not supported.
+artifacts be available in the code's classpath.
 
-The `FilePasswordProvider` has support for a retry mechanism via its `handleDecodeAttemptResult`. When the code detects an encrypted private key,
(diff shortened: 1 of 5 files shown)


lgoldstein:
* Added lazy evaluation of the system property - NOTE: if bad system property value is provided, then it is silently ignored and default is used.
* Added programmatic API to set the max. allowed rounds at any time (also unit tested it)

tomaswolf:
This works, so OK from my side.

However, I'd consider simplifying this. Now that there is a setter for this MAX_ROUNDS value:

1.  It wouldn't be necessary to have a system property. If an application using the library wants to have a system property, it can implement that bit itself and then use the provided setter.
2. It then is not necessary to synchronize anymore.

Simply
```
private static final AtomicInteger MAX_ROUNDS = new AtomicInteger(DEFAULT_MAX_ROUNDS);

public static void setMaxAllowedRounds(int value) {
  ValidateUtils.checkTrue(value > 0, "Invalid max. rounds value: %d", value);
  MAX_ROUNDS.set(value);
}

public static int getMaxAllowedRounds() {
  return MAX_ROUNDS.get();
}
```
would be good enough.

In fact, this can be done even without `AtomicInteger`. References are read and written atomically anyway in Java, so the following also works and is thread-safe:

```
private static Integer MAX_ROUNDS = Integer.valueOf(DEFAULT_MAX_ROUNDS);

public static void setMaxAllowedRounds(int value) {
  ValidateUtils.checkTrue(value > 0, "Invalid max. rounds value: %d", value);
  MAX_ROUNDS = Integer.valueOf(value);
}

public static int getMaxAllowedRounds() {
  Integer value = MAX_ROUNDS;
  return value.intValue();
}
```

Less code == less opportunity for bugs. And less synchronized.

> from what I read it involves a considerable load on the CPU

16 or 33 doesn't make much difference. It won't melt the cores. Did you try it? On my machine (which isn't specially powerful) it just about doubles the running time for the test loading the encrypted ed25519 key from 0.2 to 0.4 sec. (As expected. It's a linear number of rounds after all.)

BTW, DSA is always 1024 bits.

lgoldstein:
>> It wouldn't be necessary to have a system property. If an application using the library wants to have a system property, it can implement that bit itself and then use the provided setter.

Agreed.

>> 16 or 33 doesn't make much difference. It won't melt the cores. Did you try it? 

I see no reason to - once we have this in place it is up to each implementation to decide on its max. acceptable rounds value

lgoldstein:
All done (also removed the system property and updated the documentation).

PR #84, opened 2018-12-16
Capture test coverage using jacoco and add jacoco, index and javadoc site reports

msohn:


lgoldstein:
Nice work, however I would like to have this as a Maven profile - e.g. `mvn -Pjacoco clean install` - i.e., not sure I want to run _jacoco_ every time I run a build.

pom.xml
```diff
@@ -1097,6 +1097,7 @@
                             <goal>clean</goal>
                         </goals>
                         <configuration>
+                            <excludeDefaultDirectories>true</excludeDefaultDirectories>
```
lgoldstein: Leaving `target` folder as-is is dangerous - when I clean something I want it to be clean.

pom.xml
```diff
@@ -1271,6 +1272,7 @@
                         <java.awt.headless>true</java.awt.headless>
                         <org.slf4j.simpleLogger.logFile>System.out</org.slf4j.simpleLogger.logFile>
                     </systemPropertyVariables>
+                    <argLine>@{argLine}</argLine>
```
lgoldstein: Shouldn't this be `$(argLine}` ?

pom.xml
```diff
@@ -1315,16 +1362,79 @@
...
+                            <report>index</report>
+                        </reports>
+                    </reportSet>
+                </reportSets>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-javadoc-plugin</artifactId>
+                <version>2.9</version>
+                <configuration>
```
lgoldstein: - We already have a definition for the plugin (and with a more advanced version)
- I don't want the plugin to run as part of the normal build cycle
- I am not sure we want the tests javadoc
- The `javase/8` is hardcoded - should be derived from one of the existing properties (e.g., `javac.source` or similar)

sshd-osgi/pom.xml
```diff
@@ -84,7 +84,7 @@
                 <executions>
                     <execution>
                         <id>unpack-dependencies</id>
-                        <phase>compile</phase>
+                        <phase>package</phase>
```
lgoldstein: Have you tested it and made 100% sure that it works as before ? I seem to remember wanting to do this, but something was wrong and only the `compile` phase fixed it. If it is not absolutely necessary for `jacoco` I would rather not change it, otherwise please make 100% sure that the normal build cycle (i.e., `mvn clean install`) creates a viable OSGi artifact (as it currently does).

pom.xml
```diff
@@ -1315,16 +1362,79 @@
...
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jxr-plugin</artifactId>
+                <version>3.0.0</version>
+            </plugin>
+            <plugin>
+                <groupId>org.jacoco</groupId>
+                <artifactId>jacoco-maven-plugin</artifactId>
+                <reportSets>
+                    <reportSet>
```
lgoldstein: As previously mentioned, I don't want `jacoco` to be part of the regular build cycle, but rather activated via a profile (`-Pjacoco`)

PR #86, opened 2019-01-09
[bugfix] fix transfer large file bug(more than1GB)

feixiangff:
fix a bug when using mina sshd to transfer a large source file(more than 1GB) by scp(scp username@remote:/largefile ./)
because if the source file is larger than 1GB, the function NumberUtils.getNextPowerOf2(int) can not compute a correct value for new cache and in endless loop (the value j will be negative and in next round, j will be 0, and the function comes to endless loop)

### DIFF
--- a/sshd-common/src/main/java/org/apache/sshd/common/util/buffer/ByteArrayBuffer.java
+++ b/sshd-common/src/main/java/org/apache/sshd/common/util/buffer/ByteArrayBuffer.java
@@ -206,14 +206,29 @@ public class ByteArrayBuffer extends Buffer {
         int curPos = wpos();
         int remaining = maxSize - curPos;
         if (remaining < capacity) {
-            int minimum = curPos + capacity;
-            int actual = growthFactor.applyAsInt(minimum);
-            if (actual < minimum) {
-                throw new IllegalStateException("ensureCapacity(" + capacity + ") actual (" + actual + ") below min. (" + minimum + ")");
+            if (maxSize < MAX_LEN) {
+                //increase data by growthFactor
+                int minimum = curPos + capacity;
+                int actual = growthFactor.applyAsInt(minimum);
+                if (actual < minimum) {
+                    throw new IllegalStateException("ensureCapacity(" + capacity + ") actual (" + actual + ") below min. (" + minimum + ")");
+                }
+                byte[] tmp = new byte[actual];
+                System.arraycopy(data, 0, tmp, 0, data.length);
+                data = tmp;
+            } else {
+                //if data increased to max size, compact
+                compact();
+                int maxSizeAfterCompact = size();   //shoul same as maxSize
+                int curPosAfterCompact = wpos();
+                int remainingAfterCompact = maxSizeAfterCompact - curPosAfterCompact;
+                if (remainingAfterCompact < capacity) {
+                    //after compact, if remaining still less than capacity, throws exception
+                    throw new IllegalStateException("remaining still less than capacity after compact, please decrease capacity, remainingAfterCompact(" + remainingAfterCompact
+                            + "), capacity(" + capacity + "), max buffer size(" + maxSizeAfterCompact + ")");
+                }
             }
-            byte[] tmp = new byte[actual];
-            System.arraycopy(data, 0, tmp, 0, data.length);
-            data = tmp;
+
         }
     }
 


lgoldstein:
You are right about `getNextPowerOf2` but is this indeed an issue for SCP ? FYI, a file is being transferred in chunks - never as a single packet. Have you actually encountered the bug you are describing ? If so, please provide the stack trace of the call that causes `getNextPowerOf2` to overflow as described. I have tested the SSHD code with a 1.7GB file and encountered no problems.

lgoldstein:
After reviewing the suggested fix there is another reason I am reluctant to merge it - it uses `compact` which potentially changes the `wpos`. This may prove an issue since some code remembers the `wpos` and then returns to it in order to re-write some data (e.g., when some buffer length is not known ahead of time). Therefore, unless you can point to an actual problem that occurs in an SSH session (even a malicious one) I don't think we can merge this code. I did add code that makes sure that `nextPowerOf2` does not yield negative values so it will protect the code from the issue you described that may yield an endless loop.

feixiangff:
Hello, here is the case on my computer

OS:MAC OS 10.10.5(Yosemite)
JDK:1.8.0_121
SSHD:org.apache.sshd:sshd-core:2.1.0

How I start sshd server:
`

    private SshServer sshServer;
    public SshDaemon(@Value("${git.server.ip}") String gitServerIP, @Value("${git.server.port}") String gitServerPort,
                     @Value("${inet.socket.host}") String inetSocketHost, @Value("${inet.socket.port}") Integer inetSocketPort,
                     @Value("${host.key.pem}") String hostKeyPEM) {
        InetSocketAddress addr = new InetSocketAddress(inetSocketHost,inetSocketPort);
        sshServer = SshServer.setUpDefaultServer();
        sshServer.setPort(addr.getPort());
        sshServer.setHost(addr.getHostName());

        sshServer.setKeyPairProvider(new SimpleGeneratorHostKeyProvider());
        sshServer.setPublickeyAuthenticator(AcceptAllPublickeyAuthenticator.INSTANCE);
        sshServer.setCommandFactory(ProcessShellCommandFactory.INSTANCE);    //The key point I think in this case


        try {
            sshServer.start();
        } catch (IOException e) {
            e.printStackTrace();
        }
    }
`

SCP command:
scp -P 29418 -r username@localhost:/Users/username/Desktop/ubuntu-15.10-desktop-amd64.iso ./

The size of file is 1178386432(1.1GB)

The stacktrace in endless loop

	at org.apache.sshd.common.util.NumberUtils.getNextPowerOf2(NumberUtils.java:72)
	at org.apache.sshd.common.util.buffer.BufferUtils.getNextPowerOf2(BufferUtils.java:505)
	at org.apache.sshd.common.util.buffer.ByteArrayBuffer.ensureCapacity(ByteArrayBuffer.java:207)
	at org.apache.sshd.common.util.buffer.Buffer.ensureCapacity(Buffer.java:779)
	at org.apache.sshd.common.util.buffer.ByteArrayBuffer.putRawBytes(ByteArrayBuffer.java:167)
	at org.apache.sshd.server.shell.TtyFilterInputStream.read(TtyFilterInputStream.java:158)
	at java.io.FilterInputStream.read(FilterInputStream.java:107)
	at org.apache.sshd.server.shell.InvertedShellWrapper.pumpStream(InvertedShellWrapper.java:246)
	at org.apache.sshd.server.shell.InvertedShellWrapper.pumpStreams(InvertedShellWrapper.java:202)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)


Some var values in stack:
`

	at org.apache.sshd.common.util.NumberUtils.getNextPowerOf2(NumberUtils.java:72)

        value = 1073741872
        j = 0

`

`
//The key point I think in this case

	at org.apache.sshd.common.util.buffer.ByteArrayBuffer.ensureCapacity(ByteArrayBuffer.java:207)

        maxSize = 1073741824
        curPos = 1073733680
        remaining = 8144
        minimum = 1073741872
        can not compute var 'actual' since getNextPowerOf2 is in endless loop

`

It seems that this is a bug in ByteArrayBuffer(or TtyFilterInputStream), since ByteArrayBuffer seems to cache the whole file(even the file is very large) in it's buffer, and never compact the buffer

gnodet:
I haven't looked at the compact question, but the `getNextPowerOf2` should be fixed anyway.  I suggest the following:
```
diff --git a/sshd-common/src/main/java/org/apache/sshd/common/util/NumberUtils.java b/sshd-common/src/main/java/org/apache/sshd/common/util/NumberUtils.java
index 113e74e0..7c71a3dd 100644
--- a/sshd-common/src/main/java/org/apache/sshd/common/util/NumberUtils.java
+++ b/sshd-common/src/main/java/org/apache/sshd/common/util/NumberUtils.java
@@ -60,7 +60,7 @@ public final class NumberUtils {
 
     public static long getNextPowerOf2(long value) {
         long j = 1L;
-        while (j < value) {
+        while (j < value && j > 0) {
             j <<= 1;
         }
         return j;
@@ -68,7 +68,7 @@ public final class NumberUtils {
 
     public static int getNextPowerOf2(int value) {
         int j = 1;
-        while (j < value) {
+        while (j < value && j > 0) {
             j <<= 1;
         }
         return j;
diff --git a/sshd-common/src/main/java/org/apache/sshd/common/util/buffer/BufferUtils.java b/sshd-common/src/main/java/org/apache/sshd/common/util/buffer/BufferUtils.java
index 7fff5d0c..31387db4 100644
--- a/sshd-common/src/main/java/org/apache/sshd/common/util/buffer/BufferUtils.java
+++ b/sshd-common/src/main/java/org/apache/sshd/common/util/buffer/BufferUtils.java
@@ -518,7 +518,11 @@ public final class BufferUtils {
 
     public static int getNextPowerOf2(int value) {
         // for 0-7 return 8
-        return (value < Byte.SIZE) ? Byte.SIZE : NumberUtils.getNextPowerOf2(value);
+        return (value < Byte.SIZE)
+                ? Byte.SIZE
+                : (value > (1 << 30))
+                    ? value
+                    : NumberUtils.getNextPowerOf2(value);
     }
 
     /**
diff --git a/sshd-common/src/test/java/org/apache/sshd/common/util/NumberUtilsTest.java b/sshd-common/src/test/java/org/apache/sshd/common/util/NumberUtilsTest.java
index d7c2d320..9ec0937f 100644
--- a/sshd-common/src/test/java/org/apache/sshd/common/util/NumberUtilsTest.java
+++ b/sshd-common/src/test/java/org/apache/sshd/common/util/NumberUtilsTest.java
@@ -19,6 +19,7 @@
 
 package org.apache.sshd.common.util;
 
+import org.apache.sshd.common.util.buffer.ByteArrayBuffer;
 import org.apache.sshd.util.test.JUnitTestSupport;
 import org.apache.sshd.util.test.NoIoTestCase;
 import org.junit.FixMethodOrder;
@@ -60,6 +61,11 @@ public class NumberUtilsTest extends JUnitTestSupport {
         }
     }
 
+    @Test
+    public void testNextPowerOf2Max() {
+        assertTrue(NumberUtils.getNextPowerOf2(1073741872) < 0);
+    }
+
     @Test
     public void testToInteger() {
         assertNull("Unexpected null value", NumberUtils.toInteger(null));
```

gnodet:
You're right.
Here's a more complete patch.
I agree with Lyor than using `compact()` internally may not be wise as it has side effects that could be problematic.  The real problem is rather in `TtyFilterInputStream` which has an ever-growing internal buffer.

```
diff --git a/sshd-common/src/main/java/org/apache/sshd/common/util/NumberUtils.java b/sshd-common/src/main/java/org/apache/sshd/common/util/NumberUtils.java
index 113e74e0..7c71a3dd 100644
--- a/sshd-common/src/main/java/org/apache/sshd/common/util/NumberUtils.java
+++ b/sshd-common/src/main/java/org/apache/sshd/common/util/NumberUtils.java
@@ -60,7 +60,7 @@ public final class NumberUtils {
 
     public static long getNextPowerOf2(long value) {
         long j = 1L;
-        while (j < value) {
+        while (j < value && j > 0) {
             j <<= 1;
         }
         return j;
@@ -68,7 +68,7 @@ public final class NumberUtils {
 
     public static int getNextPowerOf2(int value) {
         int j = 1;
-        while (j < value) {
+        while (j < value && j > 0) {
             j <<= 1;
         }
         return j;
diff --git a/sshd-common/src/main/java/org/apache/sshd/common/util/buffer/BufferUtils.java b/sshd-common/src/main/java/org/apache/sshd/common/util/buffer/BufferUtils.java
index 7fff5d0c..31387db4 100644
--- a/sshd-common/src/main/java/org/apache/sshd/common/util/buffer/BufferUtils.java
+++ b/sshd-common/src/main/java/org/apache/sshd/common/util/buffer/BufferUtils.java
@@ -518,7 +518,11 @@ public final class BufferUtils {
 
     public static int getNextPowerOf2(int value) {
         // for 0-7 return 8
-        return (value < Byte.SIZE) ? Byte.SIZE : NumberUtils.getNextPowerOf2(value);
+        return (value < Byte.SIZE)
+                ? Byte.SIZE
+                : (value > (1 << 30))
+                    ? value
+                    : NumberUtils.getNextPowerOf2(value);
     }
 
     /**
diff --git a/sshd-common/src/test/java/org/apache/sshd/common/util/NumberUtilsTest.java b/sshd-common/src/test/java/org/apache/sshd/common/util/NumberUtilsTest.java
index d7c2d320..9ec0937f 100644
--- a/sshd-common/src/test/java/org/apache/sshd/common/util/NumberUtilsTest.java
+++ b/sshd-common/src/test/java/org/apache/sshd/common/util/NumberUtilsTest.java
@@ -19,6 +19,7 @@
 
 package org.apache.sshd.common.util;
 
+import org.apache.sshd.common.util.buffer.ByteArrayBuffer;
 import org.apache.sshd.util.test.JUnitTestSupport;
 import org.apache.sshd.util.test.NoIoTestCase;
 import org.junit.FixMethodOrder;
@@ -60,6 +61,11 @@ public class NumberUtilsTest extends JUnitTestSupport {
         }
     }
 
+    @Test
+    public void testNextPowerOf2Max() {
+        assertTrue(NumberUtils.getNextPowerOf2(1073741872) < 0);
+    }
+
     @Test
     public void testToInteger() {
         assertNull("Unexpected null value", NumberUtils.toInteger(null));
diff --git a/sshd-core/src/main/java/org/apache/sshd/server/shell/TtyFilterInputStream.java b/sshd-core/src/main/java/org/apache/sshd/server/shell/TtyFilterInputStream.java
index 9cde926d..b8948056 100644
--- a/sshd-core/src/main/java/org/apache/sshd/server/shell/TtyFilterInputStream.java
+++ b/sshd-core/src/main/java/org/apache/sshd/server/shell/TtyFilterInputStream.java
@@ -151,6 +151,7 @@ public class TtyFilterInputStream extends FilterInputStream {
         }
 
         if (buffer.available() == 0) {
+            buffer.compact();
             int nb = this.in.read(b, off, len);
             if (nb == -1) {
                 return nb;
diff --git a/sshd-core/src/test/java/org/apache/sshd/server/shell/TtyFilterInputStreamTest.java b/sshd-core/src/test/java/org/apache/sshd/server/shell/TtyFilterInputStreamTest.java
index a0226c8d..b5b9a9dd 100644
--- a/sshd-core/src/test/java/org/apache/sshd/server/shell/TtyFilterInputStreamTest.java
+++ b/sshd-core/src/test/java/org/apache/sshd/server/shell/TtyFilterInputStreamTest.java
@@ -21,7 +21,9 @@ package org.apache.sshd.server.shell;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.OutputStream;
+import java.lang.reflect.Field;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Collection;
@@ -36,6 +38,7 @@ import java.util.stream.Stream;
 
 import org.apache.sshd.common.channel.PtyMode;
 import org.apache.sshd.common.util.GenericUtils;
+import org.apache.sshd.common.util.buffer.ByteArrayBuffer;
 import org.apache.sshd.common.util.io.IoUtils;
 import org.apache.sshd.util.test.BaseTestSupport;
 import org.apache.sshd.util.test.JUnit4ClassRunnerWithParametersFactory;
@@ -102,6 +105,34 @@ public class TtyFilterInputStreamTest extends BaseTestSupport {
         }
     }
 
+    @Test
+    public void testInternalBufferSizeDoesNotGrow() throws Exception {
+        try (TtyFilterInputStream is = new TtyFilterInputStream(new InputStream() {
+            int next = 0;
+            @Override
+            public int read() {
+                next = (next + 1) & 0xFF;
+                return next;
+            }
+        }, EnumSet.of(mode))) {
+            Field f = is.getClass().getDeclaredField("buffer");
+            f.setAccessible(true);
+            ByteArrayBuffer buffer = (ByteArrayBuffer) f.get(is);
+
+            byte[] b = new byte[256];
+            for (int i = 0 ; i < 10; i++) {
+                is.read(b, 0, b.length);
+            }
+
+            int size = buffer.capacity();
+
+            for (int i = 0 ; i < 10; i++) {
+                is.read(b, 0, b.length);
+            }
+            assertEquals(size, buffer.capacity());
+        }
+    }
+
     private static void assertCRLFCounts(PtyMode mode, int numLines, int crCount, int lfCount) {
         switch(mode) {
             case ECHO:
```

PR #87, opened 2019-02-09
clear response buffer with unsupported extension

thefourtheye:
As it is, if a client requests to use an extension which is not
supported by the server, the server is supposed to send
`SSH_FX_OP_UNSUPPORTED`. It does send that, but it doesn't clear the
`Buffer` object before sending the response. So, all the clients who
are requesting an unsupported extension are impacted with invalid
bytestream of response data and the channel becomes unusable.

This patch simply clears the buffer with `prepareReply` function before
sending the response.

---

cc @The-Yoda @sramki @lgoldstein

### DIFF
--- a/sshd-sftp/src/main/java/org/apache/sshd/server/subsystem/sftp/AbstractSftpSubsystemHelper.java
+++ b/sshd-sftp/src/main/java/org/apache/sshd/server/subsystem/sftp/AbstractSftpSubsystemHelper.java
@@ -497,7 +497,7 @@ public abstract class AbstractSftpSubsystemHelper
             return;
         }
 
-        sendHandle(prepareReply(buffer), id, handle);
+        sendHandle(buffer, id, handle);
     }
 
     /**
@@ -707,7 +707,7 @@ public abstract class AbstractSftpSubsystemHelper
             return;
         }
 
-        sendHandle(prepareReply(buffer), id, handle);
+        sendHandle(buffer, id, handle);
     }
 
     protected abstract String doOpenDir(int id, String path, Path p, LinkOption... options) throws IOException;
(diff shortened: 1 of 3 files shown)


lgoldstein:
Good catch, but I would rather not move `prepareReply` to be part of `sendStatus` (LSRP) - whoever is invoking `sendStatus` should provide a valid buffer

thefourtheye:
@lgoldstein Wouldn't this make the life of the developers a bit easier?

1. They don't have worry about clearing the buffer everywhere. Less code repetition.
2. If somebody forgets to call `send*` functions without clearing the buffer, it would break. We have to rely only on the code review process to find this.

Those refactoring commits are just to make sure that we don't end up with a problem like this again. I am fine with or without it.

---

PS. I went looking for the expansion of LSRP. Couldn't find a satisfactory one. Now I am curious 😜

lgoldstein:
LSRP - Least Single Responsibility Principle.

>> Wouldn't this make the life of the developers a bit easier?

I don't think so - my concern is not code repetition but rather unexpected code behavior - whoever invokes `sendStatus` provides a buffer that may or may not be already initialized somehow - for whatever purpose (maybe some custom feature). I would rather `sendStatus` not tinker with the input buffer beyond appending the status message parameters.

Anyway, I have made a few minor changes to your code (many thanks) - you can see them [here](https://github.com/lgoldstein/mina-sshd/commit/5f49088fb7abf464ee8b2b521235340a10b70ec3). They will be merged into the _master_ version once we release 2.2.0 (which unfortunately means the fix will have to wait until we release the next version)

thefourtheye:
Awesome. Thank you! Now that the commit is in your fork, can we close this PR?

lgoldstein:
Yes you can - again, thanks for your contribution

thefourtheye:
@lgoldstein Thanks for your help with this fix 🙂

PR #89, opened 2019-02-11
[SSHD-893] Fix SCP download with pattern issue in rooted filesystem

lgoldstein:


sshd-common/src/main/java/org/apache/sshd/common/util/io/DirectoryScanner.java
```diff
@@ -244,46 +257,35 @@ public void setIncludes(String[] includes) {
...
-                    scandir(file, name + File.separator);
-                }
-            } else if (file.isFile()) {
-                if (isIncluded(name)) {
-                    filesIncluded.add(name);
+    protected <C extends Collection<Path>> C scandir(Path rootDir, Path dir, C filesList) throws IOException {
+        try (DirectoryStream<Path> ds = Files.newDirectoryStream(dir)) {
+            for (Path p : ds) {
+                Path rel = rootDir.relativize(p);
+                String name = rel.toString();
```
The-Yoda: `rootDir.relativize(p).toString()` should do?

The-Yoda:
LGTM

sshd-common/src/main/java/org/apache/sshd/common/util/io/DirectoryScanner.java
```diff
@@ -115,49 +126,37 @@
...
+    private List<String> includePatterns;
 
     /**
-     * Whether or not the file system should be treated as a case sensitive
-     * one.
+     * Whether or not the file system should be treated as
+     * a case sensitive one.
      */
-    protected boolean isCaseSensitive = true;
+    private boolean caseSensitive = OsUtils.isUNIX();
```
thefourtheye: I am not sure if this would be true entirely. Apple's HFS is not case-sensitive. `isUNIX` just checks if it is Windows or not.

sshd-common/src/main/java/org/apache/sshd/common/util/io/DirectoryScanner.java
```diff
@@ -188,21 +187,36 @@ public File getBasedir() {
...
-        }
+    public void setIncludes(String... includes) {
+        setIncludes(GenericUtils.isEmpty(includes) ? Collections.emptyList() : Arrays.asList(includes));
+    }
+
+    /**
+     * @return Un-modifiable list of the inclusion patterns
+     */
+    public List<String> getIncludes() {
+        return includePatterns;
```
thefourtheye: The comment above effectively says "returns an unmodifiable list". But if somebody casts the returned list to an `ArrayList`, it would be modifiable, right? Perhaps a [`Collections.unmodifiableList`](https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#unmodifiableList(java.util.List)) is in order?

sshd-common/src/main/java/org/apache/sshd/common/util/io/DirectoryScanner.java
lgoldstein: The code is geared towards the 99.99% - if someone is running on top of HFS then they can override the case sensitivity

sshd-common/src/main/java/org/apache/sshd/common/util/io/DirectoryScanner.java
lgoldstein: Look closely at the code an you will see the `includePatterns` is **always** unmodifiable...

sshd-common/src/main/java/org/apache/sshd/common/util/io/DirectoryScanner.java
lgoldstein: 1. Cascaded calls are not more efficient than separate ones
2. Separating calls makes the code more debug-able...
.

sshd-common/src/main/java/org/apache/sshd/common/util/io/DirectoryScanner.java
thefourtheye: Gotcha. Thanks.

sshd-common/src/main/java/org/apache/sshd/common/util/io/DirectoryScanner.java
thefourtheye: I am trying to understand this better. You mean, overriding `caseSensitive` variable? This patch makes that `private`. Is that intentional?

Also, `DirectoryScanner` is a standalone class, it's not implementing an interface or extending a class. So, if someone has to override this variable and use, they have to override `ScpFileOpener#getMatchingFilesToSend` as well with the modified `DirectoryScanner` implementation. Is my understanding correct?

lgoldstein: I mean override `ScpFileOpener` and implement `ScpFileOpener#getMatchingFilesToSend` using a `DirectoryScanner` with some other case-sensitivity (NOTE: `DirectoryScanner#setCaseSensitivity` is available to override the default). This can be easily done when initializing the SCP client/server by setting a default `ScpFileOpener` when they are configured - thus no need to override the opener on each session.

PR #90, opened 2019-02-12
[SSHD-894] Fix Race condition when doing async auth.

The-Yoda:
Currently, when doing async auth, the response is being sent to client
before starting the service with auth status set. When the client
responds back with next request, the request will reach auth service and
fail with `No current authentication mechanism for cmd=SSH_MSG_CHANNEL_OPEN`.
This happens mostly when the system starts (as, service initialization
might take time)

This patch fixes the issue by changing the order and send the packet
after the service starts.

cc @lgoldstein @thefourtheye @sramki

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/server/session/ServerUserAuthService.java
+++ b/sshd-core/src/main/java/org/apache/sshd/server/session/ServerUserAuthService.java
@@ -339,11 +339,13 @@ public class ServerUserAuthService extends AbstractCloseable implements Service,
                 sendWelcomeBanner(session);
             }
 
-            Buffer response = session.createBuffer(SshConstants.SSH_MSG_USERAUTH_SUCCESS, Byte.SIZE);
-            session.writePacket(response);
             session.setUsername(username);
             session.setAuthenticated();
             session.startService(authService, buffer);
+
+            buffer = session.createBuffer(SshConstants.SSH_MSG_USERAUTH_SUCCESS, Byte.SIZE);
+            session.writePacket(buffer);
+
             session.resetIdleTimeout();
             log.info("Session {}@{} authenticated", username, session.getIoSession().getRemoteAddress());
         } else {


lgoldstein:
Looks fine - will merge it as-is - have you run the full `mvn clean install` on the project to make sure ?

The-Yoda:
I did run the tests. I could see failures as below (even without changes).
```[ERROR] Errors:
[ERROR] org.apache.sshd.common.forward.ApacheServerApacheClientTest.testHasLocalPortForwardingStarted(org.apache.sshd.common.forward.ApacheServerApacheClientTest)
[ERROR]   Run 1: ApacheServerApacheClientTest.createClient:82 » IO java.lang.IllegalArgumentExc...
[ERROR]   Run 2: ApacheServerApacheClientTest.stopClient:93 NullPointer```

Anything I need to do specifically for this?

lgoldstein:
If it is intermittent, then no - however, my build fails consistently on `CompressionTest.testCompression` - please rebase on latest master and check again.

The-Yoda:
It is not intermittent. It is failing consistently. I tried with latest master as well.

lgoldstein:
The it must be fixed - however I have an idea (at least about the failing `CompressionTest`) - if one is use `zlib@openssh.com` then compression should take effect only after successful authentication. This requires some changes in the session setup code after KEX negotiation.

lgoldstein:
I think I've got it - according to the [draft](https://tools.ietf.org/html/draft-miller-secsh-compression-delayed-00)

>> Specifically, this means that a server selecting this method MUST
>>   compress the first packet after SSH_MSG_USERAUTH_SUCCESS and all
>>   subsequent packets.  The client MUST compress all packets it sends
>>   after the time that it receives SSH_MSG_USERAUTH_SUCCESS.

However, the code marks the session as authenticated **before** sending the `SSH_MSG_USERAUTH_SUCCESS` which means that `SSH_MSG_USERAUTH_SUCCESS` is sent compressed when the peer is not expecting it.

lgoldstein:
I believe I have a solution - if it is successful you will see it in this PR when I push it...

lgoldstein:
Sorry - I opened https://github.com/apache/mina-sshd/pull/91 - you can close this one if you like and we will discuss #91

The-Yoda:
Sure. Thanks.

PR #91, opened 2019-02-12
[SSHD-894] Fix Race condition when doing async auth.

lgoldstein:


sshd-core/src/main/java/org/apache/sshd/server/session/ServerSession.java
```diff
@@ -61,4 +63,27 @@
...
+     *      </LI></P>
+     *
+     *      <P><LI>
+     *      Starts the specified service.
+     *      </LI></P>
+     *
+     *      <P><LI>
+     *      Sends the {@code SSH_MSG_USERAUTH_SUCCESS} message.
+     *      </LI></P>
+     * </UL
```
thefourtheye: Nit: Closing tag is not proper.

sshd-core/src/main/java/org/apache/sshd/server/session/ServerUserAuthService.java
```diff
@@ -154,6 +154,26 @@ public synchronized void process(int cmd, Buffer buffer) throws Exception {
...
+             *      When SSH_MSG_USERAUTH_SUCCESS has been sent, any
+             *      further authentication requests received after that
+             *      SHOULD be silently ignored.
+             */
+            if (session.isAuthenticated()) {
+                String username = buffer.getString();
+                String service = buffer.getString();
+                String method = buffer.getString();
+
+                if (debugEnabled) {
```
thefourtheye: Why this guard is necessary? I believe the methods themselves will have the level check.

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
```diff
@@ -728,7 +728,8 @@ public IoWriteFuture writePacket(Buffer buffer) throws IOException {
         }
     }
 
-    protected IoWriteFuture doWritePacket(Buffer buffer) throws IOException {
+    // NOTE: must acquire encodeLock when calling this method
```
thefourtheye: Why not just asserting `Thread.holdsLock(encodeLock)` at the beginning of this function?

lgoldstein: 1.  I don't want to enforce this at the moment since this is an open-source project and thus I prefer leaving the "garbage-in-garbage-out" approach
2. The "happy flow" calls this method a lot (!!!) and I don't want to add un-necessary overhead.

sshd-core/src/main/java/org/apache/sshd/server/session/ServerUserAuthService.java
lgoldstein: 1. Yes, but:

```java
void debug(String format, Object ... args);
```
Which means that an intermediate array is automatically created, pushed to the stack and then eventually discarded - i.e., un-necessary object creating on the heap.

2. The same applies for
```java
log.debug("This " + session + " is " + getState() + " since " + getStartDate());
```
It creates a `StringBuilder`, appends to it various values (converting them to string(s)), converts it `toString()` and then does nothing with it - result: 2 (or more) objects created for nothing

3. Turns out that `isDebugEnabled()` is "cheaper" then calling `log.debug` and letting it figure out that there is nothing to do - even if we ignored the other 2 reasons.

sshd-core/src/main/java/org/apache/sshd/server/session/ServerSession.java
lgoldstein: Thanks - will fix

The-Yoda:
@lgoldstein  
Looks like one of our customer in production is using compression flags and they are not able to login.
Can you please let me know when 2.3.0 will be released?

lgoldstein:
We have no release date set - in any case, it will probably be some time June or later - depends on how many bug fixes and/or feature we accumulate. Meanwhile if it is something urgent you can cut your own version out of the *master* branch (assuming it contains the fix you need) and use it until we release 2.3.0. It is quite stable and IMO can serve you well until then.

PR #93, opened 2019-02-22
[SSHD-899] Discard attributes when file exist already.

The-Yoda:
As per https://tools.ietf.org/html/draft-ietf-secsh-filexfer-13#page-29,
The 'attrs' field is ignored if an existing file is opened.
As it is, attributes are being considered even if the file exists. This results in resetting permission even when file is opened for read.

This fix will discard the attributes when file exists.

@lgoldstein  @thefourtheye @sramki

### DIFF
--- a/sshd-sftp/src/main/java/org/apache/sshd/server/subsystem/sftp/FileHandle.java
+++ b/sshd-sftp/src/main/java/org/apache/sshd/server/subsystem/sftp/FileHandle.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.channels.FileLock;
 import java.nio.channels.SeekableByteChannel;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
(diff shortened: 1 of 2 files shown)


lgoldstein:
The remark you are referencing is for version 6 - version 3 (which is the prevailing one) is mute on this subject. However, I believe that version 6 recommended behavior is the more cautious one, so it seems to me that your request is reasonable. I will merge it.

Thanks for the (continued) contribution.

lgoldstein:
I have made the changes with a a slight variation - see https://github.com/apache/mina-sshd/pull/94

* Moved the logic that ignores the attributes to the `SftpFileSystemAccessor` - this way, if any user has any issues with this behavior they can register their own implementation and change it.
* I have qualified the test assertions to Unix (on Windows it throws `UnsupportedOperationException`).  I will test it on Linux to make sure it passes before I merge it in.
    * In this context, please try to run the code on Windows if you can before publishing a PR - I usually
   do it myself, but since we are all human I may forget and then we would have code that fails on Windows (or Linux...).

Let's close this PR and continue any discussions on the PR I opened

PR #95, opened 2019-02-27
SSHD-901 Provide 'wantReply' option for client keep-alive heartbeat global request

lgoldstein:


Arunapotti:
If I am using MINA SSHD and I want to use keepalive@... with wantReply  to close any dangling client sessions from server due to network issues, do you support this mechanism?

lgoldstein:
>> to close any dangling client sessions from server due to network issues

Please bear in mind that this mechanism is a **client side** one - i.e., it is used by the client in order to detect a non-responsive server sooner than the regular TCP/IP timeout would. AFAIK there is no server side mechanism to detect "dangling" client sessions. We do have an idle-timeout mechanism that closes a session if there is no traffic for the specified amount of time (current default=10 minutes). The purpose of the `keepalive@` mechanism (besides early detection) is also to generate some traffic so that the client does not time out if there is no traffic. By activating the `wantReply` option (default=disabled) we try to ensure that the server does not time out due to no traffic. Therefore, the `keepalive@` mechanism is not likely to detect "dangling clients" on the server side.

See https://github.com/apache/mina-sshd/blob/master/docs/client-setup.md#keeping-the-session-alive-while-no-traffic on client heartbeat details, and the code for `SessionHelper#getIdleTimeout`

There is a way to do this on the server side but it is rather **non-standard** (which we support) and it involves sending {{SSH_MSG_IGNORE}} messages from the server to the client thus maintaining some traffic on the session as well as detecting "dangling clients". See `setSessionHeartbeat` API with a {{HeartbeatType}} of `IGNORE`.

Arunapotti:
Hi Goldstein,

OpenSSH has ClinetAliveInterval support that the servers can use to detect
and clean up dead/dangling clients besides for the keep-alive support.

ClientAliveInterval <http://man.openbsd.org/sshd_config#ClientAliveInterval>
Sets a timeout interval in seconds after which if no data has been received
from the client, sshd(8) <http://man.openbsd.org/sshd.8>will send a message
through the encrypted channel to request a response from the client. The
default is 0, indicating that these messages will not be sent to the client.

I am looking for an equivalent option with your SSHD.

Thanks

lgoldstein:
This seems like the idle timeout I mentioned - although I  don't know how they
>> send a message through the encrypted channel to request a response from the client.


>> I am looking for an equivalent option with your SSHD.

See my response regarding usage of `SSH_MSG_IGNORE`

Arunapotti:
[image: image.png]
Don't you openSSH internally for SSHD? If you do, then I assume there
should be a way to use openSSH supported "ClinetAliveInterval" through your
SSHD.

lgoldstein:
>> Don't you openSSH internally for SSHD? 

MINA SSHD is a **pure Java** implementation - there is no _openSSH_ native code in it.

PR #97, opened 2019-03-20
Catch NumberFormatException when parsing versions

bangnab:
This is necessary to be able to cope with the reported versions of SSH-2.0-VShell_4_2_1_1051 VShell:
"3,4,5,6,draft-ietf-secsh-filexfer-11@vandyke.com,partial-v6@vandyke.com"


lgoldstein:
Thanks for the bug fix - I have fixed it in the code but slightly differently - see https://github.com/apache/mina-sshd/pull/98 (close this one and let's continue the conversation on the PR I have opened). Please compile https://github.com/lgoldstein/mina-sshd/tree/SSHD-909 and see if indeed it fixes the issue since I don't have access to an SFTP server that declares versions such as the ones you encountered (or we would have caught this issue sooner...)

PR #99, opened 2019-04-15
Shutdown output when receiving SSH_MSG_CHANNEL_EOF message.

yuminsang:
The changes are based on commit 2727fdda8009dad98ddf1ba0cb27c67f16cb9ea7. Please help check if it is feasible. Thank you.


lgoldstein:
The patch does not provide implementation of `handleEof` for `MinaSession` and `NettyIoSession` classes so the code does not compile in full. I could implement it as `close` but it does not seem equivalent to `Nio2Session` implementation that only shuts down the output. Please advise/fix...

lgoldstein:
I have found a way around it using reflection API, but I am not too happy about using it - see https://github.com/apache/mina-sshd/pull/100.

Let's close this PR and move the discussion to the one I opened.

yuminsang:
Ok. Thanks.

PR #100, opened 2019-04-16
SSHD-902 Shutdown output when receiving SSH_MSG_CHANNEL_EOF

lgoldstein:


cegganesh84:
Hi,

We are seeing issues in our port forwarding where the EOF messages or not handled properly. Will this pull request be merged sooner, so that we can try it out in the development environment.

Thanks,
Ganesh

lgoldstein:
I  would rather you clone https://github.com/lgoldstein/mina-sshd/tree/SSHD-902 and test it locally (since it seems to reproduce on your system) before I merge it in. In any case, even if I do merge it in, it will not be available until the next release - probably sometime in June or July...

yuminsang:
Cloned https://github.com/lgoldstein/mina-sshd/tree/SSHD-902 and tested it locally, and it works for my case. Thanks.

PR #104, opened 2019-06-24
[SSHD-929] Create separate listener for remove directory.

The-Yoda:
Right now listener for deletion of file and directory is common. This makes it difficult to find whether directory / file has been removed.

The path passed to `removed` function in SftpEventListener will not tell if the deleted path was a directory or file. `Files.isDirectory(path)` will return false, as the directory doesn't exist.

This patch will have a fix to create one more listener entry for directory removal.

### DIFF
--- a/sshd-sftp/src/main/java/org/apache/sshd/server/subsystem/sftp/AbstractSftpSubsystemHelper.java
+++ b/sshd-sftp/src/main/java/org/apache/sshd/server/subsystem/sftp/AbstractSftpSubsystemHelper.java
@@ -1514,9 +1514,9 @@ public abstract class AbstractSftpSubsystemHelper
                   getServerSession(), id, path, p);
         }
         if (Files.isDirectory(p, options)) {
-            doRemove(id, p);
+            doRemove(id, p, true);
         } else {
-            throw signalRemovalPreConditionFailure(id, path, p, new NotDirectoryException(p.toString()));
+            throw signalRemovalPreConditionFailure(id, path, p, new NotDirectoryException(p.toString()), true);
         }
     }
 
@@ -1527,17 +1527,18 @@ public abstract class AbstractSftpSubsystemHelper
      * @param p {@link Path} to delete
      * @throws IOException If failed to delete
      */
-    protected void doRemove(int id, Path p) throws IOException {
+    protected void doRemove(int id, Path p, boolean isDirectory) throws IOException {
         SftpEventListener listener = getSftpEventListenerProxy();
         ServerSession session = getServerSession();
-        listener.removing(session, p);
+
+        signalRemoving(p, listener, session, isDirectory);
         try {
             Files.delete(p);
         } catch (IOException | RuntimeException e) {
-            listener.removed(session, p, e);
+            signalRemoved(p, listener, session, isDirectory, e);
             throw e;
         }
-        listener.removed(session, p, null);
+        signalRemoved(p, listener, session, isDirectory, null);
     }
 
     protected void doMakeDirectory(Buffer buffer, int id) throws IOException {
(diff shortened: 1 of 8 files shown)


sshd-sftp/src/main/java/org/apache/sshd/server/subsystem/sftp/AbstractSftpSubsystemHelper.java
```diff
@@ -1530,14 +1530,34 @@ protected void doRemoveDirectory(int id, String path, LinkOption... options) thr
     protected void doRemove(int id, Path p) throws IOException {
         SftpEventListener listener = getSftpEventListenerProxy();
         ServerSession session = getServerSession();
-        listener.removing(session, p);
+        boolean isDirectory = Files.isDirectory(p);
```
lgoldstein: According to the Javadoc:

>> false if the file does not exist, is not a directory, or it cannot be determined if the file is a directory or not.

This means that if file does not exist to begin with we might declare it a"regular" file and signal the incorrect callback.

sshd-sftp/src/main/java/org/apache/sshd/server/subsystem/sftp/SftpEventListener.java
```diff
@@ -393,6 +393,30 @@ default void removed(ServerSession session, Path path, Throwable thrown) throws
...
 
+    /**
+     * Called <U>prior</U> to removing a directory
+     *
+     * @param session The {@link ServerSession} through which the request was handled
+     * @param path    The {@link Path} about to be removed
+     * @throws IOException If failed to handle the call
+     * @see #removed(ServerSession, Path, Throwable)
+     */
+    default void removingDirectory(ServerSession session, Path path) throws IOException {
```
lgoldstein: In order to really make the distinction, the original `removing/remove` should also be renamed to `removingFile/removedFile`

The-Yoda: Ack

sshd-sftp/src/main/java/org/apache/sshd/server/subsystem/sftp/AbstractSftpSubsystemHelper.java
```diff
@@ -1527,17 +1527,36 @@ protected void doRemoveDirectory(int id, String path, LinkOption... options) thr
...
         } catch (IOException | RuntimeException e) {
-            listener.removed(session, p, e);
+            signalRemoved(p, listener, session, isDirectory, e);
             throw e;
         }
-        listener.removed(session, p, null);
+        signalRemoved(p, listener, session, isDirectory, null);
+    }
+
+    private void signalRemoving(Path p, SftpEventListener listener, ServerSession session,
```
lgoldstein: protected  please - we always try to allow users to override any behavior they like

sshd-sftp/src/main/java/org/apache/sshd/server/subsystem/sftp/AbstractSftpSubsystemHelper.java
```diff
@@ -1527,17 +1527,36 @@ protected void doRemoveDirectory(int id, String path, LinkOption... options) thr
...
+                                boolean isDirectory) throws IOException {
+        if (isDirectory) {
+            listener.removingDirectory(session, p);
+        } else {
+            listener.removingFile(session, p);
+        }
+    }
+
+    private void signalRemoved(Path p, SftpEventListener listener, ServerSession session,
```
lgoldstein: protected  please - we always try to allow users to override any behavior they like

sshd-sftp/src/main/java/org/apache/sshd/server/subsystem/sftp/AbstractSftpSubsystemHelper.java
```diff
@@ -1631,8 +1650,8 @@ protected void doRemove(int id, String path, LinkOption... options) throws IOExc
             log.debug("signalRemovalPreConditionFailure(id={})[{}] signal {} for {}: {}",
                 id, pathValue, thrown.getClass().getSimpleName(), path, thrown.getMessage());
         }
-        listener.removing(session, path);
-        listener.removed(session, path, thrown);
+        listener.removingFile(session, path);
+        listener.removedFile(session, path, thrown);
```
lgoldstein: Not sure this is correct - {{signalRemovalPreConditionFailure}} may be called for both file and directory so by what right do we decide here that it is a file and not a directory ?

The-Yoda: Thats right. It is being called for both directory and file. Fixed the same.

lgoldstein:
Was OK in general (several minor style fixes) - however, you did not run a full `mvn clean install` or you would have discovered one more location to fix (a unit test was failing). No big issue - just for future reference :-)) . I need to run it also on a *Windows* machine (just to be on the safe side) and then I will merge it in with full credit to your contribution - I believe some time over the weekend...

The-Yoda:
Thanks @lgoldstein 
May I know the style fixes you mentioned? Will help me to correct myself in upcoming PRs.

Not related to this, I'm not able to run full `mvn clean install`. Always endup with this error ```[ERROR]   OpenSSHKeyPairResourceParserPasswordTest.testLoadEncryptedFileWithPasswordRetry:139->OpenSSHKeyPairResourceParserTestSupport.testLoadKeyPairs:71->Assert.assertTrue:41->Assert.fail:88 Too many retries: 3``` and ```[ERROR]   OpenSSHKeyPairResourceParserDecodingTest.testLoadEncryptedKeyPairs:66->OpenSSHKeyPairResourceParserTestSupport.testLoadKeyPairs:71 » NoSuchAlgorithm```

lgoldstein:
>> May I know the style fixes you mentioned? Will help me to correct myself in upcoming PRs.

Compare https://github.com/lgoldstein/mina-sshd/tree/SSHD-929 with your original pull request - mainly some spacing and alignments - again, nothing major...

>> Not related to this, I'm not able to run full mvn clean install. Always endup with this error [ERROR] 

Strange - I am running `mvn clean install` often on **both** *Windows* and *Linux* and have never encountered this issue. I am curious about the `NoSuchAlgorithm` error - are you using a standard JDK (*Oracle* or *OpenJDK*) ? If so which ?

Perhaps you could debug it and specify which algorithm is reported as missing - I might be able to add some condition on the failing test. In any case, you can place a **temporary** `@Ignore` on it (don't check it in) and keep running.

The-Yoda:
Fixed the failing test.
I used @Ignore on the below tests to make `mvn clean install` to succeed.
`OpenSSHKeyPairResourceParserDecodingTest`, `OpenSSHKeyPairResourceParserPasswordTest`, `PortForwardingTest`

Did a little bit debugging on the cipher part. `aes256-cbc` is not supported in my jvm. I'm using oracle jdk (jdk1.8.0_144).
I'm not sure why `OpenSSHKeyPairResourceParserPasswordTest` tests are failing.

lgoldstein:
There used to be (and perhaps there still is) some issue about "strong" ciphers (such as aes256-cbc) where they have to be "enabled" by modifying some default JDK security settings (look it up on the internet). Anyway, I'll have a look at the failing test since our code  is supposed to detect automatically whether a certain cipher (and other security settings) are supported or not

PR #105, opened 2019-07-08
SSHD-930 Send the client version string after receiving the version string of the server

lgoldstein:
Leave default to be same as current - i.e., immediate


sshd-core/src/main/java/org/apache/sshd/client/session/ClientSessionImpl.java
```diff
@@ -90,16 +88,9 @@ public ClientSessionImpl(ClientFactoryManager client, IoSession ioSession) throw
...
-        KexExtensionHandler extHandler = getKexExtensionHandler();
-        if ((extHandler == null) || (!extHandler.isKexExtensionsAvailable(this, AvailabilityPhase.PREKEX))) {
-            kexState.set(KexState.INIT);
-            sendKexInit();
-        } else {
-            if (log.isDebugEnabled()) {
-                log.debug("<init>({}) delay KEX-INIT until server-side one received", this);
-            }
+        if (sendImmediateIdentification) {
+            initializeKexPhase();
```
tomaswolf: I don't think this is OK. This will make it impossible to implement any client-side proxy protocol to connect through SOCKS or HTTP CONNECT proxies if `!sendImmediateIdentification`. If sending the client identification is delayed, the proxy protocol must still be started before expecting an answer from the server.

lgoldstein: Please see the updated code that initializes the proxy connector regardless of whether immediate send is enabled or not.

sshd-core/src/main/java/org/apache/sshd/client/session/ClientProxyConnector.java
```diff
@@ -31,12 +31,13 @@
 @FunctionalInterface
 public interface ClientProxyConnector {
     /**
-     * Invoked just before the client identification is sent so that the
-     * proxy can send the meta-data to its peer. Upon successful return
-     * the SSH identification line is sent and the protocol proceeds as usual.
+     * Invoked just before the client identification is about to be sent so that
```
tomaswolf: This comment is not true anymore. It's called just before the SSH protocol starts.

sshd-core/src/main/java/org/apache/sshd/client/session/AbstractClientSession.java
```diff
@@ -231,37 +240,55 @@ public KeyPair removePublicKeyIdentity(KeyPair kp) {
...
+            kexState.set(KexState.INIT);
+            sendKexInit();
+        } else {
+            if (log.isDebugEnabled()) {
+                log.debug("initializeKexPhase({}) delay KEX-INIT until server-side one received", this);
+            }
+        }
+    }
 
+    protected ClientProxyConnector initializeProxyConnector() throws Exception {
```
tomaswolf: Why return the ClientProxyConnector here? The only place it's called the return value is ignored, and subclasses could get it via `getClientProxyConnector()` anyway.

sshd-core/src/main/java/org/apache/sshd/client/session/ClientSessionImpl.java
tomaswolf: That would work.

sshd-core/src/main/java/org/apache/sshd/client/session/ClientSessionImpl.java
```diff
@@ -90,16 +88,17 @@ public ClientSessionImpl(ClientFactoryManager client, IoSession ioSession) throw
...
-            }
+        /*
+         * Must be called regardless of whether the client identification
+         * is sent or not immediately in order to allow opening any underlying
+         * proxy protocol - e.g., SOCKS or HTTP CONNECT - otherwise the server's
+         * identification will never arrive
+         */
+        initializeProxyConnector();
+
+        if (sendImmediateIdentification) {
```
tomaswolf: A tiny bit ugly that `if (sendImmediateIdentification)` is here in ClientSessionImpl, but the corresponding `if (!sendImmediateIdentification)` is in the super class AbstractClientSession.

lgoldstein: Agreed, but I did not want to make too many changes...

sshd-core/src/main/java/org/apache/sshd/client/session/ClientSessionImpl.java
lgoldstein: See updated code in view of your comments (thanks for the feedback)

tomaswolf:
Looks OK to me. Detailed review of diff between this force-push and f7bde2c not possible for me on Github; I hadn't fetched the previous state.

PR #106, opened 2019-07-29
Fix typo in dependency name

FliegenKLATSCH:


lgoldstein:
Thanks, applied the patch

PR #107, opened 2019-10-07
Record which AuthorizedKeyEntry is used to authenticate

wh0:
This adds a new class, `AuthorizedKeyEntriesPublickeyAuthenticator`, based on the existing `KeySetPublickeyAuthenticator` and `resolvePublicKeyEntries`. It provides enhanced functionality over these existing routines by remembering the `AuthorizedKeyEntry`s, where the previous functionality discards everything but the public key. This class records the matching entry into the session, as an attribute, which will enable custom session functionality to make decisions based the `AuthorizedKeyEntry`'s options.

This changes `PublickeyAuthenticator.fromAuthorizedEntries` to create an `AuthorizedKeyEntriesPublickeyAuthenticator` instead of a `KeySetPublickeyAuthenticator`. This will affect `AuthorizedKeysAuthenticator`, which uses `PublickeyAuthenticator.fromAuthorizedEntries`.


gnodet:
This has been integrated.

PR #108, opened 2019-10-08
ChannelSession: Split parsing from handlers

wh0:
This splits out part of `handleXxx` into `handleXxxParsed`, which takes a bunch of named arguments instead of a `Buffer buffer`. This way, subclasses can handle these requests differently while still using the same parsing logic as ChannelSession.


gnodet:
Included with https://github.com/apache/mina-sshd/commit/75071c60d683124c93b6bdeb5d90ffe2acec0e8b and https://github.com/apache/mina-sshd/commit/adff42b650ac9bc0d2da4c3a4dec5c8099197042

PR #111, opened 2020-02-20
[SSHD-967] fixed byte buffer issue

NitinRanjan:
In the last cycle of write if read size was less than buffer size remaining buffer indexes was still containing old value and was flushed in the file.
Fix this by giving the amount of byte to be wrapped to  flushed to file

### DIFF
--- a/sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/SftpRemotePathChannel.java
+++ b/sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/SftpRemotePathChannel.java
@@ -262,7 +262,7 @@ public class SftpRemotePathChannel extends FileChannel {
                 while (totalRead < count) {
                     int read = sftp.read(handle, curPos, buffer, 0, buffer.length);
                     if (read > 0) {
-                        ByteBuffer wrap = ByteBuffer.wrap(buffer);
+                        ByteBuffer wrap = ByteBuffer.wrap(buffer, 0,  (int) Math.min(count - totalRead, buffer.length));
                         while (wrap.remaining() > 0) {
                             target.write(wrap);
                         }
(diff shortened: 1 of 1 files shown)


sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/SftpRemotePathChannel.java
```diff
@@ -262,7 +262,7 @@ public long transferTo(long position, long count, WritableByteChannel target) th
                 while (totalRead < count) {
                     int read = sftp.read(handle, curPos, buffer, 0, buffer.length);
                     if (read > 0) {
-                        ByteBuffer wrap = ByteBuffer.wrap(buffer);
+                        ByteBuffer wrap = ByteBuffer.wrap(buffer, 0,  (int) Math.min(count - totalRead, buffer.length));
```
lgoldstein: Shouldn't this be ?

```java
ByteBuffer wrap = ByteBuffer.wrap(buffer, 0,  read);
```

lgoldstein:
I would feel better if you could also add a unit test that checks specifically this issue (shouldn't be too hard). First write the unit test and make sure it fails on the **unpatched** code, then apply the patch and see that the test succeeds.

lgoldstein:
Sorry to have misled you - after reviewing the code I thing the fix is slightly more complex than that. I am working on something myself and will post it shortly for you to review and test...

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/SftpRemotePathChannel.java
NitinRanjan: Both are same, but to make it understand easily it is better to change it to read.
 I have tested with read and it working fine. Committed the change to forked branch

lgoldstein:
Please clone https://github.com/lgoldstein/mina-sshd/tree/SSHD-967 compile and try

PR #112, opened 2020-02-26
SSHD-968

gnodet:


lgoldstein:
The suggested fix seems to involve too many methods changes - not mention the fact that the extra argument is very specific to this issue. What if future fixes require more argument ? However, your fix has given me an idea I would like to explore that may provide (IMO) a better solution. Be patient for 1-2 days and I might have something.

PR #116, opened 2020-03-01
Add authorized_keys/verification support for OpenSSH "security key" key types

JackOfMostTrades:
This PR adds supports for the new OpenSSH "security key" key types added in the latest OpenSSH release. Details of the key types are available [here](https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.u2f).

This PR just intends to add support for verify peers using those key types. This does not support any private key operations, since that would probably involve interfacing the the hardware security key devices directly which is beyond the scope of what I'd really like to do. ;)

I've added unit tests that verify that authorized_keys files can be parsed with the new key types and that the new signature classes are behaving correctly (including correctly verifying the user presence flag). I've also manually tested that with these changes, a server started using the `DefaultAuthorizedKeysAuthenticator` will correctly accept connections from clients using these new key types.


sshd-common/src/main/java/org/apache/sshd/common/u2f/SkEcdsaPublicKey.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public SkEcdsaPublicKey(String appName, boolean noTouchRequired, ECPublicKey delegatePublicKey) {
+        this.appName = appName;
+        this.noTouchRequired = noTouchRequired;
+        this.delegatePublicKey = delegatePublicKey;
+    }
+
+    @Override
+    public String getAlgorithm() {
+        return "SK-ECDSA";
```
lgoldstein: Please make this a public literal constant of this class

sshd-common/src/main/java/org/apache/sshd/common/u2f/SkED25519PublicKey.java
```diff
@@ -0,0 +1,66 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.sshd.common.u2f;
+
+import net.i2p.crypto.eddsa.EdDSAPublicKey;
+
+public class SkED25519PublicKey implements SecurityKeyPublicKey<EdDSAPublicKey> {
```
lgoldstein: Please add a `toString` implementation - show appName, noTouchRequired and delegatePublicKey values... - remember to use the getters:

```java
@Override
public String toString() {
     return getClass().getSimpleName()
        + "[appName=" + getAppName()
        + ", noTouch=" + isNoTouch()
        + ", delegateKey=" + getDelegateKey()
        + "]";
}
```

sshd-common/src/main/java/org/apache/sshd/common/u2f/SkEcdsaPublicKey.java
```diff
@@ -0,0 +1,66 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.sshd.common.u2f;
+
+import java.security.interfaces.ECPublicKey;
+
+public class SkEcdsaPublicKey implements SecurityKeyPublicKey<ECPublicKey> {
```
lgoldstein: Please add a `toString` implementation (see previous remark on this issue as well)

sshd-common/src/main/java/org/apache/sshd/common/util/buffer/keys/SkECBufferPublicKeyParser.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public class SkECBufferPublicKeyParser extends AbstractBufferPublicKeyParser<SkEcdsaPublicKey> {
+    public static final SkECBufferPublicKeyParser INSTANCE = new SkECBufferPublicKeyParser();
+
+    public SkECBufferPublicKeyParser() {
+        super(SkEcdsaPublicKey.class, SkECDSAPublicKeyEntryDecoder.KEY_TYPE);
+    }
+
+    @Override
+    public Collection<String> getSupportedKeyTypes() {
```
lgoldstein: Why is this override required ? All it does is call the superclass...

sshd-common/src/main/java/org/apache/sshd/common/util/buffer/keys/SkED25519BufferPublicKeyParser.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+public class SkED25519BufferPublicKeyParser extends AbstractBufferPublicKeyParser<SkED25519PublicKey> {
+    public static final SkED25519BufferPublicKeyParser INSTANCE = new SkED25519BufferPublicKeyParser();
+
+    public SkED25519BufferPublicKeyParser() {
+        super(SkED25519PublicKey.class, SkED25519PublicKeyEntryDecoder.KEY_TYPE);
+    }
+
+    @Override
+    public Collection<String> getSupportedKeyTypes() {
```
lgoldstein: Why is this override required ? All it does is call the superclass...

sshd-core/src/main/java/org/apache/sshd/client/ClientBuilder.java
```diff
@@ -79,7 +79,9 @@
                 BuiltinSignatures.nistp256,
                 BuiltinSignatures.nistp384,
                 BuiltinSignatures.nistp521,
+                BuiltinSignatures.sk_ecdsa_sha2_nistp256,
```
lgoldstein: I am not comfortable to adding it as default for the client (for now) - users who want this should make an "effort" (at least for now).

sshd-core/src/main/java/org/apache/sshd/client/ClientBuilder.java
```diff
@@ -79,7 +79,9 @@
...
                 BuiltinSignatures.ed25519,
+                BuiltinSignatures.sk_ssh_ed25519,
```
lgoldstein: I am not comfortable to adding it as default for the client (for now) - users who want this should make an "effort" (at least for now).

sshd-common/src/main/java/org/apache/sshd/common/config/keys/impl/SkECDSAPublicKeyEntryDecoder.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    public SkEcdsaPublicKey decodePublicKey(
+            SessionContext session, String keyType, InputStream keyData, Map<String, String> headers)
+                throws IOException, GeneralSecurityException {
+        if (!KEY_TYPE.equals(keyType)) {
+            throw new InvalidKeySpecException("Invalid keyType: " + keyType);
+        }
+
+        boolean noTouchRequired = false;
+        if (headers.containsKey(NO_TOUCH_REQUIRED_HEADER)) {
```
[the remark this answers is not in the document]
JackOfMostTrades: Put this and the other boolean parsing into a helper method in the parent class.

sshd-common/src/main/java/org/apache/sshd/common/util/buffer/keys/SkECBufferPublicKeyParser.java
JackOfMostTrades: Oops, leftover code from some debugging. Removed.

JackOfMostTrades:
Added a new commit that should address the above comments. Thanks for the quick response and the thorough review!

lgoldstein:
>> Added a new commit that should address the above comments. 

Good work - I will merge it in soon (after testing it on my host) - probably tomorrow.

>> Thanks for the quick response and the thorough review!

My pleasure

JackOfMostTrades:
Create this JIRA issue for tracking: https://issues.apache.org/jira/browse/SSHD-972

lgoldstein:
Merged with many thanks + acknowledgement - see https://github.com/apache/mina-sshd/commit/ff22a9b00c525a54ebf2490266afb88f3e528185. Please close the PR.

JackOfMostTrades:
Excellent. Thanks again!

PR #118, opened 2020-04-07
[SSHD-975] Use AbstractFactoryManager's classloader for proxies

rovarga:
AbstractFactoryManager cannot make assumptions about the nature of the classloader returned by getClass().getClassLoader(). Use AbstractFactoryManager.class.getClassLoader() for loading proxies.

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/common/channel/AbstractChannel.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/channel/AbstractChannel.java
@@ -130,8 +130,7 @@ public abstract class AbstractChannel
         gracefulFuture = new DefaultCloseFuture(discriminator, futureLock);
         localWindow = new Window(this, null, client, true);
         remoteWindow = new Window(this, null, client, false);
-        channelListenerProxy = EventListenerUtils.proxyWrapper(
-            ChannelListener.class, getClass().getClassLoader(), channelListeners);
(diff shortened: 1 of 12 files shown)


rovarga:
> If this change fixes the issue then I believe we should apply it to **all** the places where proxy wrappers are being created - which means also in `ScpCommandFactory`, `AbstractSftpSubsystemHelper`, `AbstractSftpEventListenerManager`, `AbstractConnectionService`, etc...

Agreed, think we should just use the method without ClassLoader argument and instantiate the proxy in the classloader where the interface lives -- it's the simplest and guaranteed to work. What do you think?

lgoldstein:
>> Agreed, think we should just use the method without ClassLoader argument and instantiate the proxy in the classloader where the interface lives -- it's the simplest and guaranteed to work. What do you think?

My thinking exactly - let me know if it fixes your problem. If so I will be more than happy to merge it to our _master_ branch. Please make sure that `mvn clean install` succeeds on the entire project.

rovarga:
Current patch does precisely that. it passed local build for me and the windows build here seems to have failed on an environment issue.

lgoldstein:
Good - I  will merge it in and verify - thanks for the contribution

lgoldstein:
Merged with many thanks and acknowledgement - see [60bdb04642](https://github.com/apache/mina-sshd/commit/60bdb04642cbcb82501841b9f25601c5532d537b). Please close this PR

PR #119, opened 2020-04-08
Add support for openssh host key certificates

FliegenKLATSCH:
This is part of SSHD-660. Would be happy to get some feedback.


sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificate.java
```diff
@@ -0,0 +1,262 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.sshd.common.config.keys;
+
+import java.security.PrivateKey;
+import java.security.PublicKey;
+import java.util.List;
+
+public final class OpenSshCertificate implements PublicKey, PrivateKey {
```
lgoldstein, line 25: We try to make our code as open as possible - so

* as few `final` classes as possible
* fields should be `protected` - unless there is a getter/setter
* constructors should be `public`

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificate.java
```diff
@@ -0,0 +1,262 @@
...
+            this.message = message;
+            return this;
+        }
+
+        public OpenSshPublicKeyBuilder withSignature(byte[] signature) {
+            this.signature = signature;
+            return this;
+        }
+
+        public OpenSshCertificate build() {
```
lgoldstein, line 242: Since the only way to build an instantiated class is via the builder and there are only getters the fields should be `final`

sshd-common/src/main/java/org/apache/sshd/common/keyprovider/FileHostKeyCertificateProvider.java
```diff
@@ -0,0 +1,110 @@
...
+    public FileHostKeyCertificateProvider() {
+        super();
+    }
+
+    public FileHostKeyCertificateProvider(Path path) {
+        this(Collections.singletonList(Objects.requireNonNull(path, "No path provided")));
+    }
+
+    public FileHostKeyCertificateProvider(Path... files) {
+        this(Arrays.asList(files));
```
lgoldstein, line 52: Should use `ValidateUtils.notEmpty(files)`

sshd-common/src/main/java/org/apache/sshd/common/keyprovider/FileHostKeyCertificateProvider.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Override
+    public OpenSshCertificate loadCertificate(SessionContext session, String keyType) throws IOException, GeneralSecurityException {
+        return StreamSupport.stream(loadCertificates(session).spliterator(), false)
+                .filter(pubKey -> pubKey.getKeyType().equals(keyType))
+                .findFirst().orElse(null);
+    }
+
+    //    public void setPaths(Collection<? extends Path> paths) {
```
lgoldstein, line 88: So much code not used should not be commented - please remove it

sshd-common/src/main/java/org/apache/sshd/common/keyprovider/KeyPairProvider.java
```diff
@@ -83,8 +93,8 @@ public KeyPair loadKey(SessionContext session, String type) {
             }
 
             @Override
-            public Iterable<String> getKeyTypes(SessionContext session) {
-                return Collections.emptyList();
+            public Set<String> getKeyTypes(SessionContext session) {
```
lgoldstein, line 96: Why change this ? I see no reason... - please revert the change or provide compelling reason as to why this is 100% necessary. Please note that some code may rely on `Iterable`...

sshd-common/src/main/java/org/apache/sshd/common/config/keys/impl/OpenSSHCertificateDecoder.java
```diff
@@ -0,0 +1,108 @@
...
+        Objects.requireNonNull(key, "No public key provided");
+
+        String keyType = key.getKeyType();
+        KeyEntryResolver.encodeString(s, keyType);
+        s.write(key.getRawData());
+        return keyType;
+    }
+
+    @Override
+    public OpenSshCertificate clonePublicKey(OpenSshCertificate key) throws GeneralSecurityException {
```
lgoldstein, line 90: I am not comfortable with so many unimplemented methods...

sshd-core/src/main/java/org/apache/sshd/client/kex/DHGClient.java
```diff
@@ -124,10 +129,67 @@ public boolean next(int cmd, Buffer buffer) throws Exception {
 
         buffer = new ByteArrayBuffer(k_s);
         serverKey = buffer.getRawPublicKey();
-        String keyAlg = KeyUtils.getKeyType(serverKey);
+        PublicKey serverPublicHostKey = serverKey;
+
+        OpenSshCertificate openSshKey = null;
```
lgoldstein, line 134: Why is it defined outside the `if` block ? I don't see it being used outside it...

sshd-core/src/main/java/org/apache/sshd/client/kex/DHGClient.java
```diff
@@ -124,10 +129,67 @@ public boolean next(int cmd, Buffer buffer) throws Exception {
...
+            // verify signature
+            PublicKey signatureKey = openSshKey.getCaPubKey();
+            String keyAlg = KeyUtils.getKeyType(signatureKey);
+            Signature verif = ValidateUtils.checkNotNull(
+                    NamedFactory.create(session.getSignatureFactories(), keyAlg),
+                    "No verifier located for algorithm=%s", keyAlg);
+            verif.initVerifier(session, signatureKey);
+            verif.update(session, openSshKey.getMessage());
+            if (!verif.verify(session, openSshKey.getSignature())) {
+                throw new SshException(SshConstants.SSH2_DISCONNECT_KEY_EXCHANGE_FAILED,
```
lgoldstein, line 148: Let's try and find a way to move this large block of code to some method,,,

sshd-core/src/main/java/org/apache/sshd/client/kex/DHGClient.java
```diff
@@ -124,10 +129,67 @@ public boolean next(int cmd, Buffer buffer) throws Exception {
...
+                    throw new SshException(SshConstants.SSH2_DISCONNECT_KEY_EXCHANGE_FAILED,
+                            "KeyExchange signature verification failed, invalid principal: "
+                                    + openSshKey.getPrincipals());
+                }
+            } else {
+                throw new SshException(SshConstants.SSH2_DISCONNECT_KEY_EXCHANGE_FAILED,
+                        "KeyExchange signature verification failed, could not determine connect host.");
+            }
+
+            if (!openSshKey.getCriticalOptions().isEmpty()) {
```
lgoldstein, line 181: Please use `GenericUtils.isEmpty(...getCriticalOptions())` since it is null-safe

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificate.java
FliegenKLATSCH: If I remove the final I get:
`Error:(25) java: (design) FinalClass: Class OpenSshCertificate should be declared as final.`

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificate.java
FliegenKLATSCH: How can I do this without having all parameters in the constructor arguments? Again the reason for the builder is:
`Error:(47,13) java: (sizes) ParameterNumber: More than 8 parameters (found 15).`

sshd-common/src/main/java/org/apache/sshd/common/keyprovider/KeyPairProvider.java
FliegenKLATSCH: Hehe, I was pretty sure you would say this :) I don't have a reason except avoiding some back and forth converting, will change it back to `Iterable` again.

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificate.java
lgoldstein: That's because the only way to build it is via the builder. If you replace the builder with getters and setters you will no longer have this problem.

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificate.java
lgoldstein: You can pass the builder instance to the constructor

sshd-common/src/main/java/org/apache/sshd/common/keyprovider/FileHostKeyCertificateProvider.java
```diff
@@ -0,0 +1,88 @@
...
+    public Collection<? extends Path> getPaths() {
+        return files;
+    }
+
+    @Override
+    public Iterable<OpenSshCertificate> loadCertificates(SessionContext session) throws IOException, GeneralSecurityException {
+
+        List<OpenSshCertificate> certificates = new ArrayList<>();
+        for (Path file : files) {
+            List<String> lines = IoUtils.readAllLines(new FileInputStream(file.toFile()));
```
lgoldstein, line 69: Please use try-with-resource on the input stream - this code opens a stream but does not close it. Better yet - use `Files.readAllLines(file, StandardCharset.UTF8)`

sshd-common/src/main/java/org/apache/sshd/common/keyprovider/FileHostKeyCertificateProvider.java
```diff
@@ -0,0 +1,88 @@
...
+            for (String line : lines) {
+                PublicKeyEntry publicKeyEntry = PublicKeyEntry.parsePublicKeyEntry(line);
```
lgoldstein, line 71: What if the line is empty ?
What if the line is a comment ?

sshd-common/src/main/java/org/apache/sshd/common/keyprovider/FileHostKeyCertificateProvider.java
```diff
@@ -0,0 +1,88 @@
...
+
+                PublicKey publicKey = publicKeyEntry.resolvePublicKey(session, null, null);
+
+                certificates.add((OpenSshCertificate) publicKey);
```
lgoldstein, line 75: Let's check that it is indeed an `OpenSshCertificate` rather than risk `ClassCastException`:
```java
PublicKey publicKey = .....
if (publicKey == null) {
    continue;
}
if (!(publicKey instanceof OpenSshCertificate)) {
    throw new InvalidKeyException("....not and OpenSshCertificate...");
}
```

sshd-common/src/main/java/org/apache/sshd/common/keyprovider/KeyPairProvider.java
```diff
@@ -175,7 +185,7 @@ public KeyPair loadKey(SessionContext session, String type) {
             @Override
             public Iterable<String> getKeyTypes(SessionContext session) {
                 // use a LinkedHashSet so as to preserve the order but avoid duplicates
-                Collection<String> types = new LinkedHashSet<>();
+                Set<String> types = new LinkedHashSet<>();
```
lgoldstein, line 188: Please revert this type to a `Collection` as the original code...

sshd-common/src/main/java/org/apache/sshd/common/util/buffer/keys/OpenSSHCertPublicKeyParser.java
```diff
@@ -0,0 +1,104 @@
...
+        List<String> criticalOptions = buffer.getNameList();
+        List<String> extensions = buffer.getNameList();
+
+        String reserved = buffer.getString();
+
+        PublicKey signatureKey;
+        try {
+            signatureKey = buffer.getPublicKey();
+        } catch (SshException ex) {
+            throw new GeneralSecurityException("Could not parse public CA key.", ex);
```
lgoldstein, line 75: Let's add the ID of the key that we could not parse to the message for debugging purposes

sshd-common/src/main/java/org/apache/sshd/common/util/buffer/keys/OpenSSHCertPublicKeyParser.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        byte[] message = buffer.getBytesConsumed();
+        byte[] signature = buffer.getBytes();
+
+        if (buffer.rpos() != buffer.wpos()) {
+            throw new GeneralSecurityException("KeyExchange signature verification failed, got more data than expected: "
```
lgoldstein, line 82: Let's add the ID of the key to the exception message - for debugging purposes

sshd-core/src/main/java/org/apache/sshd/server/session/AbstractServerSession.java
```diff
@@ -367,9 +382,26 @@ protected String resolveAvailableSignaturesProposal(FactoryManager proposedManag
...
+                    .collect(Collectors.toSet());
+
+                HostKeyCertificateProvider hostKeyCertificateProvider = getHostKeyCertificateProvider();
+                if (hostKeyCertificateProvider != null) {
+                    Iterable<OpenSshCertificate> certificates = hostKeyCertificateProvider.loadCertificates(this);
+                    for (OpenSshCertificate certificate : certificates) {
+                        // Add the certificate alg only if the corresponding keyPair type is available
+                        if (provided.contains(certificate.getRawKeyType())) {
+                            provided.add(certificate.getKeyType());
+                        } else {
```
lgoldstein, line 398: This code interferes with a lazy-load feature (see SSHD-860) that we implemented since it consumes all available key types thus defeating the purpose of the lazy-load. We need to figure out a way to preserve that capability

sshd-common/src/main/java/org/apache/sshd/common/config/keys/impl/OpenSSHCertificateDecoder.java
```diff
@@ -0,0 +1,115 @@
...
+        try {
+            ByteArrayOutputStream outStream = new ByteArrayOutputStream();
+            String keyType = encodePublicKey(outStream, key);
+            InputStream inStream = new ByteArrayInputStream(outStream.toByteArray());
+            return decodePublicKey(null, keyType, inStream, null);
+        } catch (IOException e) {
+            throw new GeneralSecurityException("Unable to clone key.", e);
+        }
+    }
```
lgoldstein, line 99: Even though we are dealing with byte-array streams that do no require closing, as a matter of coding style I prefer to have them in try-with-resource blocks:
```java
try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
    ...
    try (InputStream inStream = new ByteArrayInputStream(outStream.toByteArray())) {
         ...
         return ....;
    }
}
```

sshd-common/src/main/java/org/apache/sshd/common/keyprovider/FileHostKeyCertificateProvider.java
```diff
@@ -0,0 +1,88 @@
...
+            }
+        }
+
+        return certificates;
+    }
+
+    @Override
+    public OpenSshCertificate loadCertificate(SessionContext session, String keyType) throws IOException, GeneralSecurityException {
+        return StreamSupport.stream(loadCertificates(session).spliterator(), false)
+                .filter(pubKey -> pubKey.getKeyType().equals(keyType))
```
lgoldstein, line 85: Let;s use: `filter(pubKey -> Objects.equals(keyType, pubKey.getKeyType()))`

sshd-common/src/main/java/org/apache/sshd/common/keyprovider/KeyPairProvider.java
```diff
@@ -122,7 +132,7 @@ default KeyPair loadKey(SessionContext session, String type)
     /**
      * @param session The {@link SessionContext} for invoking this load command - may
      * be {@code null} if not invoked within a session context (e.g., offline tool).
-     * @return The available {@link Iterable} key types in preferred order - never {@code null}
```
lgoldstein, line 125: Please note that **order** is assumed here so to remove this documentation and/or behavior is a major change - especially for KEX.

sshd-core/src/main/java/org/apache/sshd/client/kex/DHGClient.java
```diff
@@ -154,4 +168,54 @@ public boolean next(int cmd, Buffer buffer) throws Exception {
...
+        }
+
+        if (openSshKey.getType() != OpenSshCertificate.SSH_CERT_TYPE_HOST) {
+            throw new SshException(SshConstants.SSH2_DISCONNECT_KEY_EXCHANGE_FAILED,
+                    "KeyExchange signature verification failed, not a host key (2): "
+                            + openSshKey.getType());
+        }
+
+        long now = System.currentTimeMillis() / 1000;
+        if (now <= openSshKey.getValidAfter() || now >= openSshKey.getValidBefore()) {
```
lgoldstein, line 192: This condition seems wrong - shouldn't it be the other way around ? I.e., it is valid if `before <= now <= after`

sshd-core/src/main/java/org/apache/sshd/client/kex/DHGClient.java
```diff
@@ -154,4 +168,54 @@ public boolean next(int cmd, Buffer buffer) throws Exception {
...
+                            + openSshKey.getValidAfter() + "-" + openSshKey.getValidBefore());
+        }
+
+        SocketAddress connectSocketAddress = getClientSession().getConnectAddress();
+        if (connectSocketAddress instanceof SshdSocketAddress) {
+            connectSocketAddress = ((SshdSocketAddress) connectSocketAddress).toInetSocketAddress();
+        }
+        if (connectSocketAddress instanceof InetSocketAddress) {
+            String hostName = ((InetSocketAddress) connectSocketAddress).getHostString();
+            if (!openSshKey.getPrincipals().contains(hostName)) {
```
lgoldstein, line 204: What if `getPrincipals` returns _null_ ? Perhaps better:
```java
Collection<String> principals = key.getPrincipals();
if (GenericUtils.isEmpty(principals) || (!principals.contains(hostName))) {
   throw...
}
```

lgoldstein, line 204: The code is assuming that the host name is the connect address - but what if user used an  IP address instead of a fully-qualified host name to connect ? Do we do a reverse DNS ? (I  am not too keen on it).

FliegenKLATSCH: The IP would have to be in the principals of the certificate in the case you want to connect with the IP. OpenSSH also does not do a reverse lookup and I also would not expect this.

lgoldstein: Good to know - let's place a comment in the code explaining this

sshd-common/src/main/java/org/apache/sshd/common/keyprovider/KeyPairProvider.java
FliegenKLATSCH: This diff is very misleading. The line 125 was not changed, the change was in the doc of org.apache.sshd.common.keyprovider.KeyPairProvider#getKeyTypes in line 135, which returns a HashSet, which does NOT keep any order. The order is important in the other "part" later in AbstractServerSession: "expected" vs. "available" signatures...

sshd-core/src/main/java/org/apache/sshd/client/kex/DHGClient.java
FliegenKLATSCH: >A certificate is considered valid if:
The <= should be a strict <, but with the negation it should be correct besides this.

sshd-core/src/main/java/org/apache/sshd/server/session/AbstractServerSession.java
FliegenKLATSCH: Not sure if I understand the issue. The call to `kpp.getKeyTypes(this)` did not change.
Loading all provided certificates should not be an issue and is required to get their type? (They should not be password protected since they are public.)

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificate.java
FliegenKLATSCH: I would prefer the builder over setters, because IMO it should not be possible to change parts of the certificate later on, wdyt?

sshd-core/src/main/java/org/apache/sshd/client/kex/DHGClient.java
lgoldstein: Yes - written this way seems clearer

sshd-core/src/main/java/org/apache/sshd/server/session/AbstractServerSession.java
lgoldstein: The idea is that we do not want to iterate over the returned key types until really necessary - SSHD-860 implements "lazy" iterators that load keys only when 'next()' is called.  Therefore 
```java
StreamSupport.stream(kpp.getKeyTypes(this).spliterator(), false).collect(Collectors.toSet());
```
violates this "laziness" since it collects all the key types - something we wanted to avoid doing in SSHD-860

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificate.java
lgoldstein: Let me suggest sort of a compromise:

* Define `OpenSshCertificate` to be an **interface** that exposes only getters
* Define a class `OpenSshCertificateImpl` that has getters and setters and implements `OpenSshCertificate`
* Leave all the current code that checks `instanceof OpenSshCertificate`, casts to it an uses its getters as-is
* Remove the builder and use the `OpenSshCertificateImpl` to create an `OpenSshCertificate` instance when needed

This way the code maintains its "read-only" exposure of the getters, while at the same time providing more flexibility for mocks, future extensions, other implementations, etc....

sshd-common/src/main/java/org/apache/sshd/common/keyprovider/KeyPairProvider.java
lgoldstein: After reviewing your comment I  can live with the new phrasing

sshd-core/src/main/java/org/apache/sshd/server/session/AbstractServerSession.java
FliegenKLATSCH: Yes, but the call to `kpp.getKeyTypes(this)` did not change - it was there already before. I iterate only over an already existing `Set` of `String`s. That doesn't change anything regarding this lazy loading? Please correct me, if I am mistaken..

lgoldstein: I have reviewed the code and reached the conclusion that you are right - "laziness" is not an issue in this case

lgoldstein:
Just a reminder - please rebase your branch on the latest master version (I pushed some changes in the last few days) so it will be easier to merge this change.

I would still love to see some jUnit test code for the following scenarios:

* "Mixed" certificate and regular keys are available - make sure the certificate availability does not interfere with the "legacy" behavior
* Only certificate keys are available
* Bad certificate code behavior - make sure that if the certificate is invalid then code indeed disconnects as required

lgoldstein:
Seems in order - I am willing to merge it in as-is (provided you have re-based it) and wait for a separate PR with unit tests - let me know how you would like to proceed.

FliegenKLATSCH:
> * Bad certificate code behavior - make sure that if the certificate is invalid then code indeed disconnects as required

OpenSSH actually does a fallback to the plain host key, maybe we should do the same instead of aborting the connection if the certificate is invalid. Makes especially sense if the certificate is expired, you still want to be able to connect..
> Certificate invalid: name is not a listed principal
debug1: No matching CA found. Retry with plain key

I am currently on the unit tests, having some issues with RSA key mismatch exception... 512 vs 256 .. need to investigate further...

lgoldstein:
>> OpenSSH actually does a fallback to the plain host key, maybe we should do the same instead of aborting the connection if the certificate is invalid. Makes especially sense if the certificate is expired, you still want to be able to connect..

I can live with that - just a suggestion - you can make the behavior configurable via a property that you can retrieve from the session (default can be whatever you  decide).

>> I am currently on the unit tests, having some issues with RSA key mismatch exception... 512 vs 256 .. need to investigate further...

Great -  will wait for you to let me know when you feel the code is ready for more review and merging.

FliegenKLATSCH:
Please note commit https://github.com/apache/mina-sshd/pull/119/commits/a8cdbdec3fe6c7c3e248ec1854b4f0adc27c5863 This relates to [SSHD-895](https://issues.apache.org/jira/browse/SSHD-895), if client and server negotiated e.g. rsa-sha512 the signature verification would fail with ssh-rsa retrieved from `KeyUtils.getKeyType`.

lgoldstein:
Merged with acknowledgement and many thanks (see [47f779f06cb345c7cb706cb81f1214c37dab1fda](https://github.com/apache/mina-sshd/commit/47f779f06cb345c7cb706cb81f1214c37dab1fda)). Please note that I added a [commit](https://github.com/apache/mina-sshd/commit/7ebb469ea86460878abd009f357992378e03bde8) to fix some minor styling issues as well as update the relevant documentation.

Please close this PR

FliegenKLATSCH:
Thanks a lot, but it's not only server side support but also client side (I started with client side actually and had two different commits..).
And according to your latest comment in SSHD-895 we should revert the changes in ClientBuilder?
(Although I am still convinced that we should even add sha256/512 to have the broadest support and best security per default..)

lgoldstein:
>> Thanks a lot, but it's not only server side support but also client side (I started with client side actually and had two different commits..).

Sorry - too late now - the important thing is that the code is in

>> And according to your latest comment in SSHD-895 we should revert the changes in ClientBuilder?

I have reviewed the code before merging and have reached the conclusion that I can live with the change you suggested - i.e., include the certified keys by default. I have decided though not to include the `ssh-rsa-sah256` yet since it is still a rather recent feature.

FliegenKLATSCH:
Ok, thanks. Changelog might just be misleading.

alex-sherwin:
Do you know what is missing from this to make client certificates work?

I've been hacking away for a few nights to try and figure it out, and it feels like it nearly works if you simply remove the hard-coded check that only allows decoding the host cert variant here https://github.com/apache/mina-sshd/commit/47f779f06cb345c7cb706cb81f1214c37dab1fda#diff-4067485581e7df0fec6ea6f408772edbbcd5eb7a1d9833e5ed9130a1de9015d1R74

Then things nearly "just work", but currently I'm trying to debug the publickey auth process in https://tools.ietf.org/html/rfc4252#section-7

The mina client will send the RSA-CERT in the first SSH_MSG_USERAUTH_REQUEST, and the openssh server accepts it, but the followup SSH_MSG_USERAUTH_REQUEST that includes a signature fails on the openssh server side with

```
debug3: mm_answer_keyverify: publickey 0x7f2e7882b0a0 signature unverified: key type does not match
Failed publickey for user01 from 172.17.0.1 port 64500 ssh2: RSA-CERT SHA256:gBrWshYSkWIDTUkeNmUWQaCsyO+NIGVsQljoeL64+bU ID user01 (serial 0) CA RSA SHA256:6vk74tKCkS3WqaeL68mM2TMvnCyaFsqt1ac26fktSqQ
```

It to me like the mina client code it doing all the right things (I think), the instructions on https://tools.ietf.org/html/rfc4252#section-7 say that "The 'public key blob' may contain certificates" and the changes in this PR for Buffer encode the client cert into the Buffer when UserAuthPublicKey invokes `buffer.putPublicKey(pubKey)` where pubKey is a `OpenSshCertificateImpl` instance

Quick recap:

1. If you disable the hard-coded check that only decodes host certs
2. Setup a client whose `KeyPair` is an instance of the clients private key and certificate-based public key (the cert-based client pub key is parsed with `PublicKeyEntry.parsePublicKeyEntry(str).resolvePublicKey(null, null, null)`
3. Connect to an OpenSSH server

It seems to do all the right things (I think), but something appears to be wrong with the `SSH_MSG_USERAUTH_REQUEST` that contains a signature, and I'm not sure where the right place is to really figure out where this mismatch is.

It's worth nothing that if I compare the verbose OpenSSH server-side output when compared with a a OpenSSH client making a successful client-based RSA-CERT connection and mina client, the output is *identical* right up until the error from above:

```
debug3: mm_answer_keyverify: publickey 0x7f2e7882b0a0 signature unverified: key type does not match
Failed publickey for user01 from 172.17.0.1 port 64500 ssh2: RSA-CERT SHA256:gBrWshYSkWIDTUkeNmUWQaCsyO+NIGVsQljoeL64+bU ID user01 (serial 0) CA RSA SHA256:6vk74tKCkS3WqaeL68mM2TMvnCyaFsqt1ac26fktSqQ
```

All the displayed public keys and signatures and algorithms displayed in the verbose OpenSSH server-side output are identical up to here

If you could provide any guidance on what docs to read or any insights on what the problem may be, that would be great

Thanks,

tomaswolf:
What about verbose client-side output? Check what ssh -vvv tells you about the signature algorithm being used on the client side, and compare against the debug logging from the Apache sshd client.

I think the error "key type does not match" comes from [ssh-rsc.c](https://github.com/openssh/openssh-portable/blob/48f54b9/ssh-rsa.c#L270). `sigtype` should be one of ssh-rsa, rsa-sha2-256, or rsa-sha2-512 at that point. My suspicion is that the client by mistake sends "ssh-rsa-cert-v01@openssh.com" instead. If the client's debug log shows something like "send SSH_MSG_USERAUTH_REQUEST request ... type=ssh-rsa-cert-v01@openssh.com - fingerprint ..." then this is the problem.

tomaswolf:
Looking deeper into the Apache MINA sshd code, I think the problem is in `UserAuthPublicKey.appendSignature()`:

```
        bs.putByte(SshConstants.SSH_MSG_USERAUTH_REQUEST);
        bs.putString(username);
        bs.putString(service);
        bs.putString(name);
        bs.putBoolean(true);
        bs.putString(algo); // <-- This should be the key type
        bs.putPublicKey(key);

        byte[] contents = bs.getCompactData();
        byte[] sig;
        try {
            Map.Entry<String, byte[]> result = current.sign(session, algo, contents);
            String factoryName = result.getKey();
            // An RSA -cert... signature generates a ssh-rsa, rsa-sha2-256, or rsa-sha2-256 signature, so this check may not
            // make sense?
            ValidateUtils.checkState(algo.equalsIgnoreCase(factoryName),
                    "Mismatched signature type generated: requested=%s, used=%s", algo, factoryName);
            sig = result.getValue();
        } catch (Error e) {
            warn("appendSignature({})[{}][{}] failed ({}) to sign contents using {}: {}",
                    session, service, name, e.getClass().getSimpleName(), algo, e.getMessage(), e);
            throw new RuntimeSshException(e);
        }

        // Trace logging omitted here

        bs.clear();
        bs.putString(algo); // <-- But here we should have the signature algorithm name, shouldn't we?
        bs.putBytes(sig);
```

alex-sherwin:
Hi @tomaswolf 

Yes this is the problem, I had debugged the native OpenSSH client code and checked the raw buffer differences to come down to that same conclusion

I haven't put together a formal PR for this yet but I plan to soon, but I'm still not entirely sure what the right process is for determining the right value to put here for the algo value.  I can make it work with known, contrived values (`rsa-sha2-512` for rsa client certs and `ssh-ed25519` for ed25519 client certs), but I haven't spent the time yet to find out the "right" way to determine this value dynamically yet

The followup, once that's determined, will be how to properly unit test it since mina sshd server-side doesn't support client cert based auth.  I haven't investigated yet if it's plausible to unit test the client behavior in isolation or if the project will want complete end-to-end testing, which may require me to implement the server-side portion of this as well (I hope not, as I don't need that functionality)

alex-sherwin:
I've researched this some more and I don't think there's any place in the current mina-sshd project that maintains a map of these signature algorithms that are utilized, at least, not aliased to the values required here.

The `SignatureRSA` class has an internal `sshAlgorithmName` that holds the in-use algo (`rsa-sha2-512` / `rsa-sha2-256`), but the `SignatureEd25519` (which is third party anyways) has no similar internal field to even carry the `ssh-ed25519` value), for example

Best I can tell, it may be necessary to create a specific map of the signature values from the original type to the value required for this part of the `SSH_MSG_USERAUTH_REQUEST` response buf

alex-sherwin:
I've cleaned up an implementation that works well for all currently supported OpenSSH certificate formats on my fork (diff preview: https://github.com/apache/mina-sshd/compare/master...alex-sherwin:master)

I'm currently just using a hacky integration test that tests all valid OpenSSH client cert types against a local OpenSSH sshd daemon (which are all working from MINA ssh client now):

* `rsa-sha2-256-cert-v01@openssh.com`
* `rsa-sha2-512-cert-v01@openssh.com`
* `ecdsa-sha2-nistp256-cert-v01@openssh.com`
* `ecdsa-sha2-nistp384-cert-v01@openssh.com`
* `ecdsa-sha2-nistp521-cert-v01@openssh.com`
* `ssh-ed25519-cert-v01@openssh.com`

I'm not testing `ssh-dss-cert-v01@openssh.com` since it's not enabled in OpenSSH by default anymore.

I added function to `KeyUtils.getCertificateSignatureAlgorithm` which is able to determine the correct value to use for the signature in the signed `SSH_MSG_USERAUTH_REQUEST` buffer for the supported certificate types

It also simply removes the host key type check in the OpenSSH certificate decoding, which didn't really need to be there (other then that currently only host certs were supported for any functionality).  Perhaps this explicit check should be moved into the code path for the host key usage (but it appears there's already some filtering going on to discover host keys by type, so this check seems superficial)

I've love to clean up the rest of this and make a PR, but since all client unit test code I can see uses MINA-based `SshServer` instances to test against, I'm not sure how (in the current unit testing patterns) that this OpenSSH client certificate publickey auth could be covered in unit tests since the MINA sshd server doesn't support OpenSSH client certificates for publickey auth

Do you have any suggestion for that?  Or will I need to implement the server-side portion of this in tandem for the PR?

Thanks,

tomaswolf:
This is definitely going in the right direction, but will need more work before it could be incorporated into the main code.

As for testing: I think testing against a real OpenSSH server would be great. Testing only against an Apache MINA sshd server is better than nothing, but it's inbreeding and actually only shows that both our client and server side can talk to each other. It doesn't show that either implements the SSH protocol correctly; they might both implement their own completely different protocol. But I don't know if the containers used for the CI builds on Travis do actually have OpenSSH, nor do I know what the best way to configure, create and tear down an OpenSSH server during tests would be. We do run tests on Windows; I suspect there at least we might not have an OpenSSH server available. Perhaps @lgoldstein has an idea...

Some detail comments on your code: (can't comment directly on the preview diff)

- Comparing -cert keys should probably also compare the signatures.
- There really needs to be some mapping from the -cert signature algorithm names to the names of the underlying real signature. But I'd try to avoid simply stripping off the -cert-v01@openssh.com suffix. A explicit mapping table might be clearer. 
- OpenSSH has some restrictions on which signature algorithms it accepts for which -cert key; see [ssh-rsa.c, line 274](https://github.com/openssh/openssh-portable/blob/48f54b9/ssh-rsa.c#L270). In general it must be the same as used for the certificate, except for ssh-rsa-cert-v01@openssh.com, where it may be any of ssh-rsa, rsa-sha2-256, or rsa-sha2-512. But for a rsa-sha2-256-cert-v01 key, for instance, it _must_ be rsa-sha2-256. I suspect the determination of eligible algorithms in [UserAuthPublicKey.sendAuthDataRquest()](https://github.com/apache/mina-sshd/blob/master/sshd-core/src/main/java/org/apache/sshd/client/auth/pubkey/UserAuthPublicKey.java#L165) might also need to be changed. I suspect with the current "alias"-based way, we'd try for an rsa-sha2-256-cert-v01 key first rsa-sha2-512, get a failure, and only then re-try with rsa-sha2-256.

gnodet:
> This is definitely going in the right direction, but will need more work before it could be incorporated into the main code.
> 
> As for testing: I think testing against a real OpenSSH server would be great. Testing only against an Apache MINA sshd server is better than nothing, but it's inbreeding and actually only shows that both our client and server side can talk to each other. It doesn't show that either implements the SSH protocol correctly; they might both implement their own completely different protocol. But I don't know if the containers used for the CI builds on Travis do actually have OpenSSH, nor do I know what the best way to configure, create and tear down an OpenSSH server during tests would be. We do run tests on Windows; I suspect there at least we might not have an OpenSSH server available. Perhaps @lgoldstein has an idea...

Afaik, we don't have any client side test that can use an SSH server other than ours.  We could work on that by using `testcontainers`.  We're already using it for SFTP to test performance / latency, but it should be easy to setup an OpenSSH server and use it for testing.

gnodet:
@alex-sherwin it could be a good idea to open a different issue as this PR is actually closed.

alex-sherwin:
@gnodet @tomaswolf 

Sure, I didn't mean to start a whole thread here, was originally just looking to solicit info on what may be missing from this PR's original changes

Since it sounds like you're receptive to this change in general I'm happy to put together an initial PR tonight. 

I did see that testcontainers was already in use for the SFTP performance test, looked like a one-off thing to me so I wasn't sure if it was something the project wanted to adopt or not.  But I totally agree that I think it makes sense to test against OpenSSH sshd so that you're testing mina client code against a different implementation.

I have a lot of experience using testcontainers so I'm more then happy to use it in some unit tests for this PR

Thanks for your input so far

alex-sherwin:
@gnodet @tomaswolf I've opened a JIRA here https://issues.apache.org/jira/browse/SSHD-1161 for this work

A colleague & myself should hopefully have a PR ready in the next day or two

noskojan:
Hello,
I am wondering is the support of using SSH client certificate authentication available in Mina client, please? Since the PR is already closed, but issues SSHD-1161 is still opened. Also, I am not able to find any documentation for this. Can you please help me to understand if Mina SSH client is capable of using this kind of authentication or not?
Thanks

alex-sherwin:
@noskojan 

I'm not a maintainer here, but, FYI, the features have all been merged and are covered by unit tests

However there has been no stable release since those PR's were merged, and best I can tell the current 2.9.0-SNAPSHOT version has not been published to maven central (looking at https://repository.apache.org/content/groups/snapshots/org/apache/sshd/sshd-core/)

So you'd have to build 2.9.0-SNAPSHOT yourself to use it currently

lgoldstein:
I believe we merged these changes into the released 2.8.0 version

PR #120, opened 2020-04-11
Increase log level to error for scp error logs

The-Yoda:


lgoldstein:
While I understand that it may solve some specific problem you are having and perhaps the advantages of such a change, this change is not according to the (informal but widely observed) logging policy where stack trace details are either DEBUG or TRACE level (usually TRACE). I would be inclined to go along with a _protected_ method that does the logging (by default same as before), but that would also require major code changes in order to standardize the logging policy not only in this case but throughout the code.

The-Yoda:
There are cases where scp fails from different set of servers which is very hard to debug. And making the error log trace creates lot of friction, as it creates GBs of logs in a minute in production servers. It make sense to log exception in hotspots, as it incurs performance penalty, but not here.
Let me know how I can make this better. Should I just make the debug statement(just above the change) to error so that there won't be any logging of stack trace?

lgoldstein:
>> There are cases where scp fails from different set of servers which is very hard to debug. And making the error log trace creates lot of friction, as it creates GBs of logs in a minute in production servers. It make sense to log exception in hotspots, as it incurs performance penalty, but not here.
Let me know how I can make this better. Should I just make the debug statement(just above the change) to error so that there won't be any logging of stack trace?

I understand this and will have to think about it some more... however while I am doing this

>> making the error log trace creates lot of friction, as it creates GBs of logs in a minute in production servers

You don't have to enable logging for `org.apache.sshd` - you could enable only the relevant SCP package

gnodet:
A few thoughts:
  * we could check if the log is at debug or trace and print the stack trace at error level, else issue the error statement with just the exception's string representation
  * have 2 statements: one at error level without the stack trace and another at debug level with the stack trace

lgoldstein:
I am not sure I understand exactly how this would work - however, my main concern is not specifically with this code but rather our overall logging policy. On one hand, we do want to have abnormal situations logged with enough detail to allow debugging, but on the other hand we do not want to "pour" a lot of log messages - especially long stack traces.

>> issue the error statement with just the exception's string representation

This sounds reasonable, except that unfortunately many exceptions are thrown without any explanatory message. Too often have I seen `IOException: null` - i.e., the code threw `IOException` with no explanation as to what was wrong. Furthermore, I am not sure what `the exception's string representation`means - if it means `toString` then I  think it includes the stack trace - which I am not sure is a good idea.

lgoldstein:
I have been considering perhaps using a dedicated logger for exceptions where in addition to logging as we currently do we also invoke this logger - e.g.:

```java
try {
    ... some code...
} catch (SomeException  e) {
   ...log as usual...
    LoggingUtils.logException(getClass(), "....message...",  e);
}
```

and then have  some mechanism by which users can intercept the calls to `LoggingUtils#logException` and decide for themselves at which level to log and what details to include.

The-Yoda:
Provided that it is a utility with only static methods, I'm not sure how user can intercept `LoggingUtils.logException(getClass(), "....message...",  e);`.

One way I could think of is plugging in the logging related code while configuring while creating the server or client(similar to EventListener). But this creates lot of churn to the user to go through all the trouble to add logging and they won't be aware of it. 
Also, there will be hotspots with which user configuration will create more noise. So, they will need to do at class level. And it wouldn't be a good idea to let the user know what is there underneath.
In short, giving control of logging errors may create more problems than it solves.

The-Yoda:
Now updated the code similar to @gnodet suggestion.

lgoldstein:
I can live with this... will merge soon

lgoldstein:
Merged - see commit [4320886b19c8385ac017945d708680fad9c46f04](https://github.com/apache/mina-sshd/commit/4320886b19c8385ac017945d708680fad9c46f04) - can close the PR

PR #123, opened 2020-04-20
SFTP transfer speeds, clean API

gnodet:


sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/DefaultSftpClient.java
```diff
@@ -501,4 +484,74 @@ public int negotiateVersion(SftpVersionSelector selector) throws IOException {
         versionHolder.set(selected);
         return selected;
     }
+
+    private class SftpChannelSubsystem extends ChannelSubsystem {
```
lgoldstein, line 488: Let's make it protected (and the constructor) in accordance with our policy to allow maximum flexibility to users who wish to modify/enhance the behavior of the relevant class

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/SftpInputStreamAsync.java
```diff
@@ -0,0 +1,312 @@
...
+import org.apache.sshd.common.SshConstants;
+import org.apache.sshd.common.subsystem.sftp.SftpConstants;
+import org.apache.sshd.common.subsystem.sftp.SftpHelper;
+import org.apache.sshd.common.util.buffer.Buffer;
+import org.apache.sshd.common.util.buffer.ByteArrayBuffer;
+import org.apache.sshd.common.util.io.InputStreamWithChannel;
+
+public class SftpInputStreamAsync extends InputStreamWithChannel {
+
+    static class Ack {
```
lgoldstein, line 43: Let's make this class public in accordance with our policy to allow maximum flexibility to users who wish to modify/enhance the behavior of the relevant class

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/SftpInputStreamAsync.java
```diff
@@ -0,0 +1,312 @@
...
+        int id;
+        long offset;
+        int length;
```
lgoldstein, line 46: Should be public + final

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/SftpInputStreamAsync.java
```diff
@@ -0,0 +1,312 @@
...
+                }
+                buffer.rpos(buffer.rpos() + toRead);
+                clientOffset += toRead;
+                max -= toRead;
+            }
+        }
+        return clientOffset - orgOffset;
+    }
+
+    @SuppressWarnings("PMD.MissingOverride")
```
lgoldstein, line 182: If an `@Override` is in order here why omit it and suppress the warning ?

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/SftpInputStreamAsync.java
```diff
@@ -0,0 +1,312 @@
...
+            throw new IOException("skip(" + getPath() + ") stream closed");
+        }
+        if (clientOffset == 0 && pendingReads.isEmpty()) {
+            clientOffset = n;
+            return n;
+        }
+        return super.skip(n);
+    }
+
+    boolean hasNoData() {
```
lgoldstein, line 217: Let's make it protected in accordance with our policy to allow maximum flexibility to users who wish to modify/enhance the behavior of the relevant class

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/SftpInputStreamAsync.java
```diff
@@ -0,0 +1,312 @@
...
+        return buffer == null || buffer.available() == 0;
+    }
+
+    void sendRequests() throws IOException {
```
lgoldstein, line 221: Let's make it protected in accordance with our policy to allow maximum flexibility to users who wish to modify/enhance the behavior of the relevant class

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/SftpInputStreamAsync.java
```diff
@@ -0,0 +1,312 @@
...
+                buf.putLong(requestOffset);
+                buf.putInt(bufferSize);
+                int reqId = client.send(SftpConstants.SSH_FXP_READ, buf);
+                pendingReads.add(new Ack(reqId, requestOffset, bufferSize));
+                requestOffset += bufferSize;
+            }
+        }
+    }
+
+    void fillData() throws IOException {
```
lgoldstein, line 240: Let's make it protected in accordance with our policy to allow maximum flexibility to users who wish to modify/enhance the behavior of the relevant class

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/SftpInputStreamAsync.java
```diff
@@ -0,0 +1,312 @@
...
+                    eofIndicator = dlen < 0 || eof.get() != null && eof.get();
+                    cur += dlen;
+                }
+                buffer.getRawBytes(data, nb, buffer.available());
+                buffer = new ByteArrayBuffer(data);
+            }
+        }
+    }
+
+    void pollBuffer(Ack ack) throws IOException {
```
lgoldstein, line 262: Let's make it protected in accordance with our policy to allow maximum flexibility to users who wish to modify/enhance the behavior of the relevant class

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/SftpOutputStreamAsync.java
```diff
@@ -0,0 +1,201 @@
...
+import org.apache.sshd.common.util.io.OutputStreamWithChannel;
+
+/**
+ * Implements an output stream for a given remote file
+ *
+ * @author <a href="mailto:dev@mina.apache.org">Apache MINA SSHD Project</a>
+ */
+public class SftpOutputStreamAsync extends OutputStreamWithChannel {
+
+    static class Ack {
```
lgoldstein, line 43: Let's make it public in accordance with our policy to allow maximum flexibility to users who wish to modify/enhance the behavior of the relevant class

Isn't this exactly the same class as used by `SftpInputStreamAsync` ?

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/SftpOutputStreamAsync.java
```diff
@@ -0,0 +1,201 @@
...
+        int id;
+        long offset;
+        int length;
```
lgoldstein, line 46: Should be public + final

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/SftpRemotePathChannel.java
```diff
@@ -230,6 +227,18 @@ public long write(ByteBuffer[] srcs, int offset, int length) throws IOException
...
+        int id;
+        long offset;
+        int length;
+
+        Ack(int id, long offset, int length) {
+            this.id = id;
+            this.offset = offset;
+            this.length = length;
+        }
+    }
```
lgoldstein, line 240: See previous remarks regarding this identical class defined in other locations

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/SftpRemotePathChannel.java
```diff
@@ -346,30 +355,19 @@ public long transferTo(long position, long count, WritableByteChannel target) th
...
-                        }
-                        curPos += read;
-                        totalRead += read;
-                    } else {
-                        eof = read == -1;
-                    }
-                }
+                SftpInputStreamAsync input = new SftpInputStreamAsync(
+                        (AbstractSftpClient) sftp,
+                        copySize, position, count, getRemotePath(), handle);
```
lgoldstein, line 367: Shouldn't this be protected by a try-with-resource block (or similar) that closes it automatically when no longer needed (or exception thrown) ? I  see the comment that says not to close - but what if an exception is thrown ?

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/SftpRemotePathChannel.java
```diff
@@ -410,18 +408,23 @@ public long transferFrom(ReadableByteChannel src, long position, long count) thr
             try {
                 beginBlocking("transferFrom");
 
+                SftpOutputStreamAsync output = new SftpOutputStreamAsync(
+                        (AbstractSftpClient) sftp,
+                        copySize, getRemotePath(), handle);
```
lgoldstein, line 413: Shouldn't this be protected by a try-with-resource block (or similar) that closes it automatically when no longer needed (or exception thrown) ? I  see the comment that says not to close - but what if an exception is thrown ?

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/SftpRemotePathChannel.java
```diff
@@ -524,13 +527,13 @@ protected void endBlocking(Object actionHint, boolean completed)
      * @param  reqModes    The required modes - ignored if {@code null}/empty
      * @throws IOException If channel not open or the required modes are not satisfied
      */
-    private void ensureOpen(Collection<SftpClient.OpenMode> reqModes) throws IOException {
+    private void ensureOpen(Collection<OpenMode> reqModes) throws IOException {
```
lgoldstein, line 530: Let's make it protected in accordance with our policy to allow maximum flexibility to users.

sshd-sftp/src/test/java/org/apache/sshd/client/subsystem/sftp/SftpPerformanceTest.java
```diff
@@ -0,0 +1,243 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.junit.Ignore;
+import org.junit.Rule;
+import org.junit.Test;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.Network;
+import org.testcontainers.containers.ToxiproxyContainer;
+import org.testcontainers.containers.ToxiproxyContainer.ContainerProxy;
+
+@Ignore
```
lgoldstein, line 50: Recommend adding clarification text:
```java
@Ignore("Special class used for development only - not really a test just useful to run as such")
```

sshd-sftp/src/test/java/org/apache/sshd/client/subsystem/sftp/SftpTransferTest.java
```diff
@@ -0,0 +1,131 @@
...
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.Date;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.sshd.client.session.ClientSession;
+import org.apache.sshd.client.subsystem.sftp.fs.SftpFileSystem;
+import org.junit.Test;
+
+public class SftpTransferTest extends AbstractSftpClientTestSupport {
```
lgoldstein, line 34: Missing standard annotation `@FixMethodOrder(MethodSorters.NAME_ASCENDING)` that ensure reproducible test failures

sshd-sftp/pom.xml
```diff
@@ -85,6 +85,18 @@
...
+            <artifactId>testcontainers</artifactId>
+            <version>1.12.5</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>toxiproxy</artifactId>
+            <version>1.12.5</version>
+            <scope>test</scope>
+        </dependency>
```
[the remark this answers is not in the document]
bsideup: there is also a BOM:
`org.testcontainers:testcontainers-bom:1.14.0`

See
https://www.testcontainers.org/#managing-versions-for-multiple-testcontainers-dependencies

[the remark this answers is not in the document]
lgoldstein: Even better...

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/AbstractSftpClient.java
```diff
@@ -783,10 +786,6 @@ public void rename(String oldPath, String newPath, Collection<CopyMode> options)
     @Override
     public int read(Handle handle, long fileOffset, byte[] dst, int dstOffset, int len, AtomicReference<Boolean> eofSignalled)
             throws IOException {
-        if (eofSignalled != null) {
-            eofSignalled.set(null);
-        }
-
```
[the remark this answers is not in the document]
gnodet: We have `read` which calls `checkData` which calls `checkDataResponse`.  Having the 3 methods setting `eofSignalled` to null when entering while the only place where it can be set to a non null value is later inside the `checkDataResponse` seems a bit redundant to me.  Also, I think checking the EOF value in case an exception has been thrown is meaningless, so only keeping the `eofSignalled.set(null)` from inside `checkDataResponse` sounds enough to me.

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/AbstractSftpClient.java
```diff
@@ -802,9 +801,6 @@ public int read(Handle handle, long fileOffset, byte[] dst, int dstOffset, int l
     protected int checkData(
             int cmd, Buffer request, int dstOffset, byte[] dst, AtomicReference<Boolean> eofSignalled)
             throws IOException {
-        if (eofSignalled != null) {
-            eofSignalled.set(null);
-        }
```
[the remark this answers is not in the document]
gnodet: See above

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/SftpRemotePathChannel.java
gnodet: Calling `close` has 2 effects: blocking until the stream is fully consumed and closing the handle.  In this particular usage, the data is already consumed in the `transferTo` and we don't want to close the handle.  The `close` implementation is really focused on the standard usage of the `InputStream`, not this internal usage.  
If that really sounds too hacky, we can always use two different classes through inheritance, but I thought the comment would be sufficient.

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/SftpRemotePathChannel.java
gnodet: Same as above.

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/SftpRemotePathChannel.java
gnodet: Sure, but that's not part of my changes, so I'd rather use a different commit to fix issues in the existing code rather than mixing both.

sshd-sftp/src/test/java/org/apache/sshd/client/subsystem/sftp/SftpTransferTest.java
gnodet: From the `@FixMethodOrder` javadoc:
> the order of execution was changed to be deterministic (in JUnit 4.11)

I'll add it, but not sure when it's actually useful...

sshd-sftp/pom.xml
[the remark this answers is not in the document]
gnodet: Thx, I'll go for the BOM then...

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
```diff
@@ -954,7 +954,9 @@ protected IoWriteFuture doWritePacket(Buffer buffer) throws IOException {
         synchronized (encodeLock) {
             Buffer packet = resolveOutputPacket(buffer);
             IoSession networkSession = getIoSession();
-            return networkSession.writePacket(packet);
+            IoWriteFuture future = networkSession.writePacket(packet);
+            buffer.rpos(buffer.wpos());
+            return future;
```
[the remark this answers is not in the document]
gnodet: The code is the following:
```
        synchronized (encodeLock) {
            Buffer packet = resolveOutputPacket(buffer);
            IoSession networkSession = getIoSession();
            IoWriteFuture future = networkSession.writePacket(packet);
            buffer.rpos(buffer.wpos());
            return future;
        }
```

The `buffer` is actually read from within the call to `resolveOutputPacket` and encoded into the `packet` variable.  So while the `packet` is yet to be written, the `buffer` won't be used anymore and is consumed at the time the `buffer.rpos(buffer.wpos())` is called.

sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelAsyncOutputStream.java
```diff
@@ -107,9 +108,39 @@ protected synchronized void doWriteIfPossible(boolean resume) {
...
+                IoWriteFutureImpl f = new IoWriteFutureImpl(future.getId(), new ByteArrayBuffer(buffer.getCompactData()));
+                f.addListener(w -> future.setValue(w.getException() != null ? w.getException() : w.isWritten()));
+                pendingWrite.set(f);
+                if (log.isTraceEnabled()) {
+                    log.trace("doWriteIfPossible({})[resume={}] waiting for window space {}", this, resume,
+                            remoteWindow.getSize());
+                }
+            } else if (total > remoteWindow.getPacketSize()) {
+                if (buffer.rpos() > 0) {
+                    // do a defensive copy in case the user reuses the buffer
```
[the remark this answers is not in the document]
gnodet: Agreed, I'll split it somehow.

sshd-sftp/src/test/java/org/apache/sshd/client/subsystem/sftp/SftpTransferTest.java
lgoldstein: >> I'll add it, but not sure when it's actually useful...

My experience shows that it is possible for a unit test to fail due to some unforeseen dependency with some side-effect of some other test. However, if the order of the tests changes from one run to another (can happen if using a different JDK version) then the failure may become irreproducible. Therefore, the _surefire_ plugin is configured to run the classes in alphabetical order - but within each class we want to make sure the individual tests are also run in a predictable order. This way, if a unit test fails (for whatever reason - not only because of interaction with another test) we can re-produce it with a high probability.

Of course, the down side is that if there is some interaction we don't know about we might not discover it until by some chance the lexicographical order changes. This can be managed by having a **separate** build profile that randomizes the methods, lists the chosen permutation and then runs the tests in that particular order. This way, if there is some unknown interaction between the tests we might detect it eventually and then reproduce it using the listed permutation order.

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/SftpRemotePathChannel.java
lgoldstein: +1

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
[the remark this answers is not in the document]
lgoldstein: >>  So while the packet is yet to be written, the buffer won't be used anymore and is consumed at the time the buffer.rpos(buffer.wpos()) is called.

That was not what I meant - `networkSession#writePacket` might be asynchronous - i.e., posts the buffer for writing and returns immediately - so when it attempt to access `rpos` it might use the wrong one since we override it.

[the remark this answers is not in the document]
gnodet: Not sure to understand.  The `buffer` and `packet` variable are not the same objects.  I think we could also write:
```
  synchronized (encodeLock) {
            Buffer packet = resolveOutputPacket(buffer);
            buffer.rpos(buffer.wpos());
            IoSession networkSession = getIoSession();
            IoWriteFuture future = networkSession.writePacket(packet);
            return future;
        }
```
and it would also work.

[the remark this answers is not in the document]
lgoldstein: You are right - took me a bit to ensure that this is the case...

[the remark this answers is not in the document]
gnodet: Ok, so the problem is in the assumptions about the `writePacket`.  In the case of the nio2 implementation, it always wrap the data into a different buffer. Let me look a bit more into it.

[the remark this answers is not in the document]
lgoldstein: >> Ok, so the problem is in the assumptions about the writePacket.

I did not mean that there is any problem with the call since the packet is obviously a different instance than the original buffer. If there is some similar optimization that needs to be done on the packet then indeed we need to make it **after** being informed (via a listener) that the packet has been consumed.

PR #128, opened 2020-05-03
[SSHD-984] Writing keys in modern OpenSSH format

tomaswolf:
Add support for writing keys in the modern OpenSSH key format using
the OpenBSD bcrypt KDF for writing passphrase-protected, encrypted
private keys.

Add a new OpenSSHKeyPairResourceWriter using a specialized
OpenSSHKeyEncryptionContext to make the number of KDF rounds
configurable and to have the passphrase as a char[] instead of as a
String. Introduce a new SecureByteArrayOutputStream.

Includes test cases that generate various keys, write them to files,
and load them again. Since this only tests that Apache MINA sshd can
work with these keys, I've also manually verified that command-line
OpenSSH can use keys generated and written by Java to actually connect
to an SSH server.

### DIFF
--- a/sshd-common/src/main/java/org/apache/sshd/common/config/keys/PrivateKeyEntryDecoder.java
+++ b/sshd-common/src/main/java/org/apache/sshd/common/config/keys/PrivateKeyEntryDecoder.java
@@ -36,6 +36,7 @@ import org.apache.sshd.common.session.SessionContext;
 import org.apache.sshd.common.util.GenericUtils;
 import org.apache.sshd.common.util.NumberUtils;
 import org.apache.sshd.common.util.ValidateUtils;
+import org.apache.sshd.common.util.io.SecureByteArrayOutputStream;
 
 /**
  * @param  <PUB> Type of {@link PublicKey}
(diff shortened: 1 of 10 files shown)


tomaswolf:
Build fails because of some checkstyle rules, like having 6 private methods but checkstyle allowing only 5.

I'm not going to spend any time fighting against checkstyle. Feel free to amend this code until it matches your rules.

lgoldstein:
There are quite a few public API(s) whose argument is defined as `SecureByteArrayOutputStream` - I understand the rationale behind this class, but I don't quite understand why the restriction to those public API(s) to accept a `SecureByteArrayOutputStream` instead of a plain old `OutputStream`.

lgoldstein:
>> Build fails because of some checkstyle rules, like having 6 private methods but checkstyle allowing only 5. I'm not going to spend any time fighting against checkstyle. Feel free to amend this code until it matches your rules.

No problem - I will fix it when I merge the code.

lgoldstein:
>> have the passphrase as a char[] instead of as a String

I am not sure why many think that it is more secure than a string - but regardless, I would prefer to have it done system wide - see [SSHD-709](https://issues.apache.org/jira/browse/SSHD-709). While it is marked as "Won't fix" I am willing to re-open it.

tomaswolf:
Re char[] vs String: at least I can clear a char[] once the password isn't needed anymore. With a String I cannot do so, and sensitive data (the user-entered passphrase) can survive in memory for unspecified amounts of time, which some people consider a security risk.  If I were to write some ssh-keygen clone in Java, I might use Console.readPassword() to get the passphrase from the user and have char[] throughout. But I won't insist; if you prefer String, we can also drop this.

Re: API using SecureByteArrayOutputStream: again, limit leaking sensitive data (especially when one writes an unencrypted private key). Forces a user to use a mechanism to write keys to files as shown in the test: once the file is written, the byte array is cleared. (And the test doesn't use a buffered stream but a Channel directly.) With a general OutputStream it might be a tad simpler to write a key to a file, but one has no control over the various buffers involved. But again, feel free to change this to plain OutputStream, maybe with a recommendation in the javadoc so use some secure stream.

lgoldstein:
>>  But I won't insist; if you prefer String, we can also drop this.

Great - I re-opened SSHD-709 and would rather do it a system wide change

>> feel free to change this to plain OutputStream, maybe with a recommendation in the javadoc so use some secure stream.

Thx - I prefer it this way.

lgoldstein:
Merged - thanks for (yet another) contribution - please close the PR

PR #130, opened 2020-05-11
fixed the typo which caused 32 bytes to be reserved instead of 4

243826:


gnodet:
Retest this please.

gnodet:
I've reverted the merge and raise a broader issue https://issues.apache.org/jira/browse/SSHD-1010 in order to fix all references to `xxx.SIZE`.

243826:
really not a fan of this strategy since there is not a timely pathway to
get this fixed due to voluntary nature of resources involved. the code
without that fix a little bit lower quality than that with the fix now.

PR #134, opened 2020-05-19
Fix a possible problem where an sftp chunked packet could be lost

gnodet:
PR to validate the fix against CI

### DIFF
--- a/sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/AbstractSftpClient.java
+++ b/sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/AbstractSftpClient.java
@@ -1308,6 +1308,9 @@ public abstract class AbstractSftpClient extends AbstractSubsystemClient impleme
 
     @Override
     public InputStream read(String path, int bufferSize, Collection<OpenMode> mode) throws IOException {
+        if (bufferSize <= 0) {
+            bufferSize = getReadBufferSize();
+        }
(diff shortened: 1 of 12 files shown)


sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/AbstractSftpClient.java
```diff
@@ -1308,6 +1308,9 @@ public FileChannel openRemoteFileChannel(String path, Collection<OpenMode> modes
 
     @Override
     public InputStream read(String path, int bufferSize, Collection<OpenMode> mode) throws IOException {
+        if (bufferSize <= 0) {
+            bufferSize = (int) getClientChannel().getLocalWindow().getPacketSize() - 13;
```
lgoldstein, line 1312: Can we write a short comment why `-13`

sshd-sftp/src/main/java/org/apache/sshd/client/subsystem/sftp/impl/AbstractSftpClient.java
```diff
@@ -1329,6 +1332,9 @@ public InputStream read(String path, Collection<OpenMode> mode) throws IOExcepti
 
     @Override
     public OutputStream write(String path, int bufferSize, Collection<OpenMode> mode) throws IOException {
+        if (bufferSize <= 0) {
+            bufferSize = (int) getClientChannel().getRemoteWindow().getPacketSize() - 13;
```
lgoldstein, line 1336: Can we write a short comment why -13

PR #135, opened 2020-05-20
[SSHD-997] OpenSSH key reading: fix RSA and ed25519

tomaswolf:
ed25519: do not clear the seed read from a key file; the i2p library
references it in the generated private key.

RSA: use a RSAPrivateCrtKeySpec to generate the RSA key to ensure that
a key read from an OpenSSH file can also be written to an OpenSSH file.

Tests: add tests for round-tripping keys: write the generated key to a
file, read it again, and compare. Then write the read key to a second
file, read that, and compare again.


lgoldstein:
Thanks for the patch - merged in - can close this PR

PR #137, opened 2020-06-04
[SSHD-1003] Use asynchronous streams when forwarding ports

gnodet:


sshd-core/src/main/java/org/apache/sshd/common/io/nio2/Nio2Session.java
```diff
@@ -368,7 +368,6 @@ protected void handleReadCycleCompletion(
                     log.debug("handleReadCycleCompletion({}) Socket has been disconnected (result={}), closing IoSession now",
                             this, result);
                 }
-                close(true);
```
gnodet, line 371: This suppression seems to remove a bunch of problems with the channel being closed way too early.  The `-1` result indicates an EOF, but not an error, so closing the session is wrong.

PR #139, opened 2020-06-26
[SSHD-1024] Allow other signatures to use rsa variants

FliegenKLATSCH:
The diff looks big; I just removed the outer IF...


TWHOI:
The diff look big; I just removed the outer IF..

lgoldstein:
Thanks - merged in [b9ddc14863](https://github.com/apache/mina-sshd/commit/b9ddc148636a0c18a2dd16e4afd16c18d31594dd). Can close the PR

PR #140, opened 2020-07-13
Fix for [SSHD-1033]

ggrandes:
[Bug SSHD-1033]: Unable to use DynamicPortForwarding + LocalPortForwarding


ggrandes:
Something is wrong

PR #150, opened 2020-07-24
Add a test exercising the connection through an ssh proxy

gnodet:


sshd-core/src/test/java/org/apache/sshd/client/ProxyTest.java
```diff
@@ -0,0 +1,117 @@
...
+        return session;
+    }
+
+    @Override
+    protected SshServer setupTestServer() {
+        SshServer sshd = super.setupTestServer();
+        // setup forwarding filter to allow the local port forwarding
+        sshd.setForwardingFilter(AcceptAllForwardingFilter.INSTANCE);
+        // setup an echo command
+        sshd.setCommandFactory((session, command) -> new CommandExecutionHelper(command) {
```
tomaswolf, line 105: Would be better to set up the proxy server with no command factory, or with a factory doing something else. As is, one cannot tell whether the proxy or the server handled the command.

I'm also missing something that verifies that the request _did_ go through the proxy. As it's all on localhost, if some mis-wiring occurs somewhere, it might be possible that the client contacts the server directly...

sshd-core/src/test/java/org/apache/sshd/client/ProxyTest.java
```diff
@@ -0,0 +1,117 @@
...
+                    .verify(CONNECT_TIMEOUT).getSession();
+            proxySession.addPasswordIdentity(proxyPassword);
+            proxySession.auth().verify(AUTH_TIMEOUT);
+            SshdSocketAddress address = new SshdSocketAddress(host, port);
+            ExplicitPortForwardingTracker tracker = proxySession.createLocalPortForwardingTracker(
+                    SshdSocketAddress.LOCALHOST_ADDRESS, address);
+            SshdSocketAddress bound = tracker.getBoundAddress();
+            session = client.connect(user, bound.getHostName(), bound.getPort())
+                    .verify(CONNECT_TIMEOUT).getSession();
+            session.addCloseFutureListener(f -> IoUtils.closeQuietly(tracker));
```
tomaswolf, line 90: This listener should probably also close the proxySession?

sshd-core/src/test/java/org/apache/sshd/client/ProxyTest.java
```diff
@@ -0,0 +1,117 @@
...
+            try (ClientSession session = createSession(
+                    client,
+                    "localhost", server.getPort(), "user1", "user1",
+                    "localhost", proxy.getPort(), "user2", "user2")) {
+                ByteArrayOutputStream out = new ByteArrayOutputStream();
+                ByteArrayOutputStream errors = new ByteArrayOutputStream();
+                session.executeRemoteCommand(command, out, errors, StandardCharsets.UTF_8);
+                result = out.toString();
+            }
+            assertEquals(command, result);
```
tomaswolf, line 68: The proxySession is not closed here. I'd make proxySession a field and add here
```
assertTrue(proxySession == null || proxySession.isClosing() || proxySession.isClosed());
```

tomaswolf:
Cool, thanks! One more thing: can you try this with a real ServerKeyVerifier? Possibly even with different host keys for proxy and server? I suspect the test will then fail because of https://github.com/apache/mina-sshd/blob/master/sshd-core/src/main/java/org/apache/sshd/client/session/AbstractClientSession.java#L561, which would use the local end of the tunnel when trying to verify the server's host key.

gnodet:
@tomaswolf would you mind creating a patch / PR ? I've done a simple test and the client will have to accept both the proxy and the server keys, but I'm not sure what you'd like to see in the test.

tomaswolf:
OK, there you are: #152. I don't see how to set this up cleanly when there' a local forward in the mix. Something is missing in the SshClient interfaces to set this up properly.

I _did_ manage in the JGit SSH client; but it involves quite some changes. Basically I pass the tracker.getBoundAddress() via an attribute to connect(user, server.getHost(), server.getPort()). That way connect has both addresses, and can ensure that session.setConnectAddress() gets called with the server address, not with the bound address. (It also means that connect(user, host, port) will find the correct HostConfigEntry: the one for the server. For the tunnel's bound address, it'll never find anything.) getIoSession().getRemoteAddress() will still be the tracker's bound address though, so I also override checkKeys() to use getConnectAddress() instead.

This is all rather convoluted and not very nice, but I don't see clearly how to do better. Perhaps passing in the PortForwardingTracker, if any, to connect() might help. JGit code is at https://git.eclipse.org/r/c/jgit/jgit/+/166868 if you're interested.

PR #152, opened 2020-07-28
ProxyTest with HostKeyVerification

tomaswolf:
This just serves to show that this way of setting up a proxy won't
work with real host key verification. The session ends up trying to
find an entry for the local end of the tunnel to the server in the
knownhosts file, but of course there is none.

### DIFF
--- a/sshd-core/src/test/java/org/apache/sshd/client/ProxyTest.java
+++ b/sshd-core/src/test/java/org/apache/sshd/client/ProxyTest.java
@@ -18,12 +18,22 @@
  */
 package org.apache.sshd.client;
 
+import java.io.BufferedWriter;
 import java.io.ByteArrayOutputStream;
+import java.io.File;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.StandardOpenOption;
+import java.security.KeyPair;
 
+import org.apache.sshd.client.config.hosts.KnownHostHashValue;
+import org.apache.sshd.client.keyverifier.KnownHostsServerKeyVerifier;
+import org.apache.sshd.client.keyverifier.RejectAllServerKeyVerifier;
 import org.apache.sshd.client.session.ClientSession;
 import org.apache.sshd.client.session.forward.ExplicitPortForwardingTracker;
+import org.apache.sshd.common.config.keys.PublicKeyEntry;
+import org.apache.sshd.common.util.GenericUtils;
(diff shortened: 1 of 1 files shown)


tomaswolf:
This is a failing test; not intended to be merged as is.

tomaswolf:
@gnodet: BTW, note that the exception stack trace does not point back to user code but to `Nio2Session.handleReadCycleCompletion`. That's not helpful to debug such problems.

The reason for this is probably that `AbstractSshFuture.verifyResult()` re-throws the original exception. It would be more useful if that always threw a new `IOException` with the original as cause. At least for any future that is handed out to client code.

gnodet:
> The reason for this is probably that `AbstractSshFuture.verifyResult()` re-throws the original `IOException`. It would be more useful if that always threw a new `IOException` with the original as cause. At least for any future that is handed out to client code.

```
diff --git a/sshd-common/src/main/java/org/apache/sshd/common/future/AbstractSshFuture.java b/sshd-common/src/main/java/org/apache/sshd/common/future/AbstractSshFuture.java
index 48ab284e..c64bd7fb 100644
--- a/sshd-common/src/main/java/org/apache/sshd/common/future/AbstractSshFuture.java
+++ b/sshd-common/src/main/java/org/apache/sshd/common/future/AbstractSshFuture.java
@@ -131,15 +131,6 @@ public abstract class AbstractSshFuture<T extends SshFuture> extends AbstractLog

         if (Throwable.class.isAssignableFrom(actualType)) {
             Throwable t = GenericUtils.peelException((Throwable) value);
-            if (t != value) {
-                value = t;
-                actualType = value.getClass();
-            }
-
-            if (IOException.class.isAssignableFrom(actualType)) {
-                throw (IOException) value;
-            }
-
             Throwable cause = GenericUtils.resolveExceptionCause(t);
             throw formatExceptionMessage(
                     msg -> new SshException(msg, cause),
```

tomaswolf:
> ```
> diff --git a/sshd-common/src/main/java/org/apache/sshd/common/future/AbstractSshFuture.java ...
> [...]
> ```

Yes. But the javadoc would need to be adapted, too. (It explicitly says that IOEs are re-thrown directly.)

Also (and that's why I just mentioned it here instead of doing a PR for it): if there's long "chains" of futures, this may result in long exception cause chains. Don't know if that could be a problem.

gnodet:
The updated test should now be included in https://github.com/apache/mina-sshd/pull/154.

PR #153, opened 2020-07-29
Do not rethrow, but always wrap in an IOException

gnodet:


sshd-common/src/main/java/org/apache/sshd/common/future/AbstractSshFuture.java
```diff
@@ -131,15 +125,6 @@ public boolean awaitUninterruptibly(long timeoutMillis) {
...
             Throwable t = GenericUtils.peelException((Throwable) value);
-            if (t != value) {
-                value = t;
-                actualType = value.getClass();
-            }
-
-            if (IOException.class.isAssignableFrom(actualType)) {
-                throw (IOException) value;
-            }
-
```
tomaswolf, line 142: Looks like this should be
```
if (t != value) {
    value = t;
}

if (t instanceof SshException) {
    throw new SshException(((SshException) t).getDisconnectCode(), t.getMessage(), t);
}
```

gnodet: The `value` variable isn't used, but the disconnect code is important to keep, I agree.

PR #157, opened 2020-08-10
[SSHD-1004] Using a more constant time MAC validation to minimize timing side channel information leak

lgoldstein:


lgoldstein:
Good idea - will do

PR #158, opened 2020-08-14
[SSHD-1053] Fix OpenSSH client "key type does not match" error when use Mina SSHD with OpenSSH Host Public Key Cert

fengjiajie:
https://issues.apache.org/jira/browse/SSHD-1053

### DIFF
--- a/sshd-common/src/main/java/org/apache/sshd/common/config/keys/KeyUtils.java
+++ b/sshd-common/src/main/java/org/apache/sshd/common/config/keys/KeyUtils.java
@@ -80,6 +80,7 @@ import org.apache.sshd.common.keyprovider.KeyPairProvider;
 import org.apache.sshd.common.u2f.SkED25519PublicKey;
 import org.apache.sshd.common.u2f.SkEcdsaPublicKey;
 import org.apache.sshd.common.util.GenericUtils;
+import org.apache.sshd.common.util.MapEntryUtils.MapBuilder;
 import org.apache.sshd.common.util.MapEntryUtils.NavigableMapBuilder;
 import org.apache.sshd.common.util.OsUtils;
 import org.apache.sshd.common.util.ValidateUtils;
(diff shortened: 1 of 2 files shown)


sshd-common/src/main/java/org/apache/sshd/common/config/keys/KeyUtils.java
```diff
@@ -141,6 +142,13 @@
     public static final String RSA_SHA256_CERT_TYPE_ALIAS = "rsa-sha2-256-cert-v01@openssh.com";
     public static final String RSA_SHA512_CERT_TYPE_ALIAS = "rsa-sha2-512-cert-v01@openssh.com";
 
+    private static final Map<String, String> CERT_TYPE_HASH_ALGORITHM_TYPE
```
lgoldstein, line 145: This should be wrapped inside a `Collections.unmodifiableMap`

sshd-common/src/main/java/org/apache/sshd/common/config/keys/KeyUtils.java
```diff
@@ -1196,4 +1204,13 @@ public static boolean compareSkEd25519Keys(SkED25519PublicKey k1, SkED25519Publi
...
     }
+
+    public static String getHashAlgorithmType(String keyType) {
+        String hashAlgorithmType = CERT_TYPE_HASH_ALGORITHM_TYPE.get(keyType);
+        if (hashAlgorithmType != null) {
+            return hashAlgorithmType;
+        } else {
+            return keyType;
+        }
+    }
```
lgoldstein, line 1215: Code can be simplified as `return CERT_TYPE_HASH_ALGORITHM_TYPE.getOrDefault(keyYpe, keyType)`

lgoldstein:
There already is a `KeyUtils#getCanonicalKeyType` - can't it be used somehow ?

fengjiajie:
> There already is a `KeyUtils#getCanonicalKeyType` - can't it be used somehow ?

KeyUtils.getCanonicalKeyType("rsa-sha2-256-cert-v01@openssh.com") return "ssh-rsa-cert-v01@openssh.com"
but we need is "rsa-sha2-256"

lgoldstein:
What about applying this fix to `DHGEXServer, DHGClient, DHGEXClient` classes ? Please note that some code regarding certificate keys is already in place.

Also, can we have a unit test in place testing this ?

fengjiajie:
> What about applying this fix to `DHGEXServer, DHGClient, DHGEXClient` classes ? Please note that some code regarding certificate keys is already in place. I think the fix is not good enough and should be more along the lines of `DHGEXClient`-s handling of certificate keys.
> 
> Also, can we have a unit test in place testing this ?

I don't know much about the SSH protocol and I also don't think this is a good approach.
I located this code through debugging, and I do not really know the other classes do for(and even the one I modified I'm not quite clearing...), so it may need to be done by someone else or in the future as I learn more about it.

PR #162, opened 2020-08-25
[SSHD-1063] Fixed known-hosts file server key verifier matching of same host on different ports

lgoldstein:


ggrandes:
Works fine!

PR #163, opened 2020-09-06
[SSHD-1070] Limit read-ahead in BufferedIoOutputStream

tomaswolf:
Use a LinkedBlockingQueue to throttle a fast producer to the speed
of writing. Otherwise a fast producer might queue up too many write
futures with buffers, which might lead to an OOME.

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/common/channel/BufferedIoOutputStream.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/channel/BufferedIoOutputStream.java
@@ -20,8 +20,8 @@ package org.apache.sshd.common.channel;
 
 import java.io.EOFException;
 import java.io.IOException;
-import java.util.Queue;
-import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.sshd.common.Closeable;
@@ -36,7 +36,7 @@ import org.apache.sshd.common.util.closeable.AbstractInnerCloseable;
  */
 public class BufferedIoOutputStream extends AbstractInnerCloseable implements IoOutputStream {
     protected final IoOutputStream out;
-    protected final Queue<IoWriteFutureImpl> writes = new ConcurrentLinkedQueue<>();
+    protected final BlockingQueue<IoWriteFutureImpl> writes = new LinkedBlockingQueue<>(10);
     protected final AtomicReference<IoWriteFutureImpl> currentWrite = new AtomicReference<>();
     protected final Object id;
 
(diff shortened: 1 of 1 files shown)


tomaswolf:
Build failure appears to be unrelated. Maven build cannot find a particular maven plugin version.

sshd-core/src/main/java/org/apache/sshd/common/channel/BufferedIoOutputStream.java
```diff
@@ -56,7 +56,12 @@ public IoWriteFuture writePacket(Buffer buffer) throws IOException {
         }
 
         IoWriteFutureImpl future = new IoWriteFutureImpl(getId(), buffer);
-        writes.add(future);
+        try {
+            writes.put(future);
+        } catch (InterruptedException e) {
+            Thread.interrupted();
+            throw new IOException(e);
+        }
```
lgoldstein, line 64: This changes the behavior of the code in significant way `put` involves blocking the calling thread for and indeterminate amount of time:

>> Inserts the specified element into this queue, waiting if necessary for space to become available.

whereas `add` 

>> Inserts the specified element into this queue if it is possible to do so immediately without violating capacity restrictions, returning {@code true} upon success and throwing an {@code IllegalStateException} if no space is currently available.

which in original code means that it never blocks since we do not place any restriction. I would feel better if `offer` was used with a configurable timeout (defined in `CoreModuleProperties`):

```java
Duration duration = CoreModuleProperties.MAX_PENDING_FUTURE_ENQUEUE_WAIT.getRequired(this);
if (!writes.offer(future, duration.toMillis(), TimeUnit.MILLISECONDS)) {
    throw ...;
}
```

A similar property should be used to initialize the queue so that users can tailor the throughput limitation according to their needs.

lgoldstein (changes requested):

sshd-core/src/main/java/org/apache/sshd/common/channel/BufferedIoOutputStream.java
tomaswolf: I know about the semantics of `put`. Yes, it'll block indefinitely until there is space again - which will be when the slow consumer has caught up. It doesn't make any sense to let the producer continue any earlier. But maybe there's a better place to implement this throttle than in this `BufferedIoOutputStream`.

Your suggestion is a larger refactoring that I cannot do. Using `offer` with timeout one also would have to handle in the caller the case that the internal queue was full, and wait there, and there'd need to be a way to configure the queue size. Moreover the queue size and timeout needed would depend on the relative speeds of producer and consumer, which are unknown in general. I wouldn't know how to configure such a thing such that it worked for all cases. Just failing the transfer when the queue is ever full is IMO not an option; the whole mechanism should probably still work (albeit slowly) with a very fast producer and a very slow consumer.

The problem is precisely that current code doesn't place any restrictions, and thus a fast producer can overrun a slow consumer so much that the server gets an OOME.

But in any case SSHD-1070 and SSHD-1069 should probably be resolved together.

BTW, `Thread.interrupted()`is wrong; should have been `Thread.currentThread().interrupt()`.

tomaswolf:
Looks like Feng has come up with a better solution; see SSHD-1070.

PR #165, opened 2020-09-13
[SSHD-1076] Break down ClientUserAuthService#auth method into several to allow for flexible override

lgoldstein:


tomaswolf:
Looks good. Makes the class more flexible if a user has some specialized needs.

PR #166, opened 2020-09-14
[SSHD-1070] Limit the amount of data that is kept in memory for forwa…

gnodet:
…rding channels

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/common/io/nio2/Nio2Session.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/io/nio2/Nio2Session.java
@@ -387,32 +387,28 @@ public class Nio2Session extends AbstractCloseable implements IoSession {
 
     @Override
     public void suspendRead() {
-        log.debug("suspendRead({})", this);
-        suspendRead.set(true);
+        log.trace("suspendRead({})", this);
+        if (suspendRead.compareAndSet(false, true)) {
+            log.debug("suspendRead({}) requesting read suspension", this);
+        }
     }
 
     @Override
     public void resumeRead() {
+        log.trace("resumeRead({})", this);
         Runnable runnable = readRunnable.getAndSet(null);
+        suspendRead.set(false);
         if (runnable != null) {
-            log.debug("resumeRead({})", this);
-            suspendRead.set(false);
+            log.debug("resumeRead({}) resuming read", this);
             runnable.run();
         }
     }
 
     protected void doReadCycle(ByteBuffer buffer, Nio2CompletionHandler<Integer, Object> completion) {
         if (suspendRead.get()) {
-            log.debug("doReadCycle({}) suspending read cycle", this);
-            if (!readRunnable.compareAndSet(null, () -> doReadCycle(buffer, completion))) {
-                throw new IllegalStateException();
-            }
+            log.debug("doReadCycle({}) suspending reading", this);
+            readRunnable.set(() -> doReadCycle(buffer, completion));
             return;
-        } else {
-            Runnable runnable = readRunnable.get();
-            if (runnable != null) {
-                throw new IllegalStateException();
-            }
         }
 
         AsynchronousSocketChannel socket = getSocket();
--- a/sshd-core/src/main/java/org/apache/sshd/common/io/nio2/Nio2Session.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/io/nio2/Nio2Session.java
@@ -31,7 +31,6 @@ import java.util.Objects;
 import java.util.Queue;
 import java.util.concurrent.LinkedTransferQueue;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 
@@ -71,8 +70,9 @@ public class Nio2Session extends AbstractCloseable implements IoSession {
     private final AtomicLong lastReadCycleStart = new AtomicLong();
     private final AtomicLong writeCyclesCounter = new AtomicLong();
     private final AtomicLong lastWriteCycleStart = new AtomicLong();
-    private final AtomicBoolean suspendRead = new AtomicBoolean(false);
-    private final AtomicReference<Runnable> readRunnable = new AtomicReference<>(null);
+    private final Object suspendLock = new Object();
+    private volatile boolean suspend;
+    private volatile Runnable readRunnable;
 
     public Nio2Session(
                        Nio2Service service, FactoryManager manager, IoHandler handler, AsynchronousSocketChannel socket,
@@ -388,7 +388,9 @@ public class Nio2Session extends AbstractCloseable implements IoSession {
     @Override
     public void suspendRead() {
         log.trace("suspendRead({})", this);
-        if (suspendRead.compareAndSet(false, true)) {
+        boolean prev = suspend;
+        suspend = true;
+        if (!prev) {
             log.debug("suspendRead({}) requesting read suspension", this);
         }
     }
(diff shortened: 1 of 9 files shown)


sshd-core/src/main/java/org/apache/sshd/server/forward/TcpipServerChannel.java
```diff
@@ -217,9 +225,23 @@ public void messageReceived(IoSession session, Readable message) throws Exceptio
...
+                    long total = inFlightDataSize.addAndGet(length);
+                    if (total > maxBufferSize) {
+                        session.suspendRead();
+                    }
+                    IoWriteFuture ioWriteFuture = out.writePacket(buffer);
+                    ioWriteFuture.addListener(new SshFutureListener<IoWriteFuture>() {
+                        @Override
+                        public void operationComplete(IoWriteFuture future) {
+                            long total = inFlightDataSize.addAndGet(-length);
+                            if (total <= maxBufferSize) {
```
fengjiajie, line 240: Maybe this threshold can be configured?
when maxBufferSize = 1MB,  I think ```if (total <= 500KB)``` might be better for performance? to avoid frequent suspensions and resumptions

inFlightDataSize > 1 MB,  suspendRead
inFlightDataSize < 0.5 MB, resumedRead

fengjiajie (commented):
👍

sshd-core/src/main/java/org/apache/sshd/server/forward/TcpipServerChannel.java
gnodet: Right, that makes sense, I'll split the property.

sshd-core/src/main/java/org/apache/sshd/common/io/nio2/Nio2Session.java
```diff
@@ -382,7 +385,36 @@ protected void handleReadCycleFailure(ByteBuffer buffer, Readable bufReader, Thr
...
+        log.debug("suspendRead({})", this);
+        suspendRead.set(true);
+    }
+
+    @Override
+    public void resumeRead() {
+        Runnable runnable = readRunnable.getAndSet(null);
+        if (runnable != null) {
+            log.debug("resumeRead({})", this);
+            suspendRead.set(false);
```
tomaswolf, line 399: Shouldn't this be outside the "if", between lines 396/397?

gnodet: I don't think this will change the behavior, but it will be cleaner.

sshd-core/src/main/java/org/apache/sshd/common/io/nio2/Nio2Session.java
```diff
@@ -70,6 +71,8 @@
     private final AtomicLong lastReadCycleStart = new AtomicLong();
     private final AtomicLong writeCyclesCounter = new AtomicLong();
     private final AtomicLong lastWriteCycleStart = new AtomicLong();
+    private final AtomicBoolean suspendRead = new AtomicBoolean(false);
+    private final AtomicReference<Runnable> readRunnable = new AtomicReference<>(null);
```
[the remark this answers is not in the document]
gnodet: In the current state, it's not needed imho, because this is only used for "dumb" tcp connections, i.e. there's no graceful shutdown of the sessions, the sockets are simply closed for forwarding sessions.  I've just added some javadoc to make it clear.

sshd-core/src/main/java/org/apache/sshd/common/io/nio2/Nio2Session.java
```diff
@@ -382,7 +385,32 @@ protected void handleReadCycleFailure(ByteBuffer buffer, Readable bufReader, Thr
...
+        }
+    }
+
     protected void doReadCycle(ByteBuffer buffer, Nio2CompletionHandler<Integer, Object> completion) {
+        if (suspendRead.get()) {
+            log.debug("doReadCycle({}) suspending reading", this);
+            readRunnable.set(() -> doReadCycle(buffer, completion));
+            return;
+        }
+
```
lgoldstein, line 413: Seems to me that there is a chance of a race condition as follows:

* Thread A calls `suspendRead` and marks `suspendedRead` as _true_
* Thread A is pre-empted by thread B that enters `doReadCycle`
* Thread B sees that `suspendRead` is _true_ and enters the `if` block, but is preempted by thread A before calling`readRunnable.set(...)`
* Thread A calls `resumeRead` - calls `readRunnable.getAndSet(null);` and gets _null_ - so thinks there is nothing to do
* Thread B resumes its execution and calls `.set(....)`

Result: reading is not resumed.

I think the suspend/resume should have some mutual exclusion mechanism to prevent this.

sshd-core/src/main/java/org/apache/sshd/server/forward/TcpipServerChannel.java
```diff
@@ -59,6 +62,9 @@
 import org.apache.sshd.server.channel.AbstractServerChannel;
 import org.apache.sshd.server.forward.TcpForwardingFilter.Type;
 
+import static org.apache.sshd.core.CoreModuleProperties.TCPIP_SERVER_CHANNEL_BUFFER_SIZE_THRESHOLD_HIGH;
+import static org.apache.sshd.core.CoreModuleProperties.TCPIP_SERVER_CHANNEL_BUFFER_SIZE_THRESHOLD_LOW;
+
```
lgoldstein, line 67: Doesn't _checkstyle_ flag static imports as errors ?

sshd-core/src/main/java/org/apache/sshd/common/io/nio2/Nio2Session.java
gnodet: Nice catch, i've pushed a new commit with a synchronized block to avoid that.

lgoldstein: Looks good to me...

gnodet:
@fengjiajie I'm still waiting for a test before I can commit this PR.  Do you think you could come up with a unit test for the issue ?

fengjiajie:
@gnodet Hi, I don't know how to construct this unit test now, so I had to look at how unit tests have been implemented before, then I found `PortForwardingLoadTest.testLocalForwardingPayload` , but it always fails to run on my computer, I'm looking at why it failed.

gnodet:
@fengjiajie it seems this test is failing since [SSHD-1003](https://issues.apache.org/jira/browse/SSHD-1003).

gnodet:
Not really. In fact, the test was working better before the partial revert of SSHD-1003 in https://github.com/apache/mina-sshd/commit/e512f68e2067723ce85c0f8e6a8642f2773aae9c

fengjiajie:
I have tried it,
ut failed, 45f84aab5 [SSHD-1003] Use asynchronous streams when forwarding ports
ut success, 449c2a56c (prev SSHD-1003)  [SSHD-1009] SSHD SCP does not work with WinSCP

ut failed, e512f68 [SSHD-1028] Fix SSH_MSG_DISCONNECT: Too many concurrent connections
ut failed, e040cb4dd0aad8 (prev SSHD-1028) [SSHD-1038] Refactor packages from a module into a cleaner hierarchy

gnodet:
@fengjiajie yes, unfortunately, the test was missing some assumptions.  I've seen it pass when there were errors in the transmission.  I've fixed the test in master. 
However, the result is that I can't make it pass since https://github.com/apache/mina-sshd/commit/45f84aab59b2e11d72942cffe9d810e37ab64959 so I'm thinking about somehow reverting this commit and enabling those tests by default in a first pass.  It seems the problem comes from the {{ChannelAsyncOutputStream}}, but I haven't been able to find out where.

PR #169, opened 2020-09-17
[SSHD-1080] Rework the PacketWriter to split according to the various…

gnodet:
… semantics

This is the basis for discussing possible changes


sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
```diff
@@ -254,6 +253,46 @@ protected AbstractSession(
         }
     }
 
+    /**
+     * @param  len       The packet payload size
+     * @param  blockSize The cipher block size
+     * @param  etmMode   Whether using &quot;encrypt-then-MAC&quot; mode
+     * @return           The required padding length
+     */
+    protected static int calculatePadLength(int len, int blockSize, boolean etmMode) {
```
lgoldstein, line 262: Let's make it public - even though used only by sessions - perhaps someone else (not our code) uses it.

sshd-core/src/main/java/org/apache/sshd/common/channel/throttle/DefaultChannelStreamWriter.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.io.IOException;
+
+import org.apache.sshd.common.channel.Channel;
+import org.apache.sshd.common.io.IoWriteFuture;
+import org.apache.sshd.common.util.buffer.Buffer;
+
+public class DefaultChannelStreamWriter implements ChannelStreamWriter {
+
+    private final Channel channel;
```
lgoldstein, line 29: Let's make it `protected` in case someone wants to use it as a base class for something

sshd-common/src/main/java/org/apache/sshd/common/io/IoSession.java
```diff
@@ -82,6 +84,27 @@
...
+     * Closes this session immediately or after all queued write requests are flushed. This operation is asynchronous.
+     * Wait for the returned {@link CloseFuture} if you want to wait for the session actually closed.
+     *
+     * @param  immediately {@code true} to close this session immediately. The pending write requests will simply be
+     *                     discarded. {@code false} to close this session after all queued write requests are flushed.
+     * @return             The generated {@link CloseFuture}
+     */
+    @Override
+    CloseFuture close(boolean immediately);
+
```
lgoldstein, line 107: Why is this override required ? Isn't the method inherited from `Closeable` enough?

sshd-core/src/main/java/org/apache/sshd/common/channel/throttle/DefaultChannelStreamWriter.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+package org.apache.sshd.common.channel.throttle;
+
+import java.io.IOException;
+
+import org.apache.sshd.common.channel.Channel;
+import org.apache.sshd.common.io.IoWriteFuture;
+import org.apache.sshd.common.util.buffer.Buffer;
+
+public class DefaultChannelStreamWriter implements ChannelStreamWriter {
```
lgoldstein, line 27: Missing standard javadoc comment - I think _checkstyle_ will complain

gnodet: The javadoc is a bit more specific to the session, so I think it's worth overriding the method.

sshd-common/src/main/java/org/apache/sshd/common/io/IoSession.java
gnodet: The javadoc is a bit more specific to the session, so I think it's worth overriding the method.

sshd-core/src/main/java/org/apache/sshd/common/channel/throttle/DefaultChannelStreamWriter.java
gnodet: Yes, certainly, I'm not there yet...

lgoldstein: Sorry - I thought you were publishing it and requesting comments...

gnodet: No, that's fine. I haven't checked the details yet.  I do look for comments.

sshd-common/src/main/java/org/apache/sshd/common/io/IoSession.java
lgoldstein: Good enough...

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
gnodet: Done

sshd-core/src/main/java/org/apache/sshd/common/channel/throttle/DefaultChannelStreamWriter.java
gnodet: Done

sshd-core/src/main/java/org/apache/sshd/common/channel/throttle/DefaultChannelStreamWriter.java
gnodet: Added some javadoc

PR #173, opened 2020-10-12
Fix incrementing the invocation_counter part of the IV.

norrisjeremy:
Since the invocation_counter part of the IV is treated as a uint64,
guarding against overflow by using Math.addExact() would incorrectly
throw an ArithmeticException when it hits Long.MAX_VALUE.

### DIFF
--- a/sshd-common/src/main/java/org/apache/sshd/common/cipher/BaseGCMCipher.java
+++ b/sshd-common/src/main/java/org/apache/sshd/common/cipher/BaseGCMCipher.java
@@ -78,6 +78,7 @@ public class BaseGCMCipher extends BaseCipher {
      */
     protected static class CounterGCMParameterSpec extends GCMParameterSpec {
         protected final byte[] iv;
+        protected final long initialCounter;
 
         protected CounterGCMParameterSpec(int tLen, byte[] src) {
             super(tLen, src);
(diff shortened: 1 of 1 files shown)


sshd-common/src/main/java/org/apache/sshd/common/cipher/BaseGCMCipher.java
```diff
@@ -78,19 +78,25 @@ public void update(byte[] input, int inputOffset, int inputLen) throws Exception
...
             iv = src.clone();
+            initialCounter = BufferUtils.getLong(iv, iv.length - Long.BYTES, Long.BYTES);
         }
 
         protected void incrementCounter() {
             int off = iv.length - Long.BYTES;
             long counter = BufferUtils.getLong(iv, off, Long.BYTES);
-            BufferUtils.putLong(Math.addExact(counter, 1L), iv, off, Long.BYTES);
+            long newCounter = counter + 1L;
+            if (newCounter == initialCounter) {
```
jvz, line 96: There doesn't seem to be any harm in adding this check, but the way that SSH works would force a rekey event well before this counter would ever loop back to where it started (rekey events happen far more often than every 2^64 packets).

jvz (approved):

sshd-common/src/main/java/org/apache/sshd/common/cipher/BaseGCMCipher.java
norrisjeremy: Yes, I agree, but I figured it would be better to be safer than sorry incase there was some sort of misuse or issue in the future, since IV reuse could be disastrous.

lgoldstein:
Looks good - will merge soon.

lgoldstein:
Thanks for the contribution - merged - you can close the PR

jvz:
What's the merge commit here?

lgoldstein:
[bec3884c2a41b999106bbd2842359422ff15154e](https://github.com/apache/mina-sshd/commit/bec3884c2a41b999106bbd2842359422ff15154e)

jvz:
@norrisjeremy feel free to close this as it's been merged^

PR #176, opened 2020-10-24
[SSHD-1017] Add support for chacha20-poly1305@openssh.com

jvz:
This uses BouncyCastle due to OpenSSH using the non-RFC variant of ChaCha20 as well as having a direct implementation of Poly1305 (Java 11 does not expose the Poly1305 mac engine).

https://issues.apache.org/jira/browse/SSHD-1017

Some ideas on simplification adapted from https://github.com/mwiede/jsch/pull/17

### DIFF
--- a/sshd-common/src/main/java/org/apache/sshd/common/cipher/BuiltinCiphers.java
+++ b/sshd-common/src/main/java/org/apache/sshd/common/cipher/BuiltinCiphers.java
@@ -102,6 +102,12 @@ public enum BuiltinCiphers implements CipherFactory {
      */
     @Deprecated
     blowfishcbc(Constants.BLOWFISH_CBC, 8, 0, 16, "Blowfish", 128, "Blowfish/CBC/NoPadding", 8),
+    cc20p1305_openssh(Constants.CC20P1305_OPENSSH, 8, 16, 64, "ChaCha", 256, "ChaCha", 8) {
+        @Override
+        public Cipher create() {
+            return new ChaCha20Cipher();
+        }
+    },
     /**
      * @deprecated
      * @see        <A HREF="https://issues.apache.org/jira/browse/SSHD-1004">SSHD-1004</A>
@@ -371,6 +377,7 @@ public enum BuiltinCiphers implements CipherFactory {
         public static final String ARCFOUR128 = "arcfour128";
         public static final String ARCFOUR256 = "arcfour256";
         public static final String BLOWFISH_CBC = "blowfish-cbc";
+        public static final String CC20P1305_OPENSSH = "chacha20-poly1305@openssh.com";
         public static final String TRIPLE_DES_CBC = "3des-cbc";
 
         private Constants() {
--- a/sshd-common/src/main/java/org/apache/sshd/common/cipher/ChaCha20Cipher.java
+++ b/sshd-common/src/main/java/org/apache/sshd/common/cipher/ChaCha20Cipher.java
@@ -1,0 +1,279 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.sshd.common.cipher;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+
+import javax.crypto.AEADBadTagException;
+
+import org.apache.sshd.common.mac.Mac;
+import org.apache.sshd.common.mac.Poly1305Mac;
+import org.apache.sshd.common.util.NumberUtils;
+import org.apache.sshd.common.util.ValidateUtils;
+import org.apache.sshd.common.util.buffer.BufferUtils;
+
+/**
+ * AEAD cipher based on the
+ * <a href="https://github.com/openbsd/src/blob/master/usr.bin/ssh/PROTOCOL.chacha20poly1305">OpenSSH
+ * ChaCha20-Poly1305</a> cipher extension.
+ */
+public class ChaCha20Cipher implements Cipher {
+    protected final ChaChaEngine headerEngine = new ChaChaEngine();
+    protected final ChaChaEngine bodyEngine = new ChaChaEngine();
+    protected final Mac mac = new Poly1305Mac();
+    protected Mode mode;
+
+    public ChaCha20Cipher() {
+        // empty
+    }
+
+    @Override
+    public String getAlgorithm() {
+        return "ChaCha20";
+    }
+
+    @Override
+    public void init(Mode mode, byte[] key, byte[] iv) throws Exception {
+        this.mode = mode;
+
+        bodyEngine.initKey(Arrays.copyOfRange(key, 0, 32));
+        bodyEngine.initNonce(iv);
+        mac.init(bodyEngine.polyKey());
+
+        headerEngine.initKey(Arrays.copyOfRange(key, 32, 64));
+        headerEngine.initNonce(iv);
+        headerEngine.initCounter(0);
+    }
+
+    @Override
+    public void updateAAD(byte[] data, int offset, int length) throws Exception {
+        ValidateUtils.checkState(mode != null, "Cipher not initialized");
+        ValidateUtils.checkTrue(length == 4, "AAD only supported for encrypted packet length");
+
+        if (mode == Mode.Decrypt) {
+            mac.update(data, offset, length);
+        }
+
+        headerEngine.crypt(data, offset, length, data, offset);
+
+        if (mode == Mode.Encrypt) {
+            mac.update(data, offset, length);
+        }
+    }
+
+    @Override
+    public void update(byte[] input, int inputOffset, int inputLen) throws Exception {
+        ValidateUtils.checkState(mode != null, "Cipher not initialized");
+
+        if (mode == Mode.Decrypt) {
+            mac.update(input, inputOffset, inputLen);
+            byte[] actual = mac.doFinal();
+            if (!Mac.equals(input, inputOffset + inputLen, actual, 0, actual.length)) {
+                throw new AEADBadTagException("Tag mismatch");
+            }
+        }
+
+        bodyEngine.crypt(input, inputOffset, inputLen, input, inputOffset);
+
+        if (mode == Mode.Encrypt) {
+            mac.update(input, inputOffset, inputLen);
+            mac.doFinal(input, inputOffset + inputLen);
+        }
+
+        headerEngine.advanceNonce();
+        headerEngine.initCounter(0);
+        bodyEngine.advanceNonce();
+        mac.init(bodyEngine.polyKey());
+    }
+
+    @Override
+    public String getTransformation() {
+        return "ChaCha20";
+    }
+
+    @Override
+    public int getIVSize() {
+        return 8;
+    }
+
+    @Override
+    public int getAuthenticationTagSize() {
+        return 16;
+    }
+
+    @Override
+    public int getCipherBlockSize() {
+        return 8;
+    }
+
+    @Override
+    public int getKdfSize() {
+        return 64;
+    }
+
+    @Override
+    public int getKeySize() {
+        return 256;
+    }
+
+    protected static class ChaChaEngine {
+        private static final int BLOCK_BYTES = 64;
+        private static final int BLOCK_INTS = BLOCK_BYTES / Integer.BYTES;
+        private static final int KEY_OFFSET = 4;
+        private static final int KEY_BYTES = 32;
+        private static final int KEY_INTS = KEY_BYTES / Integer.BYTES;
+        private static final int COUNTER_OFFSET = 12;
+        private static final int NONCE_OFFSET = 14;
+        private static final int NONCE_BYTES = 8;
+        private static final int NONCE_INTS = NONCE_BYTES / Integer.BYTES;
+        private static final int[] ENGINE_STATE_HEADER
+                = unpackSigmaString("expand 32-byte k".getBytes(StandardCharsets.US_ASCII));
+
+        protected final int[] x = new int[BLOCK_INTS];
+        protected final int[] engineState = new int[BLOCK_INTS];
+        protected final byte[] nonce = new byte[NONCE_BYTES];
+        protected long initialNonce;
+
+        protected ChaChaEngine() {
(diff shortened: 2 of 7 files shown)


sshd-common/src/main/java/org/apache/sshd/common/cipher/BuiltinCiphers.java
```diff
@@ -102,6 +104,31 @@ public Cipher create() {
...
+                            SecurityUtils.getCipher("ChaCha");
+                            supported = true;
+                        } catch (GeneralSecurityException ignored) {
+                            supported = false;
+                        }
+                    }
+                }
+            }
+            return supported;
+        }
```
lgoldstein, line 125: If this works only with BC then you can replace/add `SecurityUtils.isBouncyCastleRegistered()` call here

sshd-common/src/main/java/org/apache/sshd/common/cipher/ChaChaCipher.java
```diff
@@ -0,0 +1,142 @@
...
+    protected final Mac poly1305 = new BaseMac("Poly1305", 16, 32, true);
+    protected final byte[] macKey = new byte[32];
+
+    public ChaChaCipher() {
+        super(8, 16, 64, "ChaCha", 256, "ChaCha", 8);
+    }
+
+    @Override
+    protected Cipher createCipherInstance(Mode mode, byte[] key, byte[] iv) throws Exception {
+        this.mode = mode;
```
lgoldstein, line 58: Recommend adding `this.mode = Objects.requireNotNull(mode, "No cipher mode provided
);`

sshd-common/src/test/java/org/apache/sshd/common/cipher/ChaChaCipherTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.apache.sshd.common.util.buffer.BufferUtils;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class ChaChaCipherTest {
+    public ChaChaCipherTest() {
+        // empty
+    }
```
lgoldstein, line 32: Please use `extends JUnitTestSupport` and remove `import static`

sshd-common/src/test/java/org/apache/sshd/common/cipher/ChaChaCipherTest.java
```diff
@@ -0,0 +1,60 @@
...
+import java.nio.charset.StandardCharsets;
+
+import org.apache.sshd.common.util.buffer.BufferUtils;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class ChaChaCipherTest {
+    public ChaChaCipherTest() {
+        // empty
```
lgoldstein, line 31: Please call `super()` explicitly

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
```diff
@@ -1757,14 +1764,18 @@ protected void receiveNewKeys() throws Exception {
...
+
+    protected void resetCounters() {
         inBytesCount.set(0L);
         outBytesCount.set(0L);
         inPacketsCount.set(0L);
         outPacketsCount.set(0L);
         inBlocksCount.set(0L);
         outBlocksCount.set(0L);
-        lastKeyTimeValue.set(Instant.now());
-        firstKexPacketFollows = null;
```
lgoldstein, line 1767: Why were these deleted ? They are used in the code...

lgoldstein (changes requested):

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
jvz: Not deleted, just moved a little bit. This method keeps hitting the 100 line limit from checkstyle, so I had to refactor it a little.

sshd-common/src/main/java/org/apache/sshd/common/cipher/BuiltinCiphers.java
jvz: I just came across that method today as well. This could technically be supported by just copying the source code for ChaCha20 and Poly1305 since I had to combine them into the OpenSSH-specific AEAD form.

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
lgoldstein: If it becomes too cumbersome you can use `@SuppressWarnings("checkstyle:...the rule")` - I do not encourage it, but sometimes it is better than the alternative

jvz: It's not too cumbersome, though this particular method is justifiably being flagged by checkstyle. It's exactly 100 lines, and every SSH extension that doesn't follow the normal key exchange process (e.g., this chacha extension, and to some extent even the aes/gcm one) complicates this code further. Ideally, some form of strategy APIs could be devised around these areas, but it's possible that _what_ strategies are useful won't be apparent until more extensions are supported or created.

As for one potential related simplification to do in this method might be to extract some sort of data class to handle the c2s and s2c cipher/mac/IV/etc. data. Of course, that can always be done later.

sshd-common/src/main/java/org/apache/sshd/common/cipher/ChaChaCipher.java
jvz: To match the superclass behavior, this should actually be changed to default to `Mode.Decrypt` when `mode` is `null`.

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
jvz: I've added a checkstyle suppression for now. My ideas on how this class could be refactored are a bit larger and orthogonal to this PR, so I added a TODO comment. My refactoring idea would be essentially to encapsulate the incoming and outgoing fields into a pair of instances of a sort of session data class.

gnodet:
@jvz What's the status of this PR ? We're planning a 2.7.0 release this week, so wondering if I should merge this PR in.

sshd-core/src/main/java/org/apache/sshd/common/BaseBuilder.java
```diff
@@ -70,6 +70,7 @@
      */
     public static final List<BuiltinCiphers> DEFAULT_CIPHERS_PREFERENCE = Collections.unmodifiableList(
             Arrays.asList(
+                    // BuiltinCiphers.cc20p1305_openssh, // TODO: enable by default when BouncyCastle available
```
tomaswolf, line 73: Why this? The AES256 ciphers also may not be available, but are unconditionally in this list.

jvz:
I'll need to revisit this PR. I've done a lot more cryptography work since then, and I probably have some simplifications now. 😁

sshd-core/src/main/java/org/apache/sshd/common/BaseBuilder.java
jvz: I'll probably just copy in an implementation of ChaCha here for a fallback when BC isn't available.

tomaswolf: I may be missing something here... but this BuiltinCiphers code already can have ciphers in there that have isSupported() == false. So I wonder why we'd need to have a ChaCha implementation in sshd itself. Either some provider provides it, or not.

I would expect the ChaCha cipher never to be proposed or used if it isn't supported. Having it supported only if BC is available would be fine by me. (We seem to require BC >= 1.68. If there's any such BC version that doesn't provide ChaCha, then the feature test would need to be more precise.)

Don't know what the stance of the other maintainers is on having more crypto code inside sshd. AFAIK the only crypto code in sshd is the OpenBSD bcrypt implementation needed to handle OpenSSH encrypted keys. That one was copied because it was felt that having another dependency on an external artifact just for one class was to be avoided, and the inclusion was cleared beforehand here at Apache _and_ with the Eclipse legal team. (The latter only because I wanted it to be used in Eclipse; including it here if then Eclipse legal had objected would have only complicated matters.)

jvz: The copy of ChaCha I'm thinking of is one I wrote in the first place based on the public domain reference implementation. However, if there's a clean way to simply require BouncyCastle to enable this cipher, then we can avoid any additional cipher code being copied. As I mentioned before, the main problem here is that OpenSSH decided to use a pre-standardized version of ChaCha20-Poly1305 which works slightly differently than the one bundled with Java 11+.

tomaswolf: If I understand you right, BC provides this pre-standard ChaCha cipher. But SecurityUtils gives no clean way to ask for a "ChaCha" cipher from a particular provider ("BC"). So that's what I was missing. BuiltinCiphers.cc20p1305_openssh does check whether BC is registered, but in ChaChaCipher when calling SecurityUtils.getCipher() there's no guarantee you get the one from BC. Bypass SecurityUtils and ask via Cipher.getInstance("ChaCha", "BC") directly? Or indeed include your own implementation.

tomaswolf:
> [...] cleared it with the relevant legal teams (which we should do to **every** piece of code we import...)

At least in the case of ChaCha, there is indeed an undeniably public domain base in the OpenSSH sources[1] (original at [2]), and Matt is an Apache member. So this should not be necessary; including Matt's own code based on a public domain implementation is straight-forward; much simpler than including someone else's ISC-licensed code as was the case with bcrypt.

[1] https://github.com/openssh/openssh-portable/blob/master/chacha.c
[2] https://cr.yp.to/chacha.html ("merged" version)

lgoldstein:
>> At least in the case of ChaCha, there is indeed an undeniably public domain base in the OpenSSH sources[1] (original at [2]), and Matt is an Apache member. So this should not be necessary; including Matt's own code based on a public domain implementation is straight-forward; much simpler than including someone else's ISC-licensed code as was the case with bcrypt.

Like I said, I am not opposed to it, and if it is as you describe then even better - let's just make sure with the legal team that there is no issue here - from what you describe it should be pretty straightforward.

jvz:
Let me clarify the IP provenance then. I have two versions of ChaCha20 and Poly1305 in my O(1) Cryptography library. There's a Java port that I wrote based on DJB's papers about ChaCha and Salsa (which are themselves public domain and patent free) and the corresponding RFC that standardized it. There's also a C version that is copied from one of the linked implementations from https://cr.yp.to/chacha.html which is independent (and added much later). O1C is my library where I'm the only copyright holder other than some C code that's imported from elsewhere (the Java code is all pure Java, all done by reimplementing or porting public domain algorithms from other languages).

Suffice to say, I've done due diligence in the library, and I can trivially relicense any parts right now since nobody else has contributed yet.

sshd-core/src/main/java/org/apache/sshd/common/BaseBuilder.java
jvz: Yeah, that was the problem. I'm adding in an implementation for this to simplify things.

jvz:
I've ported in the code I mentioned earlier. Based on files from my other project (ISC licensed, but with my ICLA on file, this is dual-licensed here as Apache 2.0):

* https://github.com/o1c-dev/o1c/blob/ad51541705d117f10cfb71907a7df99028ea9925/src/main/java/dev/o1c/impl/chacha20/ChaCha20.java
* https://github.com/o1c-dev/o1c/blob/ad51541705d117f10cfb71907a7df99028ea9925/src/main/java/dev/o1c/impl/chacha20/Poly1305.java
* https://github.com/o1c-dev/o1c/blob/ad51541705d117f10cfb71907a7df99028ea9925/src/main/java/dev/o1c/util/ByteOps.java (some methods copied from here, but they're all trivial)

I tested this out interactively with OpenSSH_8.1p1, LibreSSL 2.7.3 on macOS 11.3.1.

jvz:
Also, I tested this with Java 15 (openjdk version "15.0.2" 2021-01-19) which was the latest JDK in Homebrew.

sshd-common/src/main/java/org/apache/sshd/common/cipher/ChaCha20Cipher.java
```diff
@@ -0,0 +1,279 @@
...
+                int lo = ++engineState[COUNTER_OFFSET];
+                if (lo == 0) {
+                    // overflow
+                    ++engineState[COUNTER_OFFSET + 1];
+                }
+            }
+        }
+
+        protected byte[] polyKey() {
+            byte[] block = new byte[32];
```
tomaswolf, line 206: Can't we just `return block;` below?

sshd-common/src/main/java/org/apache/sshd/common/mac/Poly1305Mac.java
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    @Override
+    public String getAlgorithm() {
+        return "Poly1305";
+    }
+
+    @Override
+    public void init(byte[] key) throws Exception {
+        if (NumberUtils.length(key) != 32) {
```
tomaswolf, line 71: Maybe add a Poly1305Mac.KEY_BYTES constant for this 32, and use in ChachaEngine.polyKey()?

jvz:
I don't see why the CI build is failing. Works on my machine.

tomaswolf (commented):
Nice. Two little suggestions.

tomaswolf:
> I don't see why the CI build is failing. Works on my machine.

Neither do I. Maybe just an unstable test?

tomaswolf:
LoadTest is unstable for me locally. Didn't dig deep; my suspicion is that in that test the channel gets closed sometimes before all data has been written back to the client.

jvz:
That one hasn't failed for me locally, though I've seen the SFTP transfer test randomly hang once in a while.

tomaswolf:
`SftpTransferTest` hanging is now issue [SSHD-1191](https://issues.apache.org/jira/browse/SSHD-1191).

It took a while, but I finally managed to get a clean test run locally.

@jvz , is it OK for you if I squash all this and push onto this PR?

jvz:
Go ahead!

tomaswolf:
CI build still unstable; problems with that `LoadTest`. Works locally, so in it goes.

Thanks a lot, Matt!

tomaswolf:
@jvz , I had to revert this. The hang in `SftpTransportTest`is caused by a KEX failure that occurs only if this chacha20-poly1305 implementation is used.

A debug log I captured is attached in [SSHD-1191](https://issues.apache.org/jira/browse/SSHD-1191).

Usually KEX fails in rekeying, but the debug log shows a failure in the initial KEX. Something is definitely not right yet.

tomaswolf:
And it's in again.

Chacha20-poly1305 has a block size of 8, and `AbstractSession` computed rekey limits for ciphers with block sizes < 16 wrongly, which caused SSHD-1191. That's fixed now. If there are remaining problems with rekeying, they're in any case not caused by the choice of cipher.

This chacha20-poly1305 implementation itself is perfectly fine.

jvz:
Oh good to hear that because I was never able to figure out why anything wasn't working. Since I started these PRs, I've switched jobs where we use TLS instead of SSH as the secure transport behind most things, so I'm already starting to forget the protocol details from SSH. :)

PR #177, opened 2020-11-01
[SSHD-704] Add support for RFC 8731

jvz:
This adds support for curve25519-sha256 (and its libssh version) along with curve448-sha512. These key exchange methods are supported for both Java 11 and Bouncy Castle.

https://issues.apache.org/jira/browse/SSHD-704
https://tools.ietf.org/html/rfc8731
https://tools.ietf.org/html/rfc7748 (for the X25519/X448 algorithms)

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/common/kex/MontgomeryCurve.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/kex/MontgomeryCurve.java
@@ -1,0 +1,158 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.sshd.common.kex;
+
+import java.io.IOException;
+import java.io.StreamCorruptedException;
+import java.io.UncheckedIOException;
+import java.security.GeneralSecurityException;
+import java.security.InvalidKeyException;
+import java.security.KeyFactory;
+import java.security.KeyPair;
+import java.security.KeyPairGenerator;
+import java.security.PublicKey;
+import java.security.spec.InvalidKeySpecException;
+import java.security.spec.X509EncodedKeySpec;
+
+import javax.crypto.KeyAgreement;
+
+import org.apache.sshd.common.OptionalFeature;
+import org.apache.sshd.common.digest.BuiltinDigests;
+import org.apache.sshd.common.digest.Digest;
+import org.apache.sshd.common.digest.DigestFactory;
+import org.apache.sshd.common.keyprovider.KeySizeIndicator;
+import org.apache.sshd.common.util.io.der.ASN1Object;
+import org.apache.sshd.common.util.io.der.DERParser;
+import org.apache.sshd.common.util.security.SecurityUtils;
+
+/**
+ * Provides implementation details for Montgomery curves and their key exchange algorithms Curve25519/X25519 and
+ * Curve448/X448 specified in RFC 7748 and RFC 8731. Montgomery curves provide improved security and flexibility over
+ * Weierstrass curves used in ECDH.
+ *
+ * @see <a href="https://www.rfc-editor.org/info/rfc7748">RFC 7748</a>
+ * @see <a href="https://www.rfc-editor.org/info/rfc8731">RFC 8731</a>
+ */
+public enum MontgomeryCurve implements KeySizeIndicator, OptionalFeature {
+
+    /**
+     * X25519 uses Curve25519 and SHA-256 with a 32-byte key size.
+     */
+    x25519("X25519", 32, BuiltinDigests.sha256,
+           new byte[] { 0x30, 0x2a, 0x30, 0x05, 0x06, 0x03, 0x2b, 0x65, 0x6e, 0x03, 0x21, 0x00 }),
+    /**
+     * X448 uses Curve448 and SHA-512 with a 56-byte key size.
+     */
+    x448("X448", 56, BuiltinDigests.sha512,
+         new byte[] { 0x30, 0x42, 0x30, 0x05, 0x06, 0x03, 0x2b, 0x65, 0x6f, 0x03, 0x39, 0x00 });
+
+    private final String algorithm;
+    private final int keySize;
+    private final boolean supported;
+    private final DigestFactory digestFactory;
+    private final KeyPairGenerator keyPairGenerator;
+    private final KeyFactory keyFactory;
+    private final byte[] encodedPublicKeyPrefix;
+    private final int encodedKeySize;
+
+    MontgomeryCurve(String algorithm, int keySize, DigestFactory digestFactory, byte[] encodedPublicKeyPrefix) {
+        this.algorithm = algorithm;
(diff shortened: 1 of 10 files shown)


sshd-core/src/main/java/org/apache/sshd/common/kex/MontgomeryCurve.java
```diff
@@ -0,0 +1,158 @@
...
+        System.arraycopy(key, 0, encoded, encodedKeySize - key.length, key.length);
+        return keyFactory.generatePublic(new X509EncodedKeySpec(encoded));
+    }
+
+    private static byte[] extractSubjectPublicKey(byte[] subjectPublicKeyInfo) throws InvalidKeyException {
+        try {
+            //  SubjectPublicKeyInfo ::= SEQUENCE {
+            //   algorithm AlgorithmIdentifier,
+            //   subjectPublicKey BIT STRING }
+            ASN1Object spki = new DERParser(subjectPublicKeyInfo).readObject();
```
jvz, line 147: I found some easier ways to encode/decode keys in a separate project I'm working on: https://github.com/o1c-dev/o1c/blob/2e41f8e6268ead61562474424a01f256a7afe8e1/java8/src/main/java/dev/o1c/bc/BouncyCastleKeyPairCodec.java#L75

sshd-core/src/main/java/org/apache/sshd/common/kex/MontgomeryCurve.java
```diff
@@ -0,0 +1,158 @@
...
+
+    public byte[] encode(PublicKey key) throws InvalidKeyException {
+        return extractSubjectPublicKey(key.getEncoded());
+    }
+
+    public PublicKey decode(byte[] key) throws InvalidKeySpecException {
+        if (key.length < getKeySize()) {
+            throw new InvalidKeySpecException("Provided key is too small for " + getAlgorithm());
+        }
+        // ideally, we'd just parse the key as a BigInteger and then create a XECPublicKeySpec in Java 11
```
jvz, line 131: Note: turns out the `XECPublicKeySpec` takes a `BigInteger` parameter for `u`, but `u` when in binary is little endian. Thus, if using the spec class, the array must be reversed into BigInteger, then the XDH engine will reverse it again back to little endian for use in the algorithm. Same applies to the EdDSA APIs added in Java 11. Thus, the DER dumps of ASN.1 data seem to be both portable as well as less redundant as the key is already little endian in the DER form.

tomaswolf:
A test would be nice.

sshd-core/src/main/java/org/apache/sshd/common/kex/MontgomeryCurve.java
```diff
@@ -0,0 +1,158 @@
...
+ * @see <a href="https://www.rfc-editor.org/info/rfc7748">RFC 7748</a>
+ * @see <a href="https://www.rfc-editor.org/info/rfc8731">RFC 8731</a>
+ */
+public enum MontgomeryCurve implements KeySizeIndicator, OptionalFeature {
+
+    /**
+     * X25519 uses Curve25519 and SHA-256 with a 32-byte key size.
+     */
+    x25519("X25519", 32, BuiltinDigests.sha256,
+           new byte[] { 0x30, 0x2a, 0x30, 0x05, 0x06, 0x03, 0x2b, 0x65, 0x6e, 0x03, 0x21, 0x00 }),
```
jvz, line 59: While I originally considered this idea to be a bit hacky, I've since seen it in more than one other XDH implementation for Java. The X.509 encoded form of an X25519 or X448 key are always going to have the same prefix data (the algorithm id and key length in some ASN.1 format). Thus, we can portably use the `X509EncodedKeySpec` class to serialize and deserialize XDH public keys without needing a full ASN.1/DER encoding library like in BouncyCastle.

sshd-core/src/main/java/org/apache/sshd/common/kex/MontgomeryCurve.java
jvz: - [ ] Just extract the key data directly from the byte array as it's the last `keySize` bytes in the X.509 encoded form

tomaswolf:
Came back to this. It's not even activated; one needs to add the new DH factories in BaseBuilder first. If one does so, curve25519 gets chosen for KEX (tested against the Github SSH daemon and aginst the Gitlab SSH daemon). KEX fails; the connection is terminated immediately by the server.

Something is missing somewhere, or somewhere along the line we don't do what the RFCs say we should do.

tomaswolf:
It works if both keys (our and theirs) have the most significant bit zero.

If our key (encode()) has the most significant bit set, they close the connection.

If their key (decode()) has the most significant bit set, we fail to verify their signature.

I suspect this is related to [RFC 7748, section 5](https://datatracker.ietf.org/doc/html/rfc7748#section-5):

> When receiving such an array, implementations of X25519 (but not X448) MUST mask the most significant bit in the final byte.

(They talk little-endian here.)

tomaswolf:
@jvz any ideas?

tomaswolf:
> If our key (encode()) has the most significant bit set, they close the connection.
> 
> If their key (decode()) has the most significant bit set, we fail to verify their signature.

Got it. This is caused by a subtle difference in the DH key exchange.

* [RFC 4253](https://datatracker.ietf.org/doc/html/rfc4253#page-22): e and f are written as mpint (so prefixed by a zero byte if the high bit is set).
* [RFC 5656](https://datatracker.ietf.org/doc/html/rfc5656#page-9): Q_C and Q_S, which take the place of e and f, are written as octet strings, hence _not_ prefixed with a zero byte, even if the high bit is set.

That explains it: Apache MINA sshd implements RFC 4253, but not the modifications for ECDH/XDH from RFC 5656.

* when our key has the high bit set, we send 33 bytes, and the server disconnects because that's invalid.
* when their key (or our key) has the high bit set, we add extra zero bytes to the data the signature hash is computed over, and thus signature verification always fails.
* if neither key has the high bit set, all is fine.

The difference is pointed out in [RFC 8731](https://datatracker.ietf.org/doc/html/rfc8731#section-3):
> The key exchange procedure is similar to the ECDH method described in Section 4 of [RFC5656], though with a different wire encoding used for public values and the final shared secret.  Public ephemeral keys are encoded for transmission as standard SSH strings.

RFC 8731, section 3.1, then explains that the shared secret (32 bytes for x25519) is always interpreted as unsigned. (But according to RFC 5656, it's still written as mpint.) Our code writes it as mpint, which prefixes a zero byte if the high bit is set, so that should be fine.

With this fixed, KEX using curve25519 works against both the Github and Gitlab daemons.

@jvz, is it OK if I push a rebase + follow-up change onto this PR?

jvz:
Yes, that sounds like a great analysis and fix idea! I'll be on vacation for the next week, though I can review PRs from my phone.

tomaswolf:
Hm, that doesn't seem to work in the CI build. At least with Java 14. I'll have to investigate locally.

sshd-core/src/main/java/org/apache/sshd/common/kex/MontgomeryCurve.java
tomaswolf: Done.

sshd-core/src/main/java/org/apache/sshd/common/kex/MontgomeryCurve.java
tomaswolf: Added comments in the source explaining what these magic constants are. (Beginning of a X.509 DER encoding for x25519 or x448 public key SubjectPublicKeyInfo.) We might know it now, but someone else looking at this a few years down the road might not.

tomaswolf:
CI builds finally passed; so the failures on Windows in the two earlier CI runs were just unstable tests. Couldn't reproduce them locally either.

BTW: this problem with mpint vs. bit strings wasn't noticed earlier because ECDH keys always start with 0x04 (uncompressed) or 0x03 or 0x02 (compressed). But x25519 or x448 keys may have the high bit in the first byte set.

This is ready for review.

jvz (commented):
Changes look good to me! That definitely explains the random key length issues I recall having in this PR initially.

tomaswolf:
Thanks for taking a look. Yes, I think the "extra zero byte" handling could be removed now altogether. I decided to leave it in just in case. It doesn't hurt, and it's a minor and harmless deviation from the RFCs (which say to disconnect if you don't get exactly 32 or 56 bytes).

PR #178, opened 2020-11-26
Update client-setup.md

idontusenumbers:
Fixed case in method call to setUpDefaultClient in example

### DIFF
--- a/docs/client-setup.md
+++ b/docs/client-setup.md
@@ -119,7 +119,7 @@ Furthermore, one can change almost any configured `SshClient` parameter - althou
 sessions depends on the actual changed configuration. Here is how a typical usage would look like
 
 ```java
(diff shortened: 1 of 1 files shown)


lgoldstein (commented):
While I understand the sentiment, the method is so widely used that I don't see the value of introducing this incompatibility.

idontusenumbers:
The case in the example does not currently match the actual java code. This fixes the example so that it compiles.

lgoldstein:
Sorry - my mistake - for some reason I misunderstood and thought you were suggesting we rename the actual method. I'll be happy to make this correction in our documentation

lgoldstein:
Thanks for bringing this to our attention - I fixed it (there were a few more locations with wrong case). You can close the PR

PR #179, opened 2020-11-27
Fix filename in layout in sftp.md

The-Yoda:


lgoldstein:
Thanks for bringing this to our attention - I have fixed the example though along other lines than your suggested PR. You can close it now.

PR #183, opened 2021-03-19
[SSHD-1105] Try all configured signature algorithms for a public key

tomaswolf:
Some keys (RSA) may have several signature algorithms (rsa-sha2-512,
rsa-sha2-256, ssh-rsa). Try them all in the order defined and try
the next key only if no attempt was successful.

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/client/auth/pubkey/UserAuthPublicKey.java
+++ b/sshd-core/src/main/java/org/apache/sshd/client/auth/pubkey/UserAuthPublicKey.java
@@ -24,9 +24,13 @@ import java.security.KeyPair;
 import java.security.PublicKey;
 import java.security.spec.InvalidKeySpecException;
 import java.util.Collection;
+import java.util.Deque;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
 
 import org.apache.sshd.client.auth.AbstractUserAuth;
 import org.apache.sshd.client.auth.keyboard.UserInteraction;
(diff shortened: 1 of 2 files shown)


tomaswolf:
@lgoldstein : could you review this, please?

Is the client-side handling of SSH2_MSG_EXT_INFO server-sig-algs existing already somewhere? If so, we might not have to iterate through all algorithms configured client-side but could just use the first match common between server and client if the server sent that info.

sshd-core/src/main/java/org/apache/sshd/client/auth/pubkey/UserAuthPublicKey.java
```diff
@@ -56,6 +58,7 @@
     protected Iterator<PublicKeyIdentity> keys;
     protected PublicKeyIdentity current;
     protected List<NamedFactory<Signature>> factories;
+    protected List<String> currentAlgorithms = new LinkedList<>();
```
lgoldstein, line 61: A `LinkedList` is not really a `List` and any attempt to use it as such for random access has horrible performance (true for large lists but let's not quibble). Since the code only handles the "edges" (head/tail) I re commend defining it as
```java
Deque<String> currentAlgorithms = new LinkedList<>();
```
and the you can use `removeHead/addTail`/etc. And since `Deque extends Collection` you can also use it as such.

lgoldstein (changes requested):
I don't see the comparison between the requested signature and the one returned by the server + the warning log of their mismatch.

lgoldstein:
> Is the client-side handling of SSH2_MSG_EXT_INFO server-sig-algs existing already somewhere?

Not to my knowledge - when we introduced this feature it was left to the user to implement it. There is the `DefaultClientKexExtensionHandler` of course, but I don't think it is complete or even works as expected.

sshd-core/src/main/java/org/apache/sshd/client/auth/pubkey/UserAuthPublicKey.java
tomaswolf: Done.

lgoldstein: Great - what about  the comparison between the requested signature and the one returned by the server + the warning log of their mismatch ?

lgoldstein: Fair enough - in that case I have no further comments

PR #184, opened 2021-03-20
[SSHD-1141] Fix client-side server-sig-algs handling

tomaswolf:
Unconditionally announce that the client wants to get the server's
SSH_MSG_EXT_INFO. Otherwise the server will never send it. Add the
"ext-info-c" marker only on the very first key exchange proposal,
and make sure the client doesn't send by mistake "ext-info-s".

KexExtensions are available in all phases except PREKEX.

When server-sig-algs is received, reorder the client session's
signature factories such that algorithms the server announced as
supported come first, followed by those not announced, both in client
order.

The client determines the order and the server just says what it
supports.

Note that per RFC 8308 [1] it's possible that a server doesn't announce
_all_ the algorithms it supports, and a client is allowed to try
unsupported algorithms, but may face authentication penalties such
as back-off delays, authentication failures, or disconnections.

[1] https://tools.ietf.org/html/rfc8308

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/common/kex/extension/DefaultClientKexExtensionHandler.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/kex/extension/DefaultClientKexExtensionHandler.java
@@ -22,28 +22,19 @@ package org.apache.sshd.common.kex.extension;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
-import java.util.EnumMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.NavigableSet;
-import java.util.Objects;
-import java.util.stream.Stream;
+import java.util.Set;
+import java.util.TreeSet;
 
 import org.apache.sshd.common.AttributeRepository.AttributeKey;
 import org.apache.sshd.common.NamedFactory;
-import org.apache.sshd.common.NamedResource;
-import org.apache.sshd.common.OptionalFeature;
-import org.apache.sshd.common.config.keys.KeyUtils;
 import org.apache.sshd.common.kex.KexProposalOption;
 import org.apache.sshd.common.kex.extension.parser.ServerSignatureAlgorithms;
 import org.apache.sshd.common.session.Session;
-import org.apache.sshd.common.signature.BuiltinSignatures;
 import org.apache.sshd.common.signature.Signature;
-import org.apache.sshd.common.signature.SignatureFactory;
 import org.apache.sshd.common.util.GenericUtils;
-import org.apache.sshd.common.util.MapEntryUtils;
-import org.apache.sshd.common.util.functors.UnaryEquator;
 import org.apache.sshd.common.util.logging.AbstractLoggingBean;
 
 /**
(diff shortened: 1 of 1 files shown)


sshd-core/src/main/java/org/apache/sshd/common/kex/extension/DefaultClientKexExtensionHandler.java
```diff
@@ -52,247 +43,111 @@
...
      */
-    public static final AttributeKey<Map<KexProposalOption, String>> SERVER_PROPOSAL_KEY = new AttributeKey<>();
-
-    public static final NavigableSet<String> DEFAULT_EXTRA_SIGNATURES = Collections.unmodifiableNavigableSet(
-            GenericUtils.asSortedSet(String.CASE_INSENSITIVE_ORDER,
-                    KeyUtils.RSA_SHA256_KEY_TYPE_ALIAS,
-                    KeyUtils.RSA_SHA512_KEY_TYPE_ALIAS));
-
-    public static final DefaultClientKexExtensionHandler INSTANCE = new DefaultClientKexExtensionHandler();
+    private static final AttributeKey<Boolean> CLIENT_PROPOSAL_MADE = new AttributeKey<>();
```
lgoldstein, line 56: I don't see why this should be `private` - it is immutable and someone might find it useful. In general, I try not to "hide" anything from the users (within reason) - even if they can cause harm by misuse - since one never knows how our users might need to extend/our code.

tomaswolf: I know we have different ideas about interfaces. Not the first time this issue crops up :-) If you really think this should be accessible, let's add an accessor method.

If this is public, a client could reset this -- with might break rekeying? According to RFC 8308 the indicator is to be added only on the _first_ key exchange.

lgoldstein: >> I know we have different ideas about interfaces. Not the first time this issue crops up :-)
Indeed :-)

>> If you really think this should be accessible, let's add an accessor method

How would this prevent misuse ?

>> If this is public, a client could reset this -- with might break rekeying? According to RFC 8308 the indicator is to be added only on the first key exchange.

Like I said - I prefer holding the users responsible for misuse rather than patronize them and "protect them from themselves"

tomaswolf: Done, though I'd still have preferred at least a method like `public boolean wasClientExtensionProposalMade(Session session)` instead of a public attribute. We don't have to make it easy for clients to shoot their own feet.

lgoldstein: >> Done, though I'd still have preferred at least a method like public boolean wasClientExtensionProposalMade(Session session) instead of a public attribute. We don't have to make it easy for clients to shoot their own feet.

Like you said - we differ in our approach/view on this matter :-))

lgoldstein: Do you have commit rights to _master_ or do you need me to merge your PRs ? If so, let me know when you think it is ready (since I have no more comments) and I will do so.

tomaswolf: I should have, soon. Finally got around to enable 2FA here on Github, accounts are linked, so now I'm waiting for the organizational invite. If I understood the process correctly, that should then also enable commit rights on master.

lgoldstein: >> I'm waiting for the organizational invite

Note that the _master_ I am speaking about is not the _github_ one but rather the ASF one. Github is only a **mirror** of the ASF one - we never push to github...

tomaswolf: I had asked Guillaume a while ago about where I should commit/merge; he said he just pushed to the Github repo and that worked for him. There's also a number of commits at gitbox that are PR merges. So I gathered that indeed simply merging PRs the usual Github way would work.

lgoldstein: >> I had asked Guillaume a while ago about where I should commit/merge

Please ask him then how to merge the 2 PRs - in any case, if you need me to merge them let me know

PR #185, opened 2021-03-24
[SSHD-1141] Tolerate SSH_MSG_USERAUTH_PK_OK with wrong algorithm

tomaswolf:
SSH_MSG_USERAUTH_PK_OK is supposed to say "the server will accept the
key and signature algorithm the client sent in the previous
SSH_MSG_USERAUTH_REQUEST message". It is _not_ intended as a mechanism
for the server to override the client's choice of algorithm.

See RFC 4252.[1]

Just log a warning if a server sends back an unexpected "algorithm".
(Github's SSH-2.0-babeld-383743ad was found to reply "ssh-rsa" even
if the client requested "rsa-sha2-512" and the server had previously
announced via server-sig-algs that it supports rsa-sha2-512. Subsequent
authentication with rsa-sha2-512 then still works.)

Of course the returned public key must still match exactly.

The client always uses the signature algorithm that it had announced
it would use.

[1] https://tools.ietf.org/html/rfc4252#page-9

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/client/auth/pubkey/UserAuthPublicKey.java
+++ b/sshd-core/src/main/java/org/apache/sshd/client/auth/pubkey/UserAuthPublicKey.java
@@ -23,7 +23,6 @@ import java.io.IOException;
 import java.security.KeyPair;
 import java.security.PublicKey;
 import java.security.spec.InvalidKeySpecException;
-import java.util.Collection;
 import java.util.Deque;
 import java.util.Iterator;
 import java.util.LinkedList;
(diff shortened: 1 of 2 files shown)


lgoldstein (approved):
Looks OK to me

tomaswolf:
All right. Merging via Github works, but none of the three possibilities do what I had expected. So I'll have to "merge" PRs locally and then push directly in the future.

lgoldstein:
Hi @tomaswolf , thanks for the great work. I do have one minor request - before merging into `master` please run a **full** `mvn clean install` - this is required due to the fact that during the build process the sources may be (re-)formatted, which might change them. We need to make sure that if someone pulls the code and compiles it, the `git status` command will not show any changes to our committed files. In this case, there were 2 files that were not properly formatted and they turned out when I was working on some issue and built the code - I was surprised to see 2 extra files that I have not touched marked as modified.

Again, no real harm - but let's try and avoid these potential pitfalls.

PR #186, opened 2021-03-24
[SSHD-1141] Pubkey auth: handle failure with known signature algorithm

tomaswolf:
If the server had announced via server-sig-algs which signature
algorithms it supports, UserAuthPubKey tries the known algorithms
first. If authentication with such a known signature algorithm fails,
there is no point trying other algorithms.

Trying different signature algorithms may make sense only if the client
doesn't know which algorithms the server supports.

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/client/auth/pubkey/UserAuthPublicKey.java
+++ b/sshd-core/src/main/java/org/apache/sshd/client/auth/pubkey/UserAuthPublicKey.java
@@ -38,6 +38,7 @@ import org.apache.sshd.common.NamedFactory;
 import org.apache.sshd.common.RuntimeSshException;
 import org.apache.sshd.common.SshConstants;
 import org.apache.sshd.common.config.keys.KeyUtils;
+import org.apache.sshd.common.kex.extension.DefaultClientKexExtensionHandler;
 import org.apache.sshd.common.signature.Signature;
 import org.apache.sshd.common.signature.SignatureFactoriesHolder;
 import org.apache.sshd.common.signature.SignatureFactoriesManager;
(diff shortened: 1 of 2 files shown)


tomaswolf:
@lgoldstein : a final tweak.

sshd-core/src/main/java/org/apache/sshd/client/auth/pubkey/UserAuthPublicKey.java
```diff
@@ -105,6 +106,19 @@ protected boolean sendAuthDataRequest(ClientSession session, String service) thr
...
+                Set<String> knownServerAlgorithms = session.getAttribute(
+                        DefaultClientKexExtensionHandler.SERVER_ALGORITHMS);
+                if (knownServerAlgorithms != null
+                        && knownServerAlgorithms.contains(chosenAlgorithm)) {
+                    // We've tried key 'current' with 'chosenAlgorithm', but it
+                    // failed. However, the server had told us it supported
+                    // 'chosenAlgorithm'. Thus it makes no sense to continue
+                    // with this key and other signature algorithms. Skip to the
+                    // next key, if any.
+                    currentAlgorithm = null;
```
lgoldstein, line 119: Let's add a DEBUG level log message summarizing this behavior so that user can tell when it occurred.

tomaswolf: Done.

lgoldstein:
Looks fine to me

lgoldstein:
Hi @tomaswolf , thanks for the great work. I do have one minor request - before merging into `master` please run a **full** `mvn clean install` - this is required due to the fact that during the build process the sources may be (re-)formatted, which might change them. We need to make sure that if someone pulls the code and compiles it, the `git status` command will not show any changes to our committed files. In this case, there were 2 files that were not properly formatted and they turned out when I was working on some issue and built the code - I was surprised to see 2 extra files that I have not touched marked as modified.

Again, no real harm - but let's try and avoid these potential pitfalls.

PR #187, opened 2021-03-24
[SSHD-1147] SftpInputStreamAsync: get file size before SSH_FXP_OPEN

tomaswolf:
IBM's Sterling B2B SFTP connector may limit some files to be downloaded
only once. The mechanism they use assumes that once a file has been
opened, it'll be read and closed without any other SFTP requests being
made on the same handle in between.

Get the file size via SSH_FXP_STAT before opening the file instead of
using SSH_FXP_FSTAT on the handle obtained from opening the file. That
way it should be possible to download such files from such an off-spec
SFTP server.


gnodet:
Looks good to me.

PR #192, opened 2021-05-10
[SSHD-525] Server side implementation of posix-rename@openssh.com

gnodet:


tomaswolf:
This was merged as commit 9a724be.

PR #195, opened 2021-05-12
[SSHD-1163] Identify server key type by negotiation kex result parameters

maxTraydakalo:
Wrong server key type algorithm choose (in case of use rsa-sha2-256 and rsa-sha2-512 always identify like ssh-rsa) in case of usage DHFactories with group exchange.

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/client/kex/DHGEXClient.java
+++ b/sshd-core/src/main/java/org/apache/sshd/client/kex/DHGEXClient.java
@@ -27,9 +27,9 @@ import org.apache.sshd.client.session.AbstractClientSession;
 import org.apache.sshd.common.NamedFactory;
 import org.apache.sshd.common.SshConstants;
 import org.apache.sshd.common.SshException;
-import org.apache.sshd.common.config.keys.KeyUtils;
 import org.apache.sshd.common.kex.AbstractDH;
 import org.apache.sshd.common.kex.DHFactory;
+import org.apache.sshd.common.kex.KexProposalOption;
 import org.apache.sshd.common.kex.KeyExchange;
 import org.apache.sshd.common.kex.KeyExchangeFactory;
 import org.apache.sshd.common.session.Session;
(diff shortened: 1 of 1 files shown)


tomaswolf:
Rebase and merged as commit 67eadba1. Thank you!

smilesometimes:
HAHHA, I  find this commit , this is a bug .   

At begining  i was wondering  why  the  method of getting keyAlg    in    org.apache.sshd.client.kex.DHGEXClient#next   and   org.apache.sshd.client.kex.DHGClient#next was differrent? The useage of  DHGEXClient is right.

Because the  method KeyUtils.getKeyType(serverKey);   when serverKey is "rsa-sha2-256" and  "rsa-sha2-512"  only return "ssh-rsa" , this lead to  a  error called  "KeyExchange signature verification failed for key type= ssh-rsa".

I use this way  to bypass、let the "etch" configed before the  "diffie-hellman-group", so the DHGEXClient can be userd for kex.

`
KexAlgorithms = **ecdh-sha2-nistp256,ecdh-sha2-nistp384,ecdh-sha2-nistp521**,diffie-hellman-group-exchange-sha256,diffie-hellman-group14-sha1,diffie-hellman-group1-sha1. `

tomaswolf:
@smilesometimes : I don't understand what you are writing. Are you saying this correction is wrong? Or are you saying it was wrong before, but with this correction it's good?

smilesometimes:
Oh ,    maxTraydakalo's commit is right.  Today  I have  met this bug, my device can not connect. 
`
// client config like this,   cannot connet
KexAlgorithms = diffie-hellman-group-exchange-sha256
 HostKeyAlgorithms = rsa-sha2-256 
`
`
//  client config like this,     can connect
KexAlgorithms = ecdh-sha2-nistp256
 HostKeyAlgorithms = rsa-sha2-256 
`

Susmit07:
@smilesometimes can you have a look to this below issue, 

https://issues.apache.org/jira/browse/SSHD-1163

Will this fix can fix my use case as well?

smilesometimes:
yes,  this can fix your case.   



> @smilesometimes can you have a look at this below issue,
> 
> https://issues.apache.org/jira/browse/SSHD-1163
> 
> Will this fix can fix my use case as well?

PR #196, opened 2021-05-15
[SSHD-1164] - fixed parsing of sshd_config "Host" lines to collapse spaces

alex-sherwin:
The `ssh_config` man page defines the format for `Host` a bit vaguely as:

```
If more than one pattern is provided, they should be separated by whitespace.
```

When running the unit test suite that read my `~/.ssh/config` file, there was a `Host` line which had two spaces between some values which failed parsing in MINA, but doesn't cause any problem for OpenSSH.

This PR simply makes the parsing of `Host` directive values more lenient by removing blank strings from the split values

### DIFF
--- a/sshd-common/src/main/java/org/apache/sshd/client/config/hosts/HostConfigEntry.java
+++ b/sshd-common/src/main/java/org/apache/sshd/client/config/hosts/HostConfigEntry.java
@@ -898,6 +898,8 @@ public class HostConfigEntry extends HostPatternsHolder implements MutableUserHo
             List<String> valsList = parseConfigValue(value);
 
             if (HOST_CONFIG_PROP.equalsIgnoreCase(key)) {
+                // parseConfigValue may result in entries that are empty strings if > 1 space is used between values
+                valsList = GenericUtils.filterToNotBlank(valsList);
                 if (GenericUtils.isEmpty(valsList)) {
                     throw new StreamCorruptedException("Missing host pattern(s) at line " + lineNumber + ": " + line);
                 }
--- a/sshd-common/src/main/java/org/apache/sshd/common/util/GenericUtils.java
+++ b/sshd-common/src/main/java/org/apache/sshd/common/util/GenericUtils.java
@@ -196,6 +196,49 @@ public final class GenericUtils {
         return !isEmpty(cs);
     }
 
+    /**
+     * <p>
+     * Checks if a CharSequence is empty (""), null or whitespace only.
+     * </p>
+     *
+     * <p>
+     * Whitespace is defined by {@link Character#isWhitespace(char)}.
+     * </p>
+     *
+     * <pre>
+     * StringUtils.isBlank(null)      = true
+     * StringUtils.isBlank("")        = true
+     * StringUtils.isBlank(" ")       = true
+     * StringUtils.isBlank("bob")     = false
+     * StringUtils.isBlank("  bob  ") = false
+     * </pre>
+     *
+     * @param  cs the CharSequence to check, may be null
+     * @return    {@code true} if the CharSequence is null, empty or whitespace only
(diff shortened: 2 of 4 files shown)


sshd-common/src/main/java/org/apache/sshd/client/config/hosts/HostConfigEntry.java
```diff
@@ -1087,6 +1087,7 @@ public static void writeHostConfigEntries(
      */
     public static List<String> parseConfigValue(String value) {
         String s = GenericUtils.replaceWhitespaceAndTrim(value);
+        s = GenericUtils.collapseSpaces(s);
```
tomaswolf, line 1090: I could imagine that there may be pathological cases involving quoted values where this is not the right thing to do; but as far as I see this HostConfigEntry doesn't handle quoted values at all.

tomaswolf:
> When running the unit test suite that read my `~/.ssh/config` file

Completely unrelated: I noticed this, too, and I think this is a bug in the test setups. They should _never_ read the real ~/.ssh/config, that may make tests fail depending on what a user's config contains. Tests should _always_ use a mocked/faked home and ~/.ssh directory, and a mocked/faked config with known static content (maybe empty).

sshd-common/src/main/java/org/apache/sshd/client/config/hosts/HostConfigEntry.java
alex-sherwin: I see, this is re-used for various types of config file values.  In the case I'm trying to fix it's limited to `Host` only, but I suppose the problem would be generic to any kind of config that supports multiple values (not sure how many of those exist)

I could change the PR to work inside the `if (HOST_CONFIG_PROP.equalsIgnoreCase(key)) {` block, but it would work differently, at this point it would have to filter out empty string values from the list that was parsed incorrectly

alex-sherwin: In searching the `ssh_config` man page for "multiple" and "Multiple" there's quite a few directives that support multiple values, but how they are specified is all over the place. Some are CSV's, some require repeating the directive, some say it supports multiple values but fails to specify how to do so.  

`Include`, for example, contains path-like values and can support multiple values, but it fails to saw how to accomplish this (I'm going to assume multiple `Include` directives, but, it's not explicitly stated like `IdentityFile`)

So it is probably prudent to make this fix in the context of `Host` only to avoid any possible conflict with any other directive-specific

alex-sherwin:
I updated this to only work based on the parsed `Host` entries so it cannot impact anything else

sshd-common/src/main/java/org/apache/sshd/common/util/GenericUtils.java
```diff
@@ -196,6 +196,49 @@ public static boolean isNotEmpty(CharSequence cs) {
...
+     * StringUtils.isBlank("")        = true
+     * StringUtils.isBlank(" ")       = true
+     * StringUtils.isBlank("bob")     = false
+     * StringUtils.isBlank("  bob  ") = false
+     * </pre>
+     *
+     * @param  cs the CharSequence to check, may be null
+     * @return    {@code true} if the CharSequence is null, empty or whitespace only
+     * @since     2.0
+     * @since     3.0 Changed signature from isBlank(String) to isBlank(CharSequence)
```
tomaswolf, line 219: So this was copied from somewhere. Looks like StringUtils from org.apache.commons.lang, so license-wise OK.

I'll remove the misleading "since" tags and fix the comment.

tomaswolf (commented):
LGTM. I'll rebase and submit.

tomaswolf:
I've merged this as commit 267887a8.

PR #197, opened 2021-05-26
SSHD-1168, SSHD-1171, SSHD-1172: OpenSshCertificate fixes

tomaswolf:
Fixes:

- Comparing certificate timestamps as unsigned longs
- Test certificates without expiration date
- Checking for correct certificate types (user/host) being used in correct contexts (host key, user auth)


alex-sherwin:
I have a pending PR I was going to open tonight that touches some of this too, and in mine I changed the `OpenSshCertificate` valid before/after variables to `Instant` as it feels like a better developer experience to interact with (in my opinion).

I also found that there is some additional invalid encoding in the OpenSshCertificate code from the original PR that has gone unnoticed with the Critical Options/Extensions data.

My upcoming PR was to include code which can easily generate signed OpenSshCertificates, but ended up making the aforementioned fixes/changes as well (preview: https://github.com/alex-sherwin/mina-sshd2/pull/1)

@tomaswolf Would you prefer if I un-do my change of using `Instant` before submitting the final PR?

tomaswolf:
@alex-sherwin sorry about that. I had been thinking about Instant, too, but in the end I decided to just remove these two methods since I didn't need them. Converting to Instant would also have to take care of that unsigned long issue, and I didn't see a way to do that cleanly; best I could think of was to saturate the 64bit value at LONG_MAX. But if you have a clean solution I have nothing against using Instant at all.

If your change undoes some of what I did and solves it differently that's no problem.

PR #198, opened 2021-05-27
SSHD-1166 - Support creating signed OpenSSH Certificates

alex-sherwin:
This adds new OpenSshCertificateSigner & OpenSshCertificateSignRequest types which are used to use a CA KeyPair to sign a OpenSshCertificateSignRequest, resulting in an instance of OpenSshCertificate with the signature bytes filled in.

The resulting OpenSshCertificate instance is completely valid and can be serialized over the wire using the Buffer.put*PublicKey functions or serialized using the existing OpenSSHKeyPairResourceWriter class.

This change also:

* Fixed a subtle bug observed in KeyUtils mapping data, where order matters for discovery in a few places.  There are places which iterate over the map entries and check if a unknown Key instance is an instance of the registered map PublicKey or PrivateKey type.  Unfortunately, some of these registrations register `java.security.PrivateKey`, which all private keys are an instance of, thus matching this registration (which may actually be wrong).  By making order matter here, all more specific registered types have a chance to succeed first.
* Changed the data type of OpenSshCertificate.validAfter/validBefore from a long to Instant for better developer UX (the long value encoded is actually epoch seconds, which is uncommon for developers to work with in Java).  This change also makes it nullable, which helps with the concept of representing "forever" signed OpenSSH Certificates.
* Added some constants to OpenSshCertificate to represent the magic forever epochs
* Added a new OpenSshCertificate.CriticalOption data type and modified the criticalOptions/extensions types to use it.  This allows for proper representation of critical option (and extension) data, which are not simply strings (they are in fact key/value pairs, which often happen to have no data value, but, still need to be encoded as such).
* Fixed OpenSSHCertificateDecoder encoding of a OpenSshPublicKey to call Buffer.putRawPublicKey, which properly prepends the key type before writing the key data.  Existing code paths for OpenSshCertificate usage didn't previously hit this, so it went unnoticed.
* updated Buffer with specific encode/decode functions for the new OpenSshCertificate.CriticalOption data type.  The existing code was encoding/decoding Critical Options & Extensions incorrectly (as CSV data).  This went unnoticed though, since generally speaking critical options are not used (so this was never checked by anything) and extensions are an optional implementation that must be ignored by a server if they are not recognized (so, silently ignored).  I added a unit test which now asserts decoding valid, existing OpenSSH certificates, and encoding/decoding ones which were generated by the features in this PR.


alex-sherwin:
I believe the build failure is unrelated / transient...

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificate.java
```diff
@@ -52,26 +57,34 @@
...
+     * When null, implies forever
      */
-    long getValidAfter();
+    Instant getValidAfter();
+
+    default long getValidAfterEpochSeconds() {
+        if (getValidAfter() == null) {
+            return VALID_AFTER_FOREVER_EPOCH;
+        }
+        return getValidAfter().getEpochSecond();
```
tomaswolf, line 68: I don't understand why you handle validAfter == 0 specially. That would be Instant.EPOCH.

I think you can handle all values in the range 0 .. Long.MAX_VALUE normally. If it's greater, the signed value will appear as < 0. Here I think handling the ~0L case ("forever") as null is fine; all others I'd map to Long.MAX_VALUE.

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificate.java
```diff
@@ -35,6 +36,10 @@
     int SSH_CERT_TYPE_USER = 1;
     int SSH_CERT_TYPE_HOST = 2;
 
+    long VALID_AFTER_FOREVER_EPOCH = 0L;
+    // note that this is the max value of a uint64 (java.lang.Long.MAX_VALUE is the max of a *signed* int64)
+    long VALID_BEFORE_FOREVER_EPOCH = 0xffffffffffffffffL;
```
tomaswolf, line 41: Either write this as ~0L, or use underscores to make this more readable: 0xffff_ffff_ffff_ffffL is much easier to read, and one immediately sees that there are indeed 16 'f's.

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificate.java
```diff
@@ -91,7 +104,83 @@
...
+            if (o == null || getClass() != o.getClass()) {
+                return false;
+            }
+
+            CriticalOption that = (CriticalOption) o;
+
+            if (name != null ? !name.equals(that.name) : that.name != null) {
+                return false;
+            }
+            return data != null ? data.equals(that.data) : that.data == null;
```
tomaswolf, line 175: 1. Using Objects.equals() and Objects.hash() in hashCode() below would simplify this; no need to check for null.
2. equals and hashCode based on non-final fields is not a good idea. If a CriticalOption object is stored in a HashMap and then name or data is modified, you'll never find it again.

I'd make the fields final and drop the setters.

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificateImpl.java
```diff
@@ -229,8 +228,81 @@ public String toString() {
...
+            return false;
+        }
+        if (caPubKey != null ? !caPubKey.equals(that.caPubKey) : that.caPubKey != null) {
+            return false;
+        }
+        return Arrays.equals(signature, that.signature);
+    }
+
+    @Override
+    public int hashCode() {
```
tomaswolf, line 291: Again: hashCode() on non-final fields.

With non-final fields I think this should not have content-based equals() or hashCode(). As soon as clients can modify fields, this is asking for trouble.

sshd-common/src/main/java/org/apache/sshd/common/config/keys/KeyUtils.java
```diff
@@ -47,21 +47,8 @@
 import java.security.spec.ECParameterSpec;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.RSAPublicKeySpec;
+import java.util.*;
```
tomaswolf, line 50: I don't think Apache MINA sshd uses wildcard imports?

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificate.java
```diff
@@ -52,26 +57,34 @@
...
 
     /**
-     * Retrieves the time in number of seconds since the {@link java.time.Instant#EPOCH} at which this certificate
-     * becomes or became invalid.
+     * When null, implies forever
      *
-     * @return the number of seconds since the Instant.EPOCH <em>as an unsigned 64bit value</em>
-     * @see    {{@link #isValidNow(OpenSshCertificate)}
+     * @return
      */
```
tomaswolf, line 75: Why strip down the javadoc?

tomaswolf (commented):
Some comments after a first glance. Haven't played with this yet.

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificate.java
```diff
@@ -91,7 +104,83 @@
...
+     *
+     * Critical Options is a set of bytes that is
+     *
+     * [overall length][name(string)][data(string)]...
+     *
+     * Where each Critical Option is encoded as a name (string) and data (string)
+     *
+     * Then the entire name + data strings are added as bytes (which will get a length prefix)
+     */
+    class CriticalOption {
```
tomaswolf, line 123: This is used for both getCriticalOptions() and for getExtensions(). Extensions are not critical options. I'd rename this to just "Options" or "CertificateOption" or some such.

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificate.java
alex-sherwin: My observation was that for the `validAfter` field, a certificate has a uint64 encoded as `0` in the case of a "forever" certificate, so my intention was to treat this as a special case in two ways:

1. during decoding (in `OpenSSHCertPublicKeyParser`), when it decodes the bytes as a uint64 `0`, leave the `Instant` as null (to signify forever)
2. On `OpenSshCertificate.getValidAfterEpochSeconds` (simply a helper), apply this same logic in reverse, if the value is null, return the magic "forever" epoch that should be encoded

The intention would be that the business code a developer writes only needs to interact with the `Instant` (null for forever, otherwise set a value), and `OpenSshCertificate.getValidAfterEpochSeconds` and `OpenSshCertificate.getValidBeforeEpochSeconds` are simply used when encoding the bytes of the certificate into a `Buffer`

tomaswolf: Can't validAfter be legally zero if I specify the epoch as from date? I don't see the zero value as special.

But the conversion really has to handle 64bit validLong > Long.MAX_VALUE, also for validAfter. Effectively that would be a certificate that is never valid. But `System.out.println(Instant.ofEpochSecond(-2L))` (0xffff_ffff_ffff_fffeL) yields `1969-12-31T23:59:58Z`.

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificateImpl.java
alex-sherwin: Removed equals/hashCode entirely from OpenSshCertificateImpl, this was only added to make equality checks in the test case simpler, but I just moved the equality checks to the unit test instead

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificate.java
alex-sherwin: I don't think it can be legally `0` and mean "forever"

I just re-read the man page for `ssh-keygen`'s `validity_interval` and you can use the magic values `always` for the `validAfter` and `forever` for `validBefore`

But, when encoded, there is no special flag to indicate this, they are still just epoch (seconds) uint64's, where `always` (`validAfter`) is encoded as `0` and `forever` (`validBefore`) is uint64 max value

I manually verified this by generating certs and looking at the encoded bytes, and you can also see it like this, where generating a certificate using `-V 'always:20220101' versus `-V '19790101:20220101` yield identical values.

If you inspect the cert using `ssh-keygen -L -f [file]` when using a cert generated with `-V '19790101:20220101`, as you would now expect, it prints a human-friendly output of what "always" would have shown.

What's more, is if you move the date back before UTC 1970, ssh-keygen still encodes it as `0` (which I guess they have to, since it's a uint64) 

... so maybe when they defined the spec they assumed (probably rightly so) that no one would be trying to enforce a `validAfter` time on a server that was pretending to run pre 1970?

So, it seems to me, like they do treat `0` and max uint64 as a magic value.  However, it is also entirely possible this magic behavior is just for the human-friendly output of the CLI tools...

I'm happy to change this to behave how you think it should, but I'm not really sure what that should be

tomaswolf: My point was not that someone would try to specify pre-1970 dates. The point was: if a certificate has a validAfter/Before >= 0x8000_0000_0000_0000 (unsigned long), that's a time some 300 billion years in the future. If this is passed to Instant.ofEpochSecond(), you get an exception. Some other large values (>= 0xff8f_e310_1464_1400 in unsigned long; Instant.MIN_SECOND) are even much farther in the future but will be taken as negative values and yield pre-1970 instants in Java instead.

Exactly:
> -V 'always:20220101' versus -V '19700101:20220101' yield identical values.

So zero is not special; but OpenSSH has a special token "always" for it.

If the 64bit value is >= Instant.MAX_SECOND (0x70_1cd2_fa95_78ffL) or, interpreted as signed, is <= Instant.MIN_SECOND, converting the 64bit value to an Instant is going to be problematic.

So... full round-tripping with Instant is not possible. If you want full round-tripping capabilities, long and being careful as a developer is the only way.

For _setting_ validAfter/validBefore, Instant is the right interface if there is a validation that the `value == null || value.compareTo(Instant.EPOCH) >= 0`, serializing null as 0xffff_ffff_ffff_ffffL and Instant.getEpochSecond() otherwise. For decoding the timestamp of a certificate, one would have to handle the special cases:

- value in the range [0 .. Instant.MAX_SECOND]: Instant.ofEpochSecond(value)
- 0xffff_ffff_ffff_ffffL: null
- any other: Instant.MAX

Instant.MAX is still nearly a billion years in the future, which is good enough for all practical purposes.

alex-sherwin: Would you prefer to continue to use Instant and layer on these logical constraints, or go back to long's and apply fewer logical constraints (probably just a >= 0 check to ensure it's treated as a signed value)?

I would rather continue to use Instant, even with the constraints, it's still fine for all practical purposes

tomaswolf: I think it boils down to "do we need full round-tripping?"

I lack experience in this area. I like interfaces with Instant better, too, but I'm worried about not being able to represent the full range of possible values. Couldn't this become a problem for verifying a certificate signature? At least internally I think we'll have to keep the raw 64bit value for this case. Also, if we ever want to read a received certificate, pass it around as OpenSshCertificate object, and then write it out again. We mustn't change the bits of the validAfter/validBefore fields, that would invalidate the signature.

For creating OpenSshCertificateSignRequests, Instant is OK (with validation for null or >= epoch).

sshd-core/src/main/java/org/apache/sshd/certificate/OpenSshCertificateSignRequest.java
```diff
@@ -0,0 +1,136 @@
...
+import java.security.PublicKey;
+import java.time.Instant;
+import java.util.Collection;
+import java.util.List;
+
+import org.apache.sshd.common.config.keys.OpenSshCertificate;
+
+/**
+ * Holds all the data necessary to create a signed OpenSSH Certificate
+ */
```
tomaswolf, line 31: How about a builder (with fluent setters) for this instead of a signer and a sign request?
```
OpenSshCertificate cert =
  new OpenSshCertificateBuilder()
    .setType(OpenSshCertificate.Type.USER)
    .setPublicKey(keyToSign)
    .setValidAfter(Instant.now())
    .setValidBefore(null) // forever; don't expire
    .set....
    .sign(caKey);
```

alex-sherwin: I like that idea, I will make that change

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificate.java
alex-sherwin: All very good points, I'll make adjustments to use longs as raw storage and some optional helper(s) for Instant (probably only useful on the setter, based on our discussion)

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificate.java
alex-sherwin: Restored these javadocs

sshd-core/src/main/java/org/apache/sshd/certificate/OpenSshCertificateSignRequest.java
alex-sherwin: This was implemented

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificate.java
alex-sherwin: I restored the raw storage to `long` and experimented with a few ideas to try and give a better developer UX, but, the best I could come up with was https://github.com/apache/mina-sshd/pull/198/commits/ffaa35ae566aa04a79a3bf94625af4d970756a55#diff-0b82051894f817aa54f0e169b2bb96d0fce8a989838e0dd4574ca7ab395d5f9eR199-R201

Which is just to have a helper that can set the values as an `Instant` but not before `Instant.EPOCH`

alex-sherwin:
This PR should be ready for re-review, all the suggestions have been implemented

sshd-common/src/main/java/org/apache/sshd/common/util/buffer/Buffer.java
```diff
@@ -341,6 +341,39 @@ public String getString() {
...
+
+        List<OpenSshCertificate.CertificateOption> list = new ArrayList<>();
+
+        while (criticalOptionsBuf.available() > 0) {
+            String name = criticalOptionsBuf.getString(charset);
+            String data = GenericUtils.trimToEmpty(criticalOptionsBuf.getString(charset));
+            list.add(new OpenSshCertificate.CertificateOption(name, data.length() > 0 ? data : null));
+        }
+
+        return list.size() > 0 ? list : null;
```
tomaswolf, line 374: In general i think returning null for collections is bad, but I see that here it may make sense given how these lists are passed around through getters and setters. Fixing that would have to start already at OpenSshCertificateImpl.getCertificateOptions().

So OK, let's leave it with returning null. I'd use list.isEmpty(), though.

Please add javadoc to OpenSshCertificate documenting that it may return null or an empty list.

sshd-common/src/main/java/org/apache/sshd/common/util/buffer/Buffer.java
```diff
@@ -341,6 +341,39 @@ public String getString() {
...
+     *
+     * Critical Options is a set of bytes that is
+     *
+     * [overall length][name(string)][data(string)]...
+     *
+     * Where each Critical Option is encoded as a name (string) and data (string)
+     *
+     * Then the entire name + data strings are added as bytes (which will get a length prefix)
+     *
+     * @return The parsed result
```
tomaswolf, line 360: Please document here the null return when there are no options.

Also missing @param in the javadoc.

sshd-common/src/main/java/org/apache/sshd/common/util/buffer/Buffer.java
```diff
@@ -761,6 +794,40 @@ public void putStringList(Collection<?> objects, Charset charset, boolean prepen
...
+     * Critical Options is a set of bytes that is
+     *
+     * [overall length][name(string)][data(string)]...
+     *
+     * Where each Critical Option is encoded as a name (string) and data (string)
+     *
+     * Then the entire name + data strings are added as bytes (which will get a length prefix)
+     */
+    public void putCertificateOptions(List<OpenSshCertificate.CertificateOption> options, Charset charset) {
+        int numObjects = GenericUtils.size(options);
```
tomaswolf, line 814: Missing @param in the javadoc. Should state that options may be null or empty.

sshd-common/src/main/java/org/apache/sshd/common/config/keys/OpenSshCertificateImpl.java
```diff
@@ -187,15 +190,47 @@ public void setValidAfter(long validAfter) {
         this.validAfter = validAfter;
     }
 
+    /**
+     * If null, uses {@link OpenSshCertificate#MIN_EPOCH}
+     *
+     * @param validAfter {@link Instant} to use for validBefore
```
tomaswolf, line 196: Copy-paste error. This is validAfter, not validBefore.

sshd-core/src/main/java/org/apache/sshd/certificate/OpenSshCertificateBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+    /**
+     * If null, uses {@link OpenSshCertificate#MIN_EPOCH}
+     *
+     * @param validAfter {@link Instant} to use for validBefore
+     */
+    public OpenSshCertificateBuilder validAfter(Instant validAfter) {
+        if (null == validAfter) {
+            validAfter(OpenSshCertificate.MIN_EPOCH);
+        } else {
+            validAfter(validAfter.getEpochSecond());
```
tomaswolf, line 133: Missing validation that the value is >= epoch.

sshd-core/src/main/java/org/apache/sshd/certificate/OpenSshCertificateBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+    /**
+     * If null, uses {@link OpenSshCertificate#MAX_EPOCH}
+     *
+     * @param validBefore {@link Instant} to use for validBefore
+     */
+    public OpenSshCertificateBuilder validBefore(Instant validBefore) {
+        if (null == validBefore) {
+            validBefore(OpenSshCertificate.MAX_EPOCH);
+        } else {
+            validBefore(validBefore.getEpochSecond());
```
tomaswolf, line 152: Missing validation that the value is >= epoch.

sshd-core/src/main/java/org/apache/sshd/certificate/OpenSshCertificateBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+import java.security.KeyPair;
+import java.security.PublicKey;
+import java.security.SecureRandom;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.sshd.client.auth.pubkey.KeyPairIdentity;
```
tomaswolf, line 31: Does this really need client.auth.pubkey.KeyPairIdentity? Can't we implement this differently to avoid importing something from *.client.*? Then this could be moved to sshd-common, too.

See my proposal on how to sign below.

sshd-core/src/main/java/org/apache/sshd/certificate/OpenSshCertificateBuilder.java
```diff
@@ -0,0 +1,273 @@
...
+
+        // create signature using CA Keypair and OpenSSH Certificate data
+        final Map.Entry<String, byte[]> sigEntry = caIdentity.sign(null, null, toBeSignedBuf.getCompactData());
+
+        // set the signature (need to prefix this data section with the sig algorithm used)
+        final ByteArrayBuffer tmpBuffer = new ByteArrayBuffer();
+        // signature algorithm
+        tmpBuffer.putString(sigEntry.getKey());
+        // signature
+        tmpBuffer.putBytes(sigEntry.getValue());
```
tomaswolf, line 249: This signing is a bit roundabout. Inlining the relevant part of KeyPairIdentity.sign() here would give something like
```
String algo = KeyUtils.getKeyType(caKeypair.getPublic());
SignatureFactory factory = BuiltinSignatures.fromFactoryName(algo);
Signature signer = factory == null ? null : factory.create();
ValidateUtils.checkNotNull(signer, "No signer could be located for key type=%s", algo);
signer.initSigner(null, caKeypair.getPrivate());
signer.update(null, toBeSignedBuf.getCompactData());

byte[] signature = signer.sign(null);

// set the signature (need to prefix this data section with the sig algorithm used)
final ByteArrayBuffer tmpBuffer = new ByteArrayBuffer();
// signature algorithm
tmpBuffer.putString(algo);
// signature
tmpBuffer.putBytes(signature);
```
Then this builder could be moved to sshd-common, too.

tomaswolf (commented):
Sorry about the delay; was too busy elsewhere. Generally looks OK, some minor things, and I'd implement signing differently and then move the builder to sshd-common, too.

The commit message still mentions Signer & SignRequest instead of Builder. It should also have an empty line after the title line.

tomaswolf:
In the `GenerateOpenSSHClientCertificateTest` the generated certificate apparently always uses rsa-sha2-512 for signing with RSA CA keys. Does the `OpenSshCertificateBuilder` need another setter `setAlgorithm(String algorithm)` to be able to specify exactly which signature should be generated? ("ssh-rsa", "rsa-sha2-256", or "rsa-sha2-512".)

One would have to verify that the algorithm set through this setter, if any, was compatible with the CA key type at all.

tomaswolf:
@alex-sherwin I got now a version of this change with all this fixed (builder still in sshd-core, though), plus with a problem in  signature verification fixed. Would it be OK if I push onto this PR?

alex-sherwin:
@tomaswolf Sure that would be fine with me

Thanks for that, apologies I didn't get to it myself yet

alex-sherwin:
@tomaswolf The changes you pushed LGTM

tomaswolf:
Merged. Thanks a lot, Alex!

vukzeka:
@alex-sherwin @tomaswolf this is amazing feature, allowing applying best practices for SSH auth but also more secure.
Though, I have noticed one issue and thought good to report.  Following is bug I have created https://issues.apache.org/jira/browse/SSHD-1266
If something I can support further please just let me know.

PR #199, opened 2021-06-11
Sshd 1181

gnodet:


gnodet:
I think the number of requests sent for downloading small files is not really optimal. This could be improved.

PR #200, opened 2021-06-15
[SSHD-1182] skip() doesn't work properly in SftpInputStreamAsync

pstetsuk:
need to modify `requestOffset` when `clientOffset` is changed from 0 to specified value


tomaswolf:
Squashed and merged as commit 6a8e2eec3. Thank you, Pavel!

PR #201, opened 2021-07-12
[SSHD-1197] Fix race condition in KEX

tomaswolf:
There was a window in AbstractSession.requestNewKeyExchange()
during which the KEX state was set already to INIT, but the
caller's proposal not set yet. If the peer had also decided to
start a new key exchange, it was possible that the peer's KEX_INIT
message arrived and was handled in that window and then proceeded
with wrong or uninitialized data.

KEX negotiation must start only when both proposals are indeed
available. Thus wait when having received a KEX_INIT from the peer
until our own proposal has been prepared before continuing with the
negotiation.

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
@@ -151,6 +151,10 @@ public abstract class AbstractSession extends SessionHelper {
     protected final AtomicReference<KexState> kexState = new AtomicReference<>(KexState.UNKNOWN);
     protected final AtomicReference<DefaultKeyExchangeFuture> kexFutureHolder = new AtomicReference<>(null);
 
+    // The kexInitializedFuture is fulfilled when this side (client or server) has prepared its own proposal. Access is
+    // synchronized on kexState; uses kexLock as lock.
+    protected DefaultKeyExchangeFuture kexInitializedFuture;
+
     /*
      * SSH packets encoding / decoding support
      */
@@ -173,6 +177,7 @@ public abstract class AbstractSession extends SessionHelper {
     protected int decoderLength;
     protected final Object encodeLock = new Object();
     protected final Object decodeLock = new Object();
+    protected final Object kexLock = new Object();
     protected final Object requestLock = new Object();
 
     /*
@@ -696,7 +701,22 @@ public abstract class AbstractSession extends SessionHelper {
     protected void doKexNegotiation() throws Exception {
         if (kexState.compareAndSet(KexState.DONE, KexState.RUN)) {
             sendKexInit();
-        } else if (!kexState.compareAndSet(KexState.INIT, KexState.RUN)) {
+        } else if (kexState.compareAndSet(KexState.INIT, KexState.RUN)) {
+            DefaultKeyExchangeFuture initFuture;
+            synchronized (kexState) {
+                initFuture = kexInitializedFuture;
+                if (initFuture == null) {
+                    initFuture = new DefaultKeyExchangeFuture(toString(), kexLock);
+                    kexInitializedFuture = initFuture;
+                }
+            }
+            // requestNewKeyExchange() is running in some other thread: wait until it has set up our own proposal. The timeout
+            // is a last resort only to avoid blocking indefinitely in case something goes catastrophically wrong somewhere; it
+            // should never be hit. If it is, an exception will be thrown.
+            //
+            // See https://issues.apache.org/jira/browse/SSHD-1197
+            initFuture.await(CoreModuleProperties.KEX_PROPOSAL_SETUP_TIMEOUT.getRequired(this));
+        } else {
             throw new IllegalStateException("Received SSH_MSG_KEXINIT while key exchange is running");
         }
 
@@ -720,6 +740,9 @@ public abstract class AbstractSession extends SessionHelper {
         }
         kex.init(v_s, v_c, i_s, i_c);
 
+        synchronized (kexState) {
+            kexInitializedFuture = null;
+        }
         signalSessionEvent(SessionListener.Event.KexCompleted);
     }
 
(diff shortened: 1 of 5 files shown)


tomaswolf:
@lgoldstein , do you agree with this way of fixing this?

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
```diff
@@ -696,7 +702,18 @@ protected void handleKexInit(Buffer buffer) throws Exception {
...
+            DefaultKeyExchangeFuture initFuture;
+            synchronized (kexState) {
+                initFuture = kexInitializedFuture;
+                if (initFuture == null) {
+                    initFuture = new DefaultKeyExchangeFuture(toString(), kexLock);
+                    kexInitializedFuture = initFuture;
+                }
+            }
+
+            initFuture.await();
```
lgoldstein, line 715: I am not comfortable with an infinite wait - just in case some deadlock/bug occurs. Let's define a `CoreModuleProperties#MAX_KEX_INIT_WAIT_TIMEOUT` and use it here (I think ~30 seconds as default value should do, but you are welcome to use some other value). If it expires then throw an exception (which will close the session(.

sshd-sftp/src/test/java/org/apache/sshd/sftp/client/SftpTransferTest.java
```diff
@@ -23,22 +23,39 @@
...
+@RunWith(Parameterized.class)
 public class SftpTransferTest extends AbstractSftpClientTestSupport {
 
     private static final int BUFFER_SIZE = 8192;
 
-    public SftpTransferTest() throws IOException {
+    private final long rekeyBlockSize;
+
+    public SftpTransferTest(long rekeyBlockSize) throws IOException {
         super();
```
lgoldstein, line 49: Please remove the `super()` call - the coding convention is that if there is no other statement in the constructor, then we call `super()` explicitly. Otherwise (i.e., more statements) we drop it.

lgoldstein (changes requested):

sshd-sftp/src/test/java/org/apache/sshd/sftp/client/SftpTransferTest.java
tomaswolf: Done.

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
tomaswolf: await(timeout) throws an exception already if it times out. Added a timeout.

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
```diff
@@ -109,6 +109,16 @@
...
 
+    /**
+     * A last-resort timeout for waiting after having received a KEX_INIT message from the peer until we have prepared
+     * our own KEX proposal. This timeout should actually never be hit unless there is a serious deadlock somewhere and
+     * the session is never closed.
+     *
+     * @see <a href="https://issues.apache.org/jira/browse/SSHD-1197">SSHD-1197</a>
+     * @see #doKexNegotiation()
+     */
+    private static final Duration KEX_PROPOSAL_SETUP_TIMEOUT = Duration.ofSeconds(42);
```
lgoldstein, line 120: Please use a `CoreModuleProperties#KEX_PROPOSAL_SETUP_TIMEOUT` as initially suggested so that users can decide to shorten/lengthen this value instead of it being hardcoded

tomaswolf: Hmmm. If we make that configurable, users can very easily shoot themselves in the foot by setting a (very) low value, like 0. I don't like that. Lengthening this should never be necessary. The thread that is preparing this side's proposal is already running.

lgoldstein: >>>  If we make that configurable, users can very easily shoot themselves in the foot by setting a (very) low value, like 0.

I understand your concern, but personally I believe in total user responsibility - I don't like being told by someone that "they know better", so I am reluctant to impose my opinion on others (BTW, this is also why I always insist that all our internal variables and/or methods be accessible either via getters or as public/protected members).

In this context, the same could be said for *all* our configuration values - users can shoot themselves in the foot by choosing a wrong or conflicting value. However, I believe this is a good thing - they will learn... IMO, we are not supposed to protect our users from mistakes that occur because they do not bother to read the documentation or understand the consequences of their actions. In this case, perhaps we could do something that would satisfy both our concerns:
```java
Duration waitTime = CoreModuleProperties.KEX_PROPOSAL_SETUP_TIMEOUT.getRequired(this);
ValidateUtils.checkTrue(waitTime.toMillis() > 0L, "Infinite/Undefined value N/A: %d", waitTime);
```

>> Lengthening this should never be necessary. The thread that is preparing this side's proposal is already running.

While it seems reasonable, we don't know what weird usages our users might encounter or what unforeseen circumstances. Therefore, I believe we should give them the flexibility to decide for themselves. 

Bottom line - I trust your judgement, and if you still feel that a constant value is more appropriate I will not object to it. In any case, let's also update the README (perhaps add a KEX section) and explain this mechanism so that if there are future questions or suspicions of bugs around this we will have a clear picture of the implementation and the choice of either a constant or configurable timeout value.

tomaswolf: I've been thinking more about this -- I'll add it with validation for now. Reason: we are indeed waiting until sendKexInit has sent the buffer (or rather, has queued it for sending on the IOSession). But I'm still not happy with that whole KEX implementation. In particular, this having to wait until the buffer is sent should not be necessary, but with the current implementation is unavoidable because of the `ReservedSessionMessagesHandler.sendKexInitRequest` hook added for SSHD-1097, which can modify the buffer and even might send it on its own.

That hook was added recently, and is used only in that "endless tarpit" thing, where it actually only suppresses writing the buffer. If that were changed to something that could only say "don't send the message", I could extract that whole KEX state machine in a separate class and it would even possible to do an implementation that didn't have to wait at all for this race condition case handled here.

I wouldn't add a full explanation of how KEX works, that's explained in the RFCs. But the new timeout property will have a comment explaining what it is for.

lgoldstein: Sounds reasonable to me

lgoldstein (approved):

PR #204, opened 2021-10-15
[SSHD-1216] Server-side server-sig-algs KEX extension

tomaswolf:
Provide a default implementation for the server-side SSH_MSG_EXT_INFO
message sending including the server-sig-algs KEX extension.[1]

A server that implements the rsa-sha2-512 or rsa-sha2-256 signature
algorithms should implement this extension, otherwise even clients that
also have these signature algorithms may fall back to ssh-rsa to avoid
authentication penalties.[2]

Apache MINA sshd servers by default do implement the SHA-2 RSA
signatures, and an Apache MINA sshd client by default does request KEX
extension information and does handle the server-sig-algs extension. So
an Apache MINA sshd server should by default implement this extension.

This implementation sends the server-sig-algs extension record once at
the end of the initial key exchange.

[1] https://tools.ietf.org/html/rfc8308
[2] https://tools.ietf.org/html/rfc8332#section-3.3

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/client/ClientBuilder.java
+++ b/sshd-core/src/main/java/org/apache/sshd/client/ClientBuilder.java
@@ -80,15 +80,9 @@ public class ClientBuilder extends BaseBuilder<SshClient, ClientBuilder> {
     protected HostConfigEntryResolver hostConfigEntryResolver;
     protected ClientIdentityLoader clientIdentityLoader;
     protected FilePasswordProvider filePasswordProvider;
-    protected KexExtensionHandler kexExtensionHandler;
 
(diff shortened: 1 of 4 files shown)


benhumphreys:
Many thanks for this @tomaswolf
I've dropped a snapshot build of your branch into Bitbucket Server to do some testing and it looks great! I've tested with OpenSSH 8.1p1 and 8.8p1 and it works as expected, with both resulting in the use of `rsa-sha2-512`.

I've also tested with OpenSSH 6.6.1p1 and it works as I'd expect; it doesn't send `server-sig-algs` because the client doesn't send `ext-info-s`.

PR #207, opened 2021-10-29
[SSHD-1220] Cache SFTP attributes on SftpPaths

tomaswolf:
This greatly reduces the number of (L)STAT calls made to get SFTP
attributes. Also, listing directories in reply to SSH_FXP_OPENDIR
and SSH_FXP_READDIR can in this way still go through the normal
`java.io.file.FileSystem` abstraction. There is no need anymore to
bypass the file system to resolve SSHD-1217.

The attributes cached on the `Path` objects returned from the
`DirectoryStream` are used only for listing the directory. This
is a similar optimization as was done in the Java libraries for
a `FileTreeWalker` operating on an Unix filesystem, which may use
cached attributes via the internal interface
`sun.nio.fs.BasicFileAttributesHolder`.

Caching is also enabled selectively in the up-front checks for file
accessibility, which also caused repeated (L)STAT calls. Rewrite
`IoUtils.checkFileExists()` to load attributes only once.

The new operations for caching attributes on `SftpPath` instances are
protected only; they are not intended to be used by client code. For
access in the SFTP implementation itself, paths are now always created
as `SftpPathImpl` instances, which inherit from `SftpPath` and publish a
safe `withAttributeCache()` method. Theoretically client code could still
access these methods, but then it's clear for everyone that these are
internal operations.


sshd-common/src/main/java/org/apache/sshd/common/util/io/IoUtils.java
```diff
@@ -372,11 +374,23 @@ public static String getFileOwner(Path path, LinkOption... options) throws IOExc
...
+        }
+        try {
+            if (followLinks) {
+                path.getFileSystem().provider().checkAccess(path);
+            } else {
+                Files.readAttributes(path, BasicFileAttributes.class, LinkOption.NOFOLLOW_LINKS);
+            }
             return Boolean.TRUE;
-        } else if (Files.notExists(path, options)) {
+        } catch (NoSuchFileException e) {
```
lgoldstein, line 391: I am not sure these 2 lines are equivalent - it  assumes that the implementation throws `NoSuchFileException` but this might not be the case. It **should** be, but we know that in reality there are some who do not. Furthermore, strange as it may seem, **both** `File.exists` and `File.notExists`  might return *false* (for "special" files - e.g., Unix domain sockets) - this is why return `Boolean` *null* to indicate such a file. Perhaps a better way would be:

```java
try {
    ... check exists ...
} catch (FileNotFoundException e) {
    return Boolean.FALSE;
}

return File.notExists(path) ? Boolean.FALSE : null;
```

sshd-sftp/src/main/java/org/apache/sshd/sftp/client/fs/SftpPath.java
```diff
@@ -25,12 +25,80 @@
...
+            setAttributes(attributes);
+        }
+    }
+
+    /**
+     * Unconditionally set the cached attributes, whether or not this instance's attribute cache is enabled.
+     *
+     * @param attributes the {@link SftpClient.Attributes} to cache
+     */
+    void setAttributes(SftpClient.Attributes attributes) {
```
lgoldstein, line 89: Recommend making this `protected` so implementors can override it

lgoldstein (changes requested):

lgoldstein:
The code is certainly more complex to follow... It seems OK, but I may have missed something. Perhaps you should ask Guillaume to take a look at it as well - maybe he can locate something we both missed.

sshd-common/src/main/java/org/apache/sshd/common/util/io/IoUtils.java
tomaswolf: No. The old code did
```
    public static Boolean checkFileExists2(Path path, LinkOption... options) {
        if (Files.exists(path, options)) {
            return Boolean.TRUE;
        } else if (Files.notExists(path, options)) {
            return Boolean.FALSE;
        } else {
            return null;
        }
    }
```
Now let's inline the `exists()` and `notExists()` calls:
```
    public static Boolean checkFileExists3(Path path, LinkOption... options) {
        try {
            if (followLinks(options)) {
                path.getFileSystem().provider().checkAccess(path);
            } else {
                // attempt to read attributes without following links
                Files.readAttributes(path, BasicFileAttributes.class,
                               LinkOption.NOFOLLOW_LINKS);
            }
            // file exists
            return Boolean.TRUE;
        } catch (IOException x) {
            // does not exist or unable to determine if file exists
            // return false;
            // Let's do the notExists check
        }
        try {
            if (followLinks(options)) {
                path.getFileSystem().provider().checkAccess(path);
            } else {
                // attempt to read attributes without following links
                Files.readAttributes(path, BasicFileAttributes.class,
                               LinkOption.NOFOLLOW_LINKS);
            }
            // file exists
            // return false;
            // In this case we returned TRUE above already
        } catch (NoSuchFileException x) {
            // file confirmed not to exist
            // return true;
            return Boolean.FALSE;
        } catch (IOException x) {
            // return false;
            // We cannot determine it
        }
        return null;
    }
```
As one can see, there are exactly three possibilities:

1. checkAccess/readAttributes doesn't throw an exception: file exists
2. NoSuchFileException is thrown: file doesn't exist
3. Some other IOException is thrown: indeterminate.

Unix domain sockets or not, and FileNotFoundException or not -- the old code did exactly what the new code does, but called checkAccess/readAttributes up to two times, while the new code does so only and exactly once.

Moreover, with attribute caching the second call to readAttributes would be no-op anyway. (Which is not a problem, since if it should fail, it should have failed the first time already.)

sshd-sftp/src/main/java/org/apache/sshd/sftp/client/fs/SftpPath.java
tomaswolf: Refactored anyway. This is now on `SftpPathImpl`.

tomaswolf:
> The code is certainly more complex to follow... It seems OK, but I may have missed something. Perhaps you should ask Guillaume to take a look at it as well - maybe he can locate something we both missed. In view of this I definitely recommend we make this a configurable behavior (default=enabled)

Well, I think you know how I feel about this... Anyway: added such a config. But IMO it only adds unnecessary complication to the code, and users should never set it to false anyway. It really hurts performance. Try the directory listing test with it set to true or false:

- true: takes some 70ms per run on my machine, makes 3 LSTAT calls
- false: takes >2s per run on my machine, makes 10040 LSTAT calls

The caching of attributes is truly _safe_: the unconditionally cached attributes are used _only_ in directory iteration; just like the cached attributes from `sun.nio.fs.BasicFileAttributesHolder` are used exclusively in `FileTreeWalker` in the Java library. And `withAttributeCaching` is used only for short code blocks that are _known_ to only make repeated calls to `readAttributes`, and they clear the cache at the end.

My proposition instead of a pointless config: either we believe this implementation is correct and then check it in without config, or we believe it isn't correct and then we abandon this idea altogether.

tomaswolf:
@gnodet , do you have time to give this code a review?

lgoldstein:
> My proposition instead of a pointless config: either we believe this implementation is correct and then check it in without config, or we believe it isn't correct and then we abandon this idea altogether. And if we can't decide what we believe, then let's also abandon this.

You convinced me - I'm with you on this - let's get rid of the configuration for this feature

sshd-common/src/main/java/org/apache/sshd/common/util/io/IoUtils.java
lgoldstein: I trust your judgement on this

tomaswolf:
All right; thanks. Here it is again without config.

docs/sftp.md
```diff
@@ -532,7 +530,27 @@ if (directoryPath instanceof SftpPath) {
...
+```java
+public void processSftpDirectory(SftpPath directoryPath, BiConsumer<Path, SftpClient.Attributes> process) throws IOException {
+  try (DirectoryStream<Path> dir = Files.newDirectoryStream(directoryPath)) {
+    for (Path path : dir) {
+      if (path instanceof SftpPath) {
+        SftpClient.Attributes attributes = ((SftpPath) path).getAttributes();
+        process.accept(path, attributes);
+      } else {
+        // A DirectoryStream on a directory given by an SftpPath always returns SftpPath instances as elements.
+        throw new IllegalStateException();
```
lgoldstein, line 545: Let's add a message to exception along the lines of "Iterated directory entry=" + path + " is not an SftpPath as expected"

lgoldstein (changes requested):
Just one minor fix...

docs/sftp.md
tomaswolf: Done.

lgoldstein (approved):

PR #210, opened 2021-11-25
[SSHD-1230] Log netty channel traffic on TRACE level

rovarga:
The hard-coded logger is overly verbose by default. Change the log level to TRACE and propagate class identity, so that Logger implementations can correctly enable/disable the messages. Also share a single LoggingHandler for all channels, reducing memory footprint.


tomaswolf:
Thanks, Robert! Much better indeed.

PR #213, opened 2022-03-19
[SSHD-1254] Client side implementation of host bound pubkey auth

tomaswolf:
Provide a KexExtensionParser for it; make the default client KEX
extension handler understand the "publickey-hostbound@openssh.com"
extension, and use it in UserAuthPublicKey.

The extension changes the auth name in the SSH_MSG_USERAUTH_REQUEST
from "pubkey" to "publickey-hostbound-v00@openssh.com" and includes
the server's public key in the message.

See https://github.com/openssh/openssh-portable/blob/807be686/PROTOCOL#L347

### DIFF
--- a/sshd-common/src/main/java/org/apache/sshd/common/kex/extension/parser/HostBoundPubkeyAuthentication.java
+++ b/sshd-common/src/main/java/org/apache/sshd/common/kex/extension/parser/HostBoundPubkeyAuthentication.java
@@ -1,0 +1,72 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.sshd.common.kex.extension.parser;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+
+import org.apache.sshd.common.util.buffer.Buffer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class HostBoundPubkeyAuthentication extends AbstractKexExtensionParser<Integer> {
+
+    public static final String NAME = "publickey-hostbound@openssh.com";
+
+    public static final String AUTH_NAME = "publickey-hostbound-v00@openssh.com";
+
+    public static final HostBoundPubkeyAuthentication INSTANCE = new HostBoundPubkeyAuthentication();
+
+    private static final Logger LOG = LoggerFactory.getLogger(HostBoundPubkeyAuthentication.class);
+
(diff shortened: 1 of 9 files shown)


sshd-core/src/main/java/org/apache/sshd/client/auth/pubkey/UserAuthPublicKey.java
```diff
@@ -321,10 +331,19 @@ protected boolean processAuthDataRequest(ClientSession session, String service,
...
+        boolean doHostBoundAuth = hostBoundPubKeyVersion != null && hostBoundPubKeyVersion.intValue() == 0;
+        if (doHostBoundAuth) {
+            buffer.putString(HostBoundPubkeyAuthentication.AUTH_NAME);
+        } else {
+            buffer.putString(name);
+        }
         buffer.putBoolean(true);
         buffer.putString(algo);
         buffer.putPublicKey(pubKey);
+        if (doHostBoundAuth) {
```
lgoldstein, line 344: I think the log message should include the correctly used algorithm name if *doHostBoundAuth* is *true*

tomaswolf: Done.

tomaswolf:
> Looks good - can we write a unit test for it ?

I'll have to see if I can set up a container test using an OpenSSH 8.9 server.

tomaswolf:
> I'll have to see if I can set up a container test using an OpenSSH 8.9 server.

alpine:3.15 seems to be the newest Alpine, but apk only gives me OpenSSH 8.8p1. The extension is available only as of 8.9p1. See the [OpenSSH release notes](https://www.openssh.com/releasenotes.html). So currently I cannot write any test for this.

lgoldstein:
> alpine:3.15 seems to be the newest Alpine, but apk only gives me OpenSSH 8.8p1. The extension is available only as of 8.9p1. See the [OpenSSH release notes](https://www.openssh.com/releasenotes.html). So currently I cannot write any test for this.

That's fine - let's keep an eye for it when it is released - perhaps open a followup issue on our JIRA to write a test for it or at least check it manually...

tomaswolf:
alpine 3.16 is out now, and it comes with OpenSSH 9.0p1. So it should be possible to write an automated integration test for this now.

sshd-common/src/main/java/org/apache/sshd/common/kex/extension/parser/HostBoundPubkeyAuthentication.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public Integer parseExtension(Buffer buffer) throws IOException {
+        return parseExtension(buffer.array(), buffer.rpos(), buffer.available());
+    }
+
+    @Override
+    public Integer parseExtension(byte[] data, int off, int len) throws IOException {
+        return parseExtension((len <= 0) ? "" : new String(data, off, len, StandardCharsets.UTF_8));
```
lgoldstein, line 49: If `len <= 0` then can return *null* here instead of relying on the called code to throw an exception. It is not a good idea to do code flow control with exception.

lgoldstein (changes requested):

sshd-common/src/main/java/org/apache/sshd/common/kex/extension/parser/HostBoundPubkeyAuthentication.java
tomaswolf: Done.

lgoldstein (approved):
Looks fine to me - nice work

lgoldstein:
I think the main README.md (or the relevant sub-document) should also be updated so as to list this new supported KEX as well.

tomaswolf:
> I think the main README.md (or the relevant sub-document) should also be updated so as to list this new supported KEX as well.

Done. Added a mention of implemented authentication methods.

PR #214, opened 2022-03-31
[SSHD-1257] Make ChannelOutputStream#flush no-op if the stream is already closed

Vlatombe:
[SSHD-1257](https://issues.apache.org/jira/browse/SSHD-1257)

If the command implementation closes the output stream at the end of its execution, the ssh server hangs.
This changes the `ChannelOutputStream#flush` to become a no-op if it has been closed already.

Without change in `src/main` this fails with

```
java.lang.AssertionError: Wrong channel state after 11004 ms.: [TIMEOUT, OPENED]
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.assertTrue(Assert.java:42)
        at org.apache.sshd.server.channel.ChannelSessionTest.closeOutputStream(ChannelSessionTest.java:101)
```
indicating that the session doesn't get closed.

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelOutputStream.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelOutputStream.java
@@ -195,9 +195,7 @@ public class ChannelOutputStream extends OutputStream implements java.nio.channe
     public synchronized void flush() throws IOException {
         Channel channel = getChannel();
         if (!isOpen()) {
-            throw new SshChannelClosedException(
-                    channel.getChannelId(),
-                    "flush(" + this + ") length=" + bufferLength + " - stream is already closed");
+            return;
         }
 
         try {
--- a/sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelOutputStream.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelOutputStream.java
@@ -195,7 +195,9 @@ public class ChannelOutputStream extends OutputStream implements java.nio.channe
     public synchronized void flush() throws IOException {
         Channel channel = getChannel();
         if (!isOpen()) {
-            return;
+            throw new SshChannelClosedException(
+                    channel.getChannelId(),
+                    "flush(" + this + ") length=" + bufferLength + " - stream is already closed");
         }
 
         try {
--- a/sshd-core/src/main/java/org/apache/sshd/server/channel/ChannelSession.java
+++ b/sshd-core/src/main/java/org/apache/sshd/server/channel/ChannelSession.java
@@ -908,7 +908,7 @@ public class ChannelSession extends AbstractServerChannel {
 
         if (!isClosing()) {
             if (out != null) {
-                out.flush();
+                out.close();
             }
             sendEof();
             sendExitStatus(exitValue);
--- a/sshd-core/src/test/java/org/apache/sshd/server/channel/ChannelSessionTest.java
+++ b/sshd-core/src/test/java/org/apache/sshd/server/channel/ChannelSessionTest.java
@@ -58,6 +58,60 @@ public class ChannelSessionTest extends BaseTestSupport {
         super();
     }
 
+    /**
+     * Testing a command closing output stream when it completes
+     */
+    @Test // see SSHD-1257
+    public void closeOutputStream() throws Exception {
+        try (SshServer server = setupTestServer();
+             SshClient client = setupTestClient()) {
+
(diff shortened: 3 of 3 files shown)


jglick (commented):
What does the stack trace look like when `src/main/` is reverted? There is already https://github.com/apache/mina-sshd/blob/f46ca411e8715dc6228d1f5a79544b30d9a333c9/sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelOutputStream.java#L272-L275 so presumably https://github.com/apache/mina-sshd/blob/f46ca411e8715dc6228d1f5a79544b30d9a333c9/sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelOutputStream.java#L281-L282 is not the culprit. So something else is calling `flush()` on a closed stream, which does not feel right.

Vlatombe:
@jglick https://github.com/apache/mina-sshd/blob/da2958172281ef20e51681afe036bd42c9cf843a/sshd-core/src/main/java/org/apache/sshd/server/channel/ChannelSession.java#L910-L912

tomaswolf:
But wouldn't it be more appropriate then to do in`ChannelSession` line 910
```
if (out != null && out.isOpen()) {
  out.flush();
}
```
?

sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelOutputStream.java
```diff
@@ -195,9 +195,7 @@ public synchronized void write(byte[] buf, int s, int l) throws IOException {
     public synchronized void flush() throws IOException {
         Channel channel = getChannel();
         if (!isOpen()) {
```
lgoldstein, line 197: I don't think this is a good idea - what if the user meant to flush the stream but did not notice it is closed. The code would simply "swallow" the exception. AFAIK the stream contract is clear: do not call **any** method after calling *close()* (one *can* call *close()* as many times as one wants since it is idempotent),

lgoldstein (changes requested):

Vlatombe:
> But wouldn't it be more appropriate then to do inChannelSession line 910

It's not as simple, because `isOpen` is only defined on `ChannelOutputStream` but depending on cases it can be wrapped in `LoggingFilterOutputStream`.

See https://github.com/apache/mina-sshd/blob/da2958172281ef20e51681afe036bd42c9cf843a/sshd-core/src/main/java/org/apache/sshd/server/channel/ChannelSession.java#L733-L743

Vlatombe:
Just thought of an easier alternative...

sshd-core/src/main/java/org/apache/sshd/server/channel/ChannelSession.java
```diff
@@ -908,7 +908,7 @@ protected void closeShell(int exitValue, boolean closeImmediately) throws IOExce
 
         if (!isClosing()) {
             if (out != null) {
-                out.flush();
```
lgoldstein, line 911: Not sure about this... "smells fishy" as well , though I cannot put my fiinger on it - will have to think about it.

tomaswolf: `out.close()` works because this is indeed the last use of this stream. Looks fine to me...

Unrelated to this change:

But the code is a bit strange anyway. The intent is obviously to get all data from the command sent before the EOF is sent. But why not also flush or close `err`? What happens if the command writes something to the error stream? `err` gets closed only in the `close()` call in line 916, so that would flush it after the EOF (and the exit status) was sent. That's probably not OK.

And anyway: stderr is traditionally "unbuffered", i.e., flushed on newlines. I don't see anything in ChannelOutputStream that would implement that for SSH_MSG_CHANNEL_EXTENDED_DATA. Maybe that's OK; the command would be responsible for flushing it frequently then if desired.

lgoldstein: > out.close() works because this is indeed the last use of this stream. Looks fine to me...

I trust your judgment @tomaswolf 

>  The intent is obviously to get all data from the command sent before the EOF is sent. But why not also flush or close err? What happens if the command writes something to the error stream? err gets closed only in the close() call in line 916, so that would flush it after the EOF (and the exit status) was sent. That's probably not OK.

I tend to agree with you...

Vlatombe:
@lgoldstein @tomaswolf Is there anything you need to consider this change for inclusion?

tomaswolf:
I just needed time to download this PR and run the tests since the CI tests are broken. (That's issue SSHD-1256.)

Squashed and merged as commit dc229b9f.

PR #217, opened 2022-04-15
[SSHD-966] Flush pending packets asynchronously at end of KEX

tomaswolf:
This PR is a series of four commits:

* Avoid synchronized methods in `ChannelAsyncOutputStream`. Do not hold locks while writing. This avoids lock inversions as in SSHD-966, and should also solve SSHD-1261, though gracefully closing such a stream may still be problematic if it is waiting on window space.
* Make `ChannelOutputStream.flush()` not synchronized. This enables the framework to asynchronously close such a stream. More fine-grained locking on the buffer to be written is used, and it is never locked when a packet is written or when the stream waits for window space. This avoids the lock inversion from SSHD-966.
* Fix a race condition in KEX as pointed out in SSHD-966, and flush the queue of pending packets asynchronously at the end of KEX. This enables the framework to handle incoming messages while flushing.
* Fix KEX for the SSH_MSG_SERVICE_REQUEST and SSH_MSG_SERVICE_ACCEPT messages, which have low message numbers but must not be sent during KEX, but which one must be able to handle in KEX state INIT. Additionally, start flushing the packet queue in KEX state KEYS as soon as the encryption for outgoing message is set (after the SSH_MSG_NEW_KEYS and a possible KEX extensions message have been written).

The code may look complicated, but the complexity for the fine-grained locking in the streams is local to these two classes, the complexity of the packet queue during KEX is localized in a new class `KeyExchangeMessagHandler`, and there is a high-level description included in the documentation with state and interaction diagrams.

Locally, this has passed the Apache MINA test suite repeatedly at each commit, and the JGit SSH tests have also passed repeatedly. (But the JGit tests do not really test for large data exchanges or KEX.)


tomaswolf:
@lgoldstein, @gnodet: this is my take on SSHD-966. The change factors out into a separate per-session object `KeyExchangeMessageHandler` all the logic about which packets to enqueue during KEX and how to flush them. Locally, I have green test runs, with a few flaky tests when Netty is used, but (a) I always have them, also without this change, and (b) the failures occur on connect() and are unrelated to KEX.

This resolves the race condition in `AbstractSession.writePacket()` I mentioned in SSHD-966, and avoids the lock inversion that could occur between the `ChannelOutputStream` and the flushing if some error occurred an the session was closed while flushing.

lgoldstein:
@tomaswolf The code is very complex and I am having a hard time following it - this is in no way criticism on it since I am aware of the complexity of the issue and the initial code state so I can truly appreciate the effort. I have reviewed it only from "30,000 feet" so to speak and nothing special stands out - however, I do admit I was not able to run through its logic with a fine toothcomb. My point is that I don't feel I can do justice to it so I trust you have reviewed it, thought about and tested it **thoroughly** since it touches on a most sensitive and crucial piece of the entire project. I appreciate the effort that must have gone into it.

tomaswolf:
I know it's non-trivial. That's why there are so many inline comments, and a lengthy commit message.

I've tested it as good as I can. But so far I've failed to write a test that would reproduce either the original lock inversion or the race (when run without this change).

I went through a number of false starts when developing this, which I discovered when tests failed. Luckily we do have a fairly comprehensive test suite, and it does some fairly unusual things in a number of places, too. If our test suite succeeds I'm fairly confident that a change is correct.

The overall state machine of KEX is not that complicated. In our code it's a bit hard to follow because it's intermingled with other code. I'll see if I can come up with some diagram that might help to understand this.

lgoldstein:
> The overall state machine of KEX is not that complicated. In our code it's a bit hard to follow because it's intermingled with other code. I'll see if I can come up with some diagram that might help to understand this.

Maybe we can make it part of the README file(s) of our documentation somehow - just a suggestion...

tomaswolf:
I changed this to "draft". I think it's the going the right way, but it still has problems. Looks like I really need to get rid of the synchronized methods in `ChannelOutputStream` and `ChannelAsyncOutputStream`.

lgoldstein:
> Looks like I really need to get rid of the synchronized methods in ChannelOutputStream and ChannelAsyncOutputStream.

This should be done **carefully** - I believe the synchronization is in place for good reason. BTW, I dimly remember that I encountered the same issue and when I tried to get rid of the synchronization (or at least limit its scope) I got into trouble. Keeping my fingers crossed you can come up with something...

BTW, at some point it may be worth re-evaluating the price in effort and code complexity vs the benefits - this is what got me to abort the attempt originally - the code was becoming way too convoluted and the effort of maintaining in future releases seemed to me to be too prohibitive.

tomaswolf:
> This should be done carefully

Of course. I'm not being reckless.

As for complexity: yes, the current code is convoluted and complex, and uses too many locks with too broad scopes. Reducing synchronization on locks with broad scope, or possibly even eliminating some of these altogether simplifies the overall structure of the code and will make it easier to understand, even if it may _locally_ increase perceived complexity.
Factoring out certain things into dedicated classes/objects may actually help reducing complexity or at least keeping it local to isolated classes and making the code easier to understand overall.

The current implementation has bugs that cause deadlocks in production servers. I know we don't provide any guarantees that our open-source software actually works or was fit for any purpose, but personally I don't like maintaining software that I know has such bugs. As long as it's fun trying to solve such problems and as long as I do have the time, I'm happy to try.

lgoldstein:
> As long as it's fun trying to solve such problems and as long as I do have the time, I'm happy to try

Go for it - that's the spirit

docs/kex.md
```diff
@@ -0,0 +1,131 @@
...
+messages:
+
+* `requestNewKeyExchange`: switches from `DONE` to `INIT` and calls `sendKexInit` to send the SSH_MSG_KEX_INIT message.
+* `handleKexInit`: receives the peer's SSH_MSG_KEX_INIT message; switches to `RUN`.
+* `sendNewKeys`: switches from `RUN` to `KEYS` and sends this side's SSH_MS_NEW_KEYS message.
+* `handleNewKeys`: receives the peer's SSH_MSG_KEX_INIT message; switches from `KEYS` to `DONE`.
+
+![KEX interactions](./kex_interaction_1.png)
+
+Because both sides my request a new key exchange at any time, it's also possible
```
lgoldstein, line 63: Typo `may request`

docs/kex.md
```diff
@@ -0,0 +1,131 @@
...
+
+In Apache MINA sshd, there are methods in `AbstractSession` for each of the KEX
+messages:
+
+* `requestNewKeyExchange`: switches from `DONE` to `INIT` and calls `sendKexInit` to send the SSH_MSG_KEX_INIT message.
+* `handleKexInit`: receives the peer's SSH_MSG_KEX_INIT message; switches to `RUN`.
+* `sendNewKeys`: switches from `RUN` to `KEYS` and sends this side's SSH_MS_NEW_KEYS message.
+* `handleNewKeys`: receives the peer's SSH_MSG_KEX_INIT message; switches from `KEYS` to `DONE`.
+
+![KEX interactions](./kex_interaction_1.png)
```
lgoldstein, line 61: I was wondering if we could use vector graphics (SVGs) instead of binary PNG(s) (see *inkscape* for such a freeware tool). As a rule of thumb I believe it is not recomended to store binary files in GIT.

docs/kex.md
```diff
@@ -0,0 +1,131 @@
+## SSH Key Exchange
```
lgoldstein, line 1: This is a very good idea to document this - I always thought about it, but never got around to it. Great initiative

docs/kex.md
tomaswolf: I can also use SVG. Won't make it much easier to change the diagrams, though. Inkscape has some problems with the SVG exported by the tool I used.

Binary files in git are totally fine. What is problematic are large files.

But I agree, let's use SVG for diagrams, and I'll also put the original source somewhere.

docs/kex.md
tomaswolf: Thanks; will be fixed in the next update of this PR.

tomaswolf:
OK, this is ready for review again. The CI build failures are SSHD-1256; that one test fails regularly. I don't know why, and I cannot reproduce locally, neither on OS X nor on Windows 10.

The four commits have commit messages explaining what is done; and the PR message at the top of this page has been updated, too.

docs/kex.md
tomaswolf: It was your idea. :-)

docs/kex.md
lgoldstein: Great - one more reason why I think SVGs are better is that they are basically text, so when one changes them GIT stores only the differences which are easier to review.

As far as inkspace is concerned - I was just suggesting it since it seems to be a popular open-source free tool (which I never used ....) but whatever SVG tool you prefer is just fine...

sshd-core/src/main/java/org/apache/sshd/client/session/ClientUserAuthService.java
```diff
@@ -70,6 +70,10 @@ public class ClientUserAuthService extends AbstractCloseable implements Service,
     private UserAuth userAuth;
     private int currentMethod;
 
+    private Object initLock = new Object();
```
lgoldstein, line 73: Recommend making *initLock* `final`

sshd-core/src/main/java/org/apache/sshd/client/session/ClientUserAuthService.java
```diff
@@ -124,7 +128,15 @@ public Map<String, Object> getProperties() {
 
     @Override
     public void start() {
-        // ignored
+        Runnable initial = null;
```
lgoldstein, line 131: No need to initialize *initial* to *null* since its value is set unconditionally inside the *synchronized* block.

sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelAsyncOutputStream.java
```diff
@@ -90,52 +93,174 @@ public byte getCommandType() {
...
+                }
             }
+            super.doCloseImmediately();
+        } finally {
+            shutdown();
         }
+    }
 
-        super.preClose();
+    private void shutdown() {
```
lgoldstein, line 158: Let's make this *protected* in the spirit of allowing users to overload anything they want in our code.

sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelAsyncOutputStream.java
```diff
@@ -90,52 +93,174 @@ public byte getCommandType() {
...
+            current = writeState.pendingWrite;
+            writeState.pendingWrite = null;
+            writeState.waitingOnIo = false;
+        }
+        if (current != null) {
+            terminateFuture(current);
+        }
+    }
+
+    private void terminateFuture(IoWriteFutureImpl future) {
```
lgoldstein, line 171: Let's make this *protected* in the spirit of allowing users to overload anything they want in our code.

sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelAsyncOutputStream.java
```diff
@@ -282,4 +435,43 @@ public boolean isSendChunkIfRemoteWindowIsSmallerThanPacketSize() {
     public void setSendChunkIfRemoteWindowIsSmallerThanPacketSize(boolean sendChunkIfRemoteWindowIsSmallerThanPacketSize) {
         this.sendChunkIfRemoteWindowIsSmallerThanPacketSize = sendChunkIfRemoteWindowIsSmallerThanPacketSize;
     }
+
+    // Marker type to avoid repeated buffering
+    private static class BufferedFuture extends IoWriteFutureImpl {
```
lgoldstein, line 440: Let's make this *protected* (as well as its constructor and methods) in the spirit of allowing users to overload anything they want in our code.

sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelAsyncOutputStream.java
```diff
@@ -282,4 +435,43 @@ public boolean isSendChunkIfRemoteWindowIsSmallerThanPacketSize() {
...
+
+        BufferedFuture(Object id, Buffer buffer) {
+            super(id, buffer);
+        }
+    }
+
+    // Collects state variables; access is always synchronized on the single instance per stream.
+    private static class WriteState {
```
lgoldstein, line 448: Let's make this *protected* (as well as its constructor, fields and methods) in the spirit of allowing users to overload anything they want in our code.

sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelOutputStream.java
```diff
@@ -42,6 +43,19 @@
  * @author <a href="mailto:dev@mina.apache.org">Apache MINA SSHD Project</a>
  */
 public class ChannelOutputStream extends OutputStream implements java.nio.channels.Channel, ChannelHolder {
+
+    private enum WriteState {
```
lgoldstein, line 47: Let's make this *protected* (or even *public*) in the spirit of allowing users to overload anything they want in our code.

sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelOutputStream.java
```diff
@@ -42,6 +43,19 @@
...
+        BUFFERED,
+        NEED_FLUSH,
+        NEED_SPACE
+    }
+
+    private enum OpenState {
```
lgoldstein, line 53: Let's make this *protected* (or even *public*) in the spirit of allowing users to overload anything they want in our code.

sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelOutputStream.java
```diff
@@ -50,12 +64,15 @@ public class ChannelOutputStream extends OutputStream implements java.nio.channe
...
+    private final AtomicReference<OpenState> openState = new AtomicReference<>(OpenState.OPEN);
+
+    private final Object bufferLock = new Object();
     private Buffer buffer;
     private int bufferLength;
     private int lastSize;
-    private boolean noDelay;
+    private boolean isFlushing;
+
+    private volatile boolean noDelay;
```
lgoldstein, line 75: Personally I am not a big fan of *volatile* as means  of multi-threading safety - I prefer *AtomicBoolean*-s - but I defer to your judgement here if you prefer *volatile*

sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelOutputStream.java
```diff
@@ -171,43 +227,52 @@ public synchronized void write(byte[] buf, int s, int l) throws IOException {
...
                                         .initCause(e);
                     }
-                }
-                session.resetIdleTimeout();
-                continue;
+                    session.resetIdleTimeout();
+                    break;
+                default:
+                    // BUFFERED implies l == 0; outer loop will terminate
+                    break;
```
lgoldstein, line 234: No need for *break* in *default* clause.

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
```diff
@@ -712,6 +762,17 @@ protected boolean handleServiceRequest(String serviceName, Buffer buffer) throws
         return true;
     }
 
+    private boolean validateServiceKexState(KexState state) {
```
lgoldstein, line 765: Let's make this *protected* in the spirit of letting our users override whatever they like.

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/KeyExchangeMessageHandler.java
```diff
@@ -0,0 +1,485 @@
...
+    // This is achieved by giving the flushing thread priority over the threads that might enqueue additional packets
+    // and flushing at least two packets at a time. Additionally the flush loop releases and shortly afterwards
+    // re-acquires the write lock, so normally not many readers (i.e., writePacket() calls) will get a chance to enqueue
+    // new packets.
+
+    /**
+     * We need the flushing thread to have priority over writing threads. So we use a lock that favors writers over
+     * readers, and any state updates and the flushing thread are writers, while writePacket() is a reader.
+     */
+    private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock(false);
```
lgoldstein, line 78: Let's make the members *protected* in the spirit of letting our users override whatever they like - e,g, provide their own derived handler.

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/KeyExchangeMessageHandler.java
```diff
@@ -0,0 +1,485 @@
...
+                            "Failed (" + e.getClass().getSimpleName() + ")" + " to check re-key necessity: " + e.getMessage()),
+                            e);
+                } catch (Exception e) {
+                    ExceptionUtils.rethrowAsIoException(e);
+                }
+            }
+        }
+    }
+
+    private IoWriteFuture writeOrEnqueue(int cmd, Buffer buffer, long timeout, TimeUnit unit) throws IOException {
```
lgoldstein, line 237: Let's make this *protected* in the spirit of letting our users override whatever they like.

lgoldstein (changes requested):
I admit I have not followed the code logic (despite the excellent documentation). I trust your skills and also our extensive test suite. There is one request I would like to make - there are quite a few *synchronized* blocks - please go over them again and try to think of the "worse" scenario so we can ensure:

* No deadlocks
* No variables are accessed without holding the proper lock

Like I  said, I was not able (lack of time...) to follow all the code logic, and since this is a major re-write of a core functionality, let's try to review it as much as possible before merging it to the master branch (I will try to find some more time to review, but I am currently swamped, so if you feel confident enough I trust your judgment to go ahead and merge it even if I have not been able to review it thoroughly so as not to delay this important change too long)

tomaswolf:
There is no hurry. I'll see if the Gerrit team can run their test suite with this.

sshd-core/src/main/java/org/apache/sshd/client/session/ClientUserAuthService.java
tomaswolf: Done.

sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelAsyncOutputStream.java
tomaswolf: Done.

sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelAsyncOutputStream.java
tomaswolf: Done.

sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelAsyncOutputStream.java
tomaswolf: Done.

sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelOutputStream.java
tomaswolf: Made protected.

sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelAsyncOutputStream.java
tomaswolf: Done.

sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelOutputStream.java
tomaswolf: Done.

sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelOutputStream.java
tomaswolf: Done.

sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelOutputStream.java
tomaswolf: Not done. I prefer to have *every* switch branch be terminated by a break, return, or throw. (I can't remember the last time I had the urge to use a fall-though.) It may not be necessary here, but it doesn't hurt.

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
tomaswolf: Done.

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/KeyExchangeMessageHandler.java
tomaswolf: Done.

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/KeyExchangeMessageHandler.java
tomaswolf: Done.

sshd-core/src/main/java/org/apache/sshd/client/session/ClientUserAuthService.java
tomaswolf: Done.

tomaswolf:
> I'll see if the Gerrit team can run their test suite with this.

https://bugs.chromium.org/p/gerrit/issues/detail?id=12758#c8

PR #218, opened 2022-04-16
[SSHD-1261] Fixed possible pending unfinished pending write request if ChannelAsyncOutputStream closed prematurely

lgoldstein:


sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelAsyncOutputStream.java
```diff
@@ -127,6 +129,28 @@ protected CloseFuture doCloseGracefully() {
...
+        if ((future != null) && (!future.isDone())) {
+            Exception error = errorProvider.get();
+            log.debug("abortCurrentWrite({}) aborting pending write={} - {} : {}",
+                    this, future, error == null ? null : error.getClass().getSimpleName(),
+                    error == null ? null : error.getMessage());
+            future.setValue(error);
+        }
+
+        return future;
+    }
```
tomaswolf, line 152: Terminating the current future is probably right, but still I don't think this will help in all situations. On a graceful shutdown, `AbstractCloseable.close` will call `doCloseGracefully()` *first*, which will block until the pending future is fulfilled.

However, it also calls `preClose()` first, which will already close the `packetWriter`.

I'm not sure a `ChannelAsyncOutputStream` *can* be closed gracefully in this situation.

lgoldstein:
> Terminating the current future is probably right, but still I don't think this will help in all situations

I gues you are right - let's plug this "hole" and then think about the rest

tomaswolf:
This is obsolete; should be fixed with the changes made in commit 1ab50fe2.

lgoldstein:
Agreed - discarding it...

PR #227, opened 2022-06-12
SSHD-1055: Half-close port-forwarded socket on EOF in tunnel

tomaswolf:
Propagate an EOF on a tunnel for port forwarding to the socket by
shutting down its output; otherwise the client connected to that
socket may hang.

For the MINA transport back-end, calling socket.shutdownOutput(), which
is possible only via reflection, leads to problems as it confuses the
main polling loop inside MINA. So don't do that but schedule the MINA
session to close once all pending writes have been written.

Enable half-closure in the Netty transport back-end, and use proper Netty
API instead of Java reflection to shut down the output on the socket.

Fix the Nio2 transport back-end to not re-try accepting connections
when the channel is already closing.

See also SSHD-964 and SSHD-902.


sshd-core/pom.xml
```diff
@@ -110,6 +110,24 @@
             <artifactId>testcontainers</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-netty-shaded</artifactId>
+            <version>1.27.2</version>
```
lgoldstein, line 116: Let's use a property for the grpc version (D.R.Y.)

tomaswolf: Done.

tomaswolf:
Some tests appear to run much slower now. In particular the sshd-mina and sshd-sftp tests. I didn't notice any such slowdown when building locally. Perhaps a fluke of the GitHub CI? But surely something that needs to be investigated.

tomaswolf:
Looks indeed like something is wrong with MINA now. It's in particular the sftp tests when run with MINA.

PR #230, opened 2022-07-01
[SSHD-1273] Add support to use env vars together with subsystem

andreid911:


andreid911:
@tomaswolf  @lgoldstein could you please take a look.

sshd-core/src/main/java/org/apache/sshd/client/channel/PtyCapableChannelSession.java
```diff
@@ -200,21 +200,6 @@ public void setPtyModes(Map<PtyMode, Integer> ptyModes) {
...
-     */
-    public Object setEnv(String key, Object value) {
-        ValidateUtils.checkNotNull(key, "No key provided");
-        if (value == null) {
-            return env.remove(key);
-        } else {
-            return env.put(key, value);
-        }
-    }
-
```
tomaswolf, line 217: Also remove the map in line 88.

sshd-core/src/main/java/org/apache/sshd/client/channel/ChannelSession.java
```diff
@@ -217,4 +223,42 @@ protected int securedRead(
...
     }
+
+    /**
+     * @param  key   The (never {@code null}) key (Note: may be empty...)
+     * @param  value The value to set - if {@code null} then the pre-existing value for the key (if any) is
+     *               <U>removed</U>.
+     * @return       The replaced/removed previous value - {@code null} if no previous value set for the key.
+     */
+    public Object setEnv(String key, Object value) {
+        ValidateUtils.checkNotNull(key, "No key provided");
```
lgoldstein, line 234: Please use `ValidateUtils#checkNotNullAndNotEmpty(key)`

sshd-core/src/main/java/org/apache/sshd/client/channel/ChannelSession.java
```diff
@@ -48,6 +53,7 @@ public class ChannelSession extends AbstractClientChannel {
 
     private CloseableExecutorService pumperService;
     private Future<?> pumper;
+    private Map<String, Object> env = new LinkedHashMap<>();
```
lgoldstein, line 56: Is there any special reason why you are using a `LinkedHashMap` rather than a `HashMap` ? Does the order of the keys matter in any way ?

lgoldstein, line 56: Should be `final` - unless there is some reason to make it mutabel (?)

sshd-core/src/main/java/org/apache/sshd/client/channel/ChannelSession.java
```diff
@@ -217,4 +223,42 @@ protected int securedRead(
...
+            // Cannot use forEach because of the IOException being thrown by writePacket
+            for (Map.Entry<String, ?> entry : env.entrySet()) {
+                String key = entry.getKey();
+                Object value = entry.getValue();
+                String str = Objects.toString(value);
+                Buffer buffer = session.createBuffer(
+                        SshConstants.SSH_MSG_CHANNEL_REQUEST, key.length() + GenericUtils.length(str) + Integer.SIZE);
+                buffer.putInt(getRecipient());
+                buffer.putString("env");
+                buffer.putBoolean(false); // want-reply
```
lgoldstein, line 257: The standard does not say that *wantReply* must be *false* - while I do agree that it is a sensible default have you given thought how yo handle *true* ? At the very least this choice should be documented and perhaps open a follow-up issue to add support for *wantReply=true* in some future release

sshd-core/src/main/java/org/apache/sshd/client/channel/PtyCapableChannelSession.java
```diff
@@ -200,21 +199,6 @@ public void setPtyModes(Map<PtyMode, Integer> ptyModes) {
...
     }
 
-    /**
-     * @param  key   The (never {@code null}) key (Note: may be empty...)
-     * @param  value The value to set - if {@code null} then the pre-existing value for the key (if any) is
-     *               <U>removed</U>.
-     * @return       The replaced/removed previous value - {@code null} if no previous value set for the key.
-     */
-    public Object setEnv(String key, Object value) {
-        ValidateUtils.checkNotNull(key, "No key provided");
```
lgoldstein, line 210: Please use `ValidateUtils#checkNotNullAndNotEmpty(key)`

sshd-core/src/main/java/org/apache/sshd/client/channel/PtyCapableChannelSession.java
```diff
@@ -289,25 +273,6 @@ protected void doOpenPty() throws IOException {
...
-            // Cannot use forEach because of the IOException being thrown by writePacket
-            for (Map.Entry<String, ?> entry : env.entrySet()) {
-                String key = entry.getKey();
-                Object value = entry.getValue();
-                String str = Objects.toString(value);
-                Buffer buffer = session.createBuffer(
-                        SshConstants.SSH_MSG_CHANNEL_REQUEST, key.length() + GenericUtils.length(str) + Integer.SIZE);
-                buffer.putInt(getRecipient());
-                buffer.putString("env");
-                buffer.putBoolean(false); // want-reply
```
lgoldstein, line 306: See previous comment regarding *wantReply=false* as hardwired value

lgoldstein (changes requested):
Looks fine - a few minor comments

sshd-core/src/main/java/org/apache/sshd/client/channel/ChannelSession.java
andreid911: agree, no reasons to leave it mutable. Fixed

sshd-core/src/main/java/org/apache/sshd/client/channel/PtyCapableChannelSession.java
andreid911: it is just deleted code :)

andreid911:
@lgoldstein fixed defects from comments and squashed to 1 commit.

tomaswolf (commented):
Looks good to me. Setting environment variables has got nothing to do with terminals. Can be merged once the CI builds succeed.

Note that this change theoretically also enables setting environment variables on a `SftpChannelSubsystem` (but it can't be used there in the default implementation because the `DefaultSftpClient` creates the channel and then opens it right away).

sshd-core/src/main/java/org/apache/sshd/client/channel/ChannelSession.java
andreid911: According to the current javadoc: key can be empty string. I am not planning to make any constraints changes to avoid behavior changes. 
@lgoldstein do you think we can change javadoc and update validation or use old validation without empty check?

andreid911: @tomaswolf Agree. Changes were reverted.

lgoldstein:
[RFC 4254 - section 6.4](https://datatracker.ietf.org/doc/html/rfc4254#section-6.4) mentions shell and/or command but not subsystem. However, I see no harm in allowing this for subsystems as well. IMO it is the **user's** responsibility to make sure that the peer handles environment settings correctly. Our code simply updates a `StandardEnvironment` instance.

PR #231, opened 2022-07-07
[SSHD-1276] Added capability to redirect command/shell STDERR stream to STDOUT one

lgoldstein:
@tomaswolf I would appreciate your feedback...


tomaswolf:
What about ```Streaming.Async```?

lgoldstein:
You are correct - fixed it...

PR #233, opened 2022-07-14
Fix container tests when run from the src tgz in the distribution

tomaswolf:
Running a "mvn clean install" from the source tar archive failed for
some container tests because unpacking the tar might not preserve
executable bits. Testcontainers need an entrypoint that is executable,
or otherwise the entrypoint script must not be run directly but via
a shell explicitly.

Rewrite the two problematic tests to ensure the entrypoint script is
always executable, irrespective of whether the test resource has the
bit set.


tomaswolf:
@gnodet, @elecharny : this should fix the problem noticed in the 2.9.0 release candidate [on the dev mailing list](https://www.mail-archive.com/dev@mina.apache.org/msg38601.html). At least for `ClientOpenSSHCertificatesTest.java` that problem should have existed already in 2.8.0.

PR #235, opened 2022-07-29
[SSHD-1285] Fix runtime linkage errors with Java 8 when building with…

norrisjeremy:
… Java versions > 8

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -263,6 +263,24 @@
         </profile>
 
         <profile>
+            <id>javac-jdk9+</id>
+            <activation>
+                <jdk>[9,)</jdk>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


elecharny:
Hi,

I think this can also be done with just a property, like what we have in MINA:

```
  <!-- use JDK8 compilation for JDK9+ compiler -->
  <profile>
    <id>java-8-compilation</id>
    <activation>
      <jdk>[9,)</jdk>
    </activation>
    <properties>
      <maven.compiler.release>8</maven.compiler.release>
    </properties>
  </profile>
```

norrisjeremy:
The `pom.xml` for this project already has the following, so it seems more consistent to do it the way I proposed:
```
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-compiler-plugin</artifactId>
                <version>3.8.1</version>
                <configuration>
                    <source>${javac.source}</source>
                    <target>${javac.target}</target>
...
```

PR #237, opened 2022-08-03
[SSHD-1283] Added configuration property to control whether ScpShell is enabled

lgoldstein:


lgoldstein:
@tomaswolf Could you take a look ? Thx

sshd-common/src/main/java/org/apache/sshd/common/config/ConfigFileReaderSupport.java
```diff
@@ -64,7 +68,26 @@ public final class ConfigFileReaderSupport {
     public static final String DEFAULT_KBD_INTERACTIVE_AUTH = "yes";
     public static final boolean DEFAULT_KBD_INTERACTIVE_AUTH_VALUE = parseBooleanValue(DEFAULT_KBD_INTERACTIVE_AUTH);
 
+    public static final String HOSTBASED_AUTH_CONFIG_PROTP = "HostbasedAuthentication";
+    public static final String DEFAULT_HOSTBASED_AUTH = "no";
+    public static final boolean DEFAULT_HOSTBASED_AUTH_VALUE = parseBooleanValue(DEFAULT_HOSTBASED_AUTH);
+
     public static final String PREFERRED_AUTHS_CONFIG_PROP = "PreferredAuthentications";
+    // NOTE: entries are ordered in the same order as the default setting for the client
+    public static final List<PreferredAuthConfigEntry> PREFERRED_AUTHS_CONFIG_ENTRIES = Collections
```
tomaswolf, line 77: Perhaps I'm missing something (can only review this in the Github Web UI right now), but I don't see any use of this list.

sshd-common/src/main/java/org/apache/sshd/common/config/PreferredAuthConfigEntry.java
```diff
@@ -0,0 +1,92 @@
...
+    private final String strValue;
+
+    public PreferredAuthConfigEntry(
+                                    String configKeyName, String authFactoryName, boolean enabledByDefault) {
+        this.configKeyName = ValidateUtils.checkNotNullAndNotEmpty(configKeyName, "No configuration key name provided");
+        this.authFactoryName
+                = ValidateUtils.checkNotNullAndNotEmpty(authFactoryName, "No authentication factory name provided");
+        this.enabledByDefault = enabledByDefault;
+        this.hashValue = Objects.hash(configKeyName, authFactoryName, enabledByDefault);
+        this.strValue
```
tomaswolf, line 43: Does it really make sense to pre-compute this toString() value? Normally I would expect toString() to be called only for debugging output.

tomaswolf (commented):
I don't see any connection between commit 4e2f817c and SSHD-1281. Please fix that typo in the commit message.

tomaswolf:
BTW: I think we should have a new release 2.9.1 with only the fixes for Java 8 (SSHD-1281/SSHD-1285), with anything else  in a later release. In particular new features should not be in 2.9.x but in 2.10.0.

sshd-common/src/main/java/org/apache/sshd/common/config/PreferredAuthConfigEntry.java
lgoldstein: True, but since all members are *final* I think it makes sense

lgoldstein:
>> BTW: I think we should have a new release 2.9.1 with only the fixes for Java 8 (SSHD-1281/SSHD-1285), with anything else in a later release. In particular new features should not be in 2.9.x but in 2.10.0.

Agreed

lgoldstein:
>> Perhaps I'm missing something (can only review this in the Github Web UI right now), but I don't see any use of this list.

No, you are not missing anything - I am still thinking about how to use it - if I cannot make up my mind I will delete it...

lgoldstein:
>> This doesn't appear to be a non-standard option. Normal OpenSSH allows overriding any config setting via "-o" on the command-line, and PreferredAuthentications is a standard SSH config key.

You are right - will re-phrase or move it to the correct section.

lgoldstein:
>> I don't see any connection between commit https://github.com/apache/mina-sshd/commit/4e2f817c1eff380d083cf43df5d1aec6d8830564 and SSHD-1281. Please fix that typo in the commit message.

Will do

sshd-common/src/main/java/org/apache/sshd/common/config/PreferredAuthConfigEntry.java
tomaswolf: It wastes space, even if toString() is never called.

lgoldstein: I'll consider it

PR #241, opened 2022-08-31
[SSHD-1293] Fix unbinding port forwarding for auto-alloc port

knalli:
This fixes de-allocating / unbinding a local port forwarding binding when using a dynamic auto-allocated port.

The information "localAddress" is controlled by the user and may contain still a port `0` which means auto-allocating a local port; the actual local port is available in 'boundAddress". However, "DefaultForwarder" tracks
the actual port mappings only.

I have run the tests locally with a Java 8 without issues.

Related issue https://issues.apache.org/jira/browse/SSHD-1293

I also added an additional commit modifying the test covering the missing signal. If you are fine with this, I'd squash them. Or maybe as an additional test?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -22,6 +22,7 @@
 
 ## Bug fixes
 
+* [SSHD-1293](https://issues.apache.org/jira/browse/SSHD-1293) ExplicitPortForwardingTracker does not unbind auto-allocated port
 * [SSHD-1294](https://issues.apache.org/jira/browse/SSHD-1294) Close MinaServiceFactory instances properly
 
 ## Major code re-factoring
--- a/sshd-core/src/main/java/org/apache/sshd/client/session/forward/ExplicitPortForwardingTracker.java
+++ b/sshd-core/src/main/java/org/apache/sshd/client/session/forward/ExplicitPortForwardingTracker.java
@@ -56,7 +56,7 @@ public class ExplicitPortForwardingTracker extends PortForwardingTracker impleme
         if (open.getAndSet(false)) {
             PortForwardingManager manager = getClientSession();
             if (isLocalForwarding()) {
-                manager.stopLocalPortForwarding(getLocalAddress());
+                manager.stopLocalPortForwarding(getBoundAddress());
             } else {
                 manager.stopRemotePortForwarding(getRemoteAddress());
             }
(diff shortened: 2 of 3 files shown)


knalli:
Note: The existing test `testLocalForwardingNative` isn't covering this (anymore) because the semaphore `tearDownSignal` is checked in the events (like `tornDownExplicitTunnel`) only -- and these events aren't called anymore! This is maybe a bigger issue actually and the root cause is the combined try-with-resources like

```
        try (ClientSession session = createNativeSession(listener);
             ExplicitPortForwardingTracker tracker = session.createLocalPortForwardingTracker(new SshdSocketAddress("", 0),
                     new SshdSocketAddress(TEST_LOCALHOST, echoPort))) {
```

Closing the session will "destroy" everything, but there is no check if the destroy has happen after closing the tracker only.

tomaswolf:
Looks good. Please rebase and squash the commits, then we can merge this.

knalli:
Oops.. I missed your comment. Squash & rebase done.

tomaswolf:
> Oops.. I missed your comment. Squash & rebase done.

Guess you didn't notice that I had already done so. I just hadn't merged yet.

knalli:
Oh, alright 🫠 

I've seen there are some changes in the forwarding topic in the meanwhile. Is it worth to check if there are other leaking situations maybe there? Or at least checking the tests cover the actual invocations (as noted in my 2nd comment here)?

tomaswolf:
> I've seen there are some changes in the forwarding topic in the meanwhile. Is it worth to check if there are other leaking situations maybe there? Or at least checking the tests cover the actual invocations (as noted in my 2nd comment here)?

Sure. There were some strange failures still in port forwarding in the Github CI builds (not caused by your change). I cannot reproduce yet locally, but it seems there are still a few problems with port forwarding. The [failures I saw](https://github.com/apache/mina-sshd/actions/runs/3078358458/jobs/4973967007) are probably timing-related; the Github builds are much slower than on my local machine.

There's also [this sporadic failure](https://github.com/apache/mina-sshd/actions/runs/3078358458/jobs/4973882256), which relates to SSHD-1256. That failure appears to reoccur only with the MINA transport; don't know yet why unbinding doesn't work or is slow sometimes.

Any help in tracking down the causes of these failures would be greatly appreciated!

PR #242, opened 2022-09-01
Fix: close session when timeout during connect

onyas:
Hi there, 

We have encouraged an issue in which the SSHD doesn’t close the session when connecting timeout. 

The code we are using is attached here. https://gist.github.com/onyas/fda8ccf2d373c743ccec8f648782d078  
The TCP socket is still ESTABLISHED after `sshClient.connect()` threw an exception. 
image

If I set the timeout to 1 millisecond, the `sshClient.connnect()` will throw an exception, since it’s timeout so the session is still null, and it doesn't help if we call `session.close()`. 

And we know we cloud call `sshClient.stop()`, but in our scenario, we reuse the same instance so we couldn’t do that.

After checking the source code, we find the SSHD didn’t close the session when timeout. So I created this pull request to help close the session. 

Please help review and thanks.

### DIFF
--- a/sshd-common/src/main/java/org/apache/sshd/common/future/DefaultSshFuture.java
+++ b/sshd-common/src/main/java/org/apache/sshd/common/future/DefaultSshFuture.java
@@ -21,6 +21,7 @@ package org.apache.sshd.common.future;
 import java.io.InterruptedIOException;
 import java.lang.reflect.Array;
 import java.util.Objects;
+import java.util.concurrent.TimeoutException;
 
 import org.apache.sshd.common.util.GenericUtils;
 import org.apache.sshd.common.util.ValidateUtils;
@@ -38,6 +39,7 @@ public class DefaultSshFuture<T extends SshFuture> extends AbstractSshFuture<T> 
     private final Object lock;
     private Object listeners;
     private Object result;
+    private boolean timeout;
 
     /**
      * Creates a new instance.
@@ -69,6 +71,7 @@ public class DefaultSshFuture<T extends SshFuture> extends AbstractSshFuture<T> 
                     lock.wait(endTime - curTime);
                 } catch (InterruptedException e) {
                     if (interruptable) {
+                        setTimeout();
                         curTime = System.currentTimeMillis();
                         throw formatExceptionMessage(msg -> {
                             InterruptedIOException exc = new InterruptedIOException(msg);
@@ -79,6 +82,9 @@ public class DefaultSshFuture<T extends SshFuture> extends AbstractSshFuture<T> 
                 }
 
                 curTime = System.currentTimeMillis();
+                if (curTime >= endTime) {
+                    setTimeout();
+                }
                 if ((result != null) || (curTime >= endTime)) {
                     return result;
                 }
@@ -105,6 +111,27 @@ public class DefaultSshFuture<T extends SshFuture> extends AbstractSshFuture<T> 
                 return;
             }
 
+            result = (newValue != null) ? newValue : GenericUtils.NULL;
+            lock.notifyAll();
+        }
+
+        notifyListeners();
+    }
+
+    /**
+     * Like {@link #setValue(Object)}; but sets the value only if there has been no time-out on this future yet.
+     *
+     * @param  newValue         The operation result
+     * @throws TimeoutException If a {@link #isTimeout()} is {@code true}
+     */
+    public void setValueIfNotTimedOut(Object newValue) throws TimeoutException {
+        synchronized (lock) {
+            if (result != null) {
+                return;
+            }
+            if (isTimeout()) {
+                throw new TimeoutException();
+            }
             result = (newValue != null) ? newValue : GenericUtils.NULL;
             lock.notifyAll();
         }
@@ -219,12 +246,35 @@ public class DefaultSshFuture<T extends SshFuture> extends AbstractSshFuture<T> 
         }
     }
 
+    /**
+     * Tells whether this future was canceled.
+     *
+     * @return {@code true} if the future is canceled, {@code false} otherwise
+     */
     public boolean isCanceled() {
         return getValue() == CANCELED;
     }
 
+    /**
+     * Cancels this future.
+     */
     public void cancel() {
         setValue(CANCELED);
+    }
+
+    /**
+     * Tells whether a {@link #await0(long, boolean)} on this future timed out already.
+     *
+     * @return {@code true} if there was a time-out, {@code false} otherwise
+     */
+    public boolean isTimeout() {
+        synchronized (lock) {
+            return timeout;
+        }
+    }
+
+    private void setTimeout() {
+        this.timeout = true;
     }
 
     @Override
--- a/sshd-core/src/main/java/org/apache/sshd/client/SshClient.java
+++ b/sshd-core/src/main/java/org/apache/sshd/client/SshClient.java
@@ -38,6 +38,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
@@ -586,9 +587,15 @@ public class SshClient extends AbstractFactoryManager implements ClientFactoryMa
                                         if (f5.isConnected()) {
                                             ClientSession clientSession = f5.getClientSession();
                                             clientSession.setAttribute(TARGET_SERVER, address);
-                                            connectFuture.setSession(clientSession);
-                                            proxySession.addCloseFutureListener(f6 -> clientSession.close(true));
-                                            clientSession.addCloseFutureListener(f6 -> proxySession.close(true));
+                                            try {
+                                                connectFuture.setSessionIfNotTimeout(clientSession);
+                                                proxySession.addCloseFutureListener(f6 -> clientSession.close(true));
+                                                clientSession.addCloseFutureListener(f6 -> proxySession.close(true));
+                                            } catch (TimeoutException e) {
+                                                clientSession.close(true);
+                                                proxySession.close(true);
+                                                connectFuture.setException(e);
+                                            }
                                         } else {
                                             proxySession.close(true);
                                             connectFuture.setException(f5.getException());
@@ -750,8 +757,11 @@ public class SshClient extends AbstractFactoryManager implements ClientFactoryMa
                             ? KeyIdentityProvider.EMPTY_KEYS_PROVIDER
                             : identities);
         }
-
-        connectFuture.setSession(session);
+        try {
+            connectFuture.setSessionIfNotTimeout(session);
+        } catch (TimeoutException e) {
+            session.close();
+        }
     }
 
     protected void setupDefaultSessionIdentities(
--- a/sshd-core/src/main/java/org/apache/sshd/client/future/ConnectFuture.java
+++ b/sshd-core/src/main/java/org/apache/sshd/client/future/ConnectFuture.java
@@ -18,6 +18,8 @@
  */
 package org.apache.sshd.client.future;
 
+import java.util.concurrent.TimeoutException;
+
 import org.apache.sshd.client.session.ClientSession;
 import org.apache.sshd.client.session.ClientSessionHolder;
 import org.apache.sshd.common.future.SshFuture;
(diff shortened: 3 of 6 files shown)


tomaswolf:
As far as I see what happens is that the asynchronous connection attempt succeeds after the timeout.

I'd move the timeout flag and the `isTimeout()` method to `DefaultSshFuture`. I also think the flag must be set and read under lock, otherwise there can still be a race condition. And instead of
```
if (connectFuture.isTimeout()) {
  session.close();
} else {
  connectFuture.setSession(session);
}
```
there needs to be a mechanism to do this atomically. Something like
```
  /** Like setValue; but sets the value only if not timed out yet. Throws TimeoutException otherwise. */
  public void setValueIfNotTimedOut(Object value) throws TimeoutException {
    synchronized(lock) {
       if (result != null ) {
         return; 
       }
       if (isTimeout()) {
         throw new TimeoutException();
       }
       result = (newValue != null) ? newValue : GenericUtils.NULL;
       lock.notifyAll();
    }
  }
```
and then
```
  private void setSessionSafely(ConnectFuture connectFuture, ClientSession session) throws IOException {
    try {
      connectFuture.setValueIfNotTimedOut(session);
    } catch (TimeoutException e) {
      session.close();
    }
  }
```
Finally, it would be really cool if this could be done before the `ClientSession` gets created at all (by closing the IoSession immediately). Creating the client session will already start the SSH protocol. But the `ClientSession` is created in `AbstractSessionIoHandler.sessionCreated(IoSession)`, where we don't have access to the ConnectFuture, so I don't quite see how this could be done.

onyas:
Hi, @tomaswolf  Thank you for your insights, and I updated the code according to your suggestions. Please help review. 

Thanks.

sshd-common/src/main/java/org/apache/sshd/common/future/DefaultSshFuture.java
```diff
@@ -50,6 +52,7 @@ public DefaultSshFuture(Object id, Object lock) {
         super(id);
 
         this.lock = (lock != null) ? lock : this;
+        this.timeout = false;
```
tomaswolf, line 55: Unnecessary; it's the default value anyway.

sshd-common/src/main/java/org/apache/sshd/common/future/DefaultSshFuture.java
```diff
@@ -227,6 +251,13 @@ public void cancel() {
         setValue(CANCELED);
     }
 
+    private boolean isTimeout() {
+        return timeout;
+    }
```
tomaswolf, line 256: Missing empty line between methods. Also make `isTimeout()` public; it may be useful in other contexts. And it must access the flag under lock.

sshd-common/src/main/java/org/apache/sshd/common/future/DefaultSshFuture.java
```diff
@@ -79,6 +82,9 @@ protected Object await0(long timeoutMillis, boolean interruptable) throws Interr
                 }
 
                 curTime = System.currentTimeMillis();
+                if (curTime >= endTime) {
+                    setTimeout();
+                }
```
tomaswolf, line 87: I think we also need to call setTimeout() above, in the exception handler, inside `if (interruptable) {`.

sshd-core/src/main/java/org/apache/sshd/client/future/ConnectFuture.java
```diff
@@ -78,4 +80,13 @@ default ClientSession getClientSession() {
      * Cancels the connection attempt and notifies all threads waiting for this future.
      */
     void cancel();
+
+
+    /**
+     * Sets the newly connected session only if not timeout and notifies all threads waiting for this future. This method is invoked by SSHD
+     * internally. Please do not call this method directly.
+     *
+     * @param session The {@link ClientSession}
```
tomaswolf, line 89: Missing `@throws` javadoc

sshd-common/src/main/java/org/apache/sshd/common/future/DefaultSshFuture.java
```diff
@@ -112,6 +118,24 @@ public void setValue(Object newValue) {
         notifyListeners();
     }
 
+    /**
+     * Like setValue; but sets the value only if not timed out yet. Throws TimeoutException otherwise.
+     *
+     * @param newValue The operation result
```
tomaswolf, line 124: Missing `@throws` javadoc.

tomaswolf (changes requested):

onyas:
Hi, @tomaswolf Thank you for the suggestions, I updated the code, please help review.

Thanks.

sshd-common/src/main/java/org/apache/sshd/common/future/DefaultSshFuture.java
```diff
@@ -227,6 +252,16 @@ public void cancel() {
         setValue(CANCELED);
     }
 
+    public boolean isTimeout() {
```
tomaswolf, line 255: Missing javadoc.

sshd-common/src/main/java/org/apache/sshd/common/future/DefaultSshFuture.java
```diff
@@ -227,6 +252,16 @@ public void cancel() {
...
+        synchronized (lock){
```
tomaswolf, line 256: Nit: missing blank before {

sshd-common/src/main/java/org/apache/sshd/common/future/DefaultSshFuture.java
```diff
@@ -227,6 +252,16 @@ public void cancel() {
...
+            return timeout;
+        }
+    }
+
+    private void setTimeout(){
```
tomaswolf, line 261: Nit: missing blank before {

tomaswolf (commented):
LGTM. Can we write a test for this?

sshd-core/src/main/java/org/apache/sshd/client/SshClient.java
```diff
@@ -586,9 +587,14 @@ protected ConnectFuture doConnect(
                                         if (f5.isConnected()) {
                                             ClientSession clientSession = f5.getClientSession();
                                             clientSession.setAttribute(TARGET_SERVER, address);
-                                            connectFuture.setSession(clientSession);
-                                            proxySession.addCloseFutureListener(f6 -> clientSession.close(true));
-                                            clientSession.addCloseFutureListener(f6 -> proxySession.close(true));
+                                            try {
+                                                setSessionSafely(connectFuture, clientSession);
+                                                proxySession.addCloseFutureListener(f6 -> clientSession.close(true));
+                                                clientSession.addCloseFutureListener(f6 -> proxySession.close(true));
```
tomaswolf, line 593: This isn't correct yet. If there is a timeout, the proxySession also needs to be closed.

tomaswolf:
> Can we write a test for this?

I've created issue [SSHD-1295](https://issues.apache.org/jira/browse/SSHD-1295) in our bug tracker for this, and have posted a test case there. Please incorporate that test into this change.

onyas:
Hi @tomaswolf  All the suggestions are implemented, please take a look.

Thanks.

sshd-core/src/test/java/org/apache/sshd/client/ClientTest.java
```diff
@@ -1530,6 +1532,29 @@ public void testConnectUsingIPv6Address() throws IOException {
...
+                fail("Timeout expected");
+            } catch (InterruptedIOException | SshException e) {
+                ClientSession session = null;
+                try {
+                    session = future.verify(CONNECT_TIMEOUT).getSession();
+                } catch (SshException e2) {
+                    assertTrue("Expected a timeout, got " + e2, e2.getMessage().contains("timeout"));
+                }
+
+                assertNull("Session should not set since client timeout", session);
```
tomaswolf, line 1551: What is missing in this test now is a verification that the session, if one got created, was indeed closed.

Maybe install a SessionListener on the client before starting it. That sessionListener gets invoked whenever a ClientSession is created and can store that session. Then check here that all sessions that got created are closed or closing.

sshd-core/src/test/java/org/apache/sshd/client/ClientTest.java
```diff
@@ -1530,6 +1532,29 @@ public void testConnectUsingIPv6Address() throws IOException {
...
     }
 
+    @Test // see SSHD-1295
+    public void testConnectTimeout() throws Exception {
+        client.start();
+        try {
+            ConnectFuture future = client.connect(getCurrentTestName(), TEST_LOCALHOST, port);
+            try {
+                future.verify(1);
+                fail("Timeout expected");
```
tomaswolf, line 1542: Somehow this test now fails in the CI tests here. See https://github.com/apache/mina-sshd/runs/8283257340?check_suite_focus=true . Any idea why?

sshd-core/src/test/java/org/apache/sshd/client/ClientTest.java
tomaswolf: Add before line 1537
```
        List<Session> sessions = new CopyOnWriteArrayList<>();
        client.addSessionListener(new SessionListener() {

            @Override
            public void sessionCreated(Session session) {
                // Delay a little bit to ensure that verify(1) does time out below
                try {
                    Thread.sleep(100);
                } catch (InterruptedException e) {
                    Thread.currentThread().interrupt();
                }
                sessions.add(session);
            }
        });
```
and then after line 1551
```
                for (Session created : sessions) {
                    assertTrue("Created session should be closed", created.isClosed() || created.isClosing());
                }
```

sshd-core/src/test/java/org/apache/sshd/client/ClientTest.java
tomaswolf: See below.

sshd-core/src/test/java/org/apache/sshd/client/ClientTest.java
onyas: Thank you. By the way, why does it work locally but fail in GitHub Action? And need to add the delay operation.

tomaswolf: It's a timing issue. When I fetched your change and ran the test the first time, it worked locally, but I had debug logging on. When I switched off debug logging, it failed for me locally, too.

tomaswolf:
Looks good now. Is it OK for you if I squash this series and fix some of the javadoc comments, and then force-push onto this branch?

onyas:
Sure, please go ahead.

tomaswolf:
@lgoldstein and @gnodet: I'd appreciate your opinions on this. Please also see my thoughts at [SSHD-1295](https://issues.apache.org/jira/browse/SSHD-1295?focusedCommentId=17602860&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17602860).

onyas:
Hi, @tomaswolf  Good day, another question, when is this version 2.9.2 planned to be released?

tomaswolf:
"2.9.2-SNAPSHOT" is just the current development version. I don't think we'll do another patch release for this. So most likely the change will come whenever 2.10.0 is released, which may be in a few months.

tomaswolf:
Just a heads up: the basic idea is going in the right direction, but cancelling a `ConnectFuture` has the same problem. I've come to the conclusion that cancellation has to be sorted out first, and then this time-out case can be based on that. (I.e., client code would get the possibility to specify in the `verify()` call whether to cancel a future automatically on a time-out.) That gives client code full control.

I'm working on this.

lgoldstein:
>  client code would get the possibility to specify in the verify() call whether to cancel a future automatically on a time-out

Let's keep backward compatibility as much as possible with existing API - i.e., keep the existing *verify* without the flag and delegate it to call the new *verify* using some default that you deem relevant.

BTW, @tomaswolf you are doing great work on the project (I really appreciate it) - I wish I had the time I used to, so I could match your prolific work both in quality and quantity.

lgoldstein:
> The new default for the existing verify() would be to cancel to avoid this problem of having a "hidden" session established after the time-out that the user cannot really get to.

Seems reasonable - I trust your judgement on this.

onyas:
> "2.9.2-SNAPSHOT" is just the current development version. I don't think we'll do another patch release for this. So most likely the change will come whenever 2.10.0 is released, which may be in a few months.

Hi, @tomaswolf  do we have a release date for 2.10.0?

tomaswolf:
> do we have a release date for 2.10.0?

No.

onyas:
Is there any possibility to release it early next year? our customer is eager to get this fixed. If not, can you provide any plan so that we can set the expectation for the customer.

tomaswolf:
I can't promise anything. All the maintainers in this project are unpaid volunteers doing this in their spare time. I wrote in October I'd be working on this, and I am, but progress depends on the amount of spare time I have and that I am willing to dedicate to this. (Evidently, I do many other things in my spare time, too.) So I can't give a deadline.

onyas:
Got it. Thank you for your contribution.

tomaswolf:
Please see #315 for my take on this.

tomaswolf:
Superceded by #315.

PR #244, opened 2022-09-09
[SSHD-822] Enable more tests for the Netty and MINA I/O back-ends

tomaswolf:
This series of commits enables most tests for all I/O back-ends. For the details, see the individual commit messages.

There are two new very minor features; see CHANGES.md.


sshd-sftp/pom.xml
```diff
@@ -150,7 +150,6 @@
 		            <plugin>
 		                <groupId>org.apache.maven.plugins</groupId>
 		                <artifactId>maven-surefire-plugin</artifactId>
-                        <version>${surefire.plugin.version}</version>
```
lgoldstein, line 153: Why was it necessary to remove this ? If I remember correctly there are quite a few artifacts that need to use this version version.

lgoldstein (approved):
All an all, very nice work - see my one question though...

sshd-sftp/pom.xml
tomaswolf: The maven-surefire-plugin version is pinned to ${surefire.plugin.version} in the parent <pluginManagement> of the parent pom. It is therefore unnecessary and in fact bad practice to define it again here.

tomaswolf:
Looks like the ClientDeadlockTest has a problem on Windows, though. I'll have to investigate that before we can merge this.

sshd-sftp/pom.xml
lgoldstein: Got it...

tomaswolf:
> Looks like the ClientDeadlockTest has a problem on Windows with Netty, though.

Actually, it also fails here in GitHub CI on Ubuntu. Still cannot reproduce, neither on Windows, OS X, nor CentOS. (Don't have any Ubuntu setup, but it's unlikely to be an OS problem anyway.)

Tests do run here about three times slower than when I run them locally. Possibly we have a race condition somewhere that somehow leads to that CLOSE event never be reported. Perhaps some other exception that occurs and prevents us from ever calling `closeFuture.setClosed()`.

But if I can't reproduce locally, I need to see the log output from the test, and then hope it gives a clue as to what might be happening.

tomaswolf:
This looks OK now.

- The problem with `ClientDeadlockTest` was [this](https://github.com/apache/mina-sshd/pull/244/commits/331018f822ccc072acdbcc5d103bfaff0a77ce37#diff-081006a0bb09ef97f2c22a6930fdda78ddfdbbc50cd44566cd254163488c66b8L263), as explained in the commit message.
- `ClientTest.testClientDisconnect()` then also got a spurious failure; the problem there was essentially [this](https://github.com/apache/mina-sshd/pull/244/commits/ff247f539980bfe2231bddf7eaeb809f5c675002#diff-859aa84d4bf7ccdee8fabf7ec1c3b6f77574de417fd7cf308f3b54f0a4b5fe03R1422): the client closed the session, and the `suspend()` call came when the session was already closed and the `NettyIoSession.context` was already `null`. Fixed by ensuring thread safety for accesses to the `ChannelHandlerContext`, and by making the test invoke `suspend()` earlier.

Some of the test runs done for the PR also had failures of `PortForwardingTest.testRemoteForwardingSecondTimeInSameSession()` when run with MINA. The cause is unclear, and it seems to be a spuriously flaky test. The test tries to re-use a previously bound port. Although SO_REUSEADDR is set, there may be a timing issue. If there is, it would be in Java/NIO or even at the OS socket level. Maybe if one tries to re-use a socket too quickly, it can still fail.

I'd suggest to merge this PR as is, and if that port forwarding test starts failing frequently, re-open SSHD-1256 and try to fix in a separate change.

PR #260, opened 2022-10-31
[SSHD-1287] SFTP: better default buffer size handling

tomaswolf:
Use the local window's packet size for reading, and the remote window's packet size for writing (both minus a bit for protocol overhead) as the default SFTP buffer size.

Fix SftpInputStreamAsync to also work for buffer sizes greater than twice what the server will ever return. Previously, the stream read data from the wrong offset in that case.


jki6768:
har di di dooo that

PR #267, opened 2022-11-10
Validate ptyModes

rovarga:
We expect each value to be a valid UINT32 once we do PtyCapableChannelSession.doOpenPty(). Make sure we validate that assumption when we setPtyModes().

Signed-off-by: Robert Varga <robert.varga@pantheon.tech>

### DIFF
--- a/sshd-common/src/main/java/org/apache/sshd/common/channel/PtyChannelConfiguration.java
+++ b/sshd-common/src/main/java/org/apache/sshd/common/channel/PtyChannelConfiguration.java
@@ -21,6 +21,7 @@ package org.apache.sshd.common.channel;
 
 import java.util.EnumMap;
 import java.util.Map;
+import org.apache.sshd.common.util.buffer.BufferUtils;
 
 /**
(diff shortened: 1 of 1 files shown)


rovarga:
This is driven by downstream issues:
- https://jira.opendaylight.org/browse/TRNSPRTPCE-701
- https://issues.apache.org/jira/browse/KARAF-7607

We should flag any attempt to set invalid configuration before we act on them (and reject them as part of SSHD-1244).

sshd-common/src/main/java/org/apache/sshd/common/channel/PtyChannelConfiguration.java
```diff
@@ -94,6 +95,8 @@ public Map<PtyMode, Integer> getPtyModes() {
 
     @Override
     public void setPtyModes(Map<PtyMode, Integer> ptyModes) {
+        ptyModes.forEach(
+            (key, value) -> BufferUtils.validateUint32Value(value, "Mode %s has invalid value %s", key, value));
```
tomaswolf, line 99: Make a copy of the map? The caller might add more values afterwards... And should getPtyModes() return a read-only copy?

PR #278, opened 2022-11-30
SimpleGeneratorHostKeyProvider: load any kind of key

tomaswolf:
Just use the general SecrityUtils.loadKeyPairIdentities() to load the key.


tomaswolf:
This was merged as commit 3b16da43.

PR #295, opened 2022-12-19
GH-294: Fix memory leaks in SftpFileSystemProvider

### LINKED ISSUE #294: Memory Leak in SftpFileSystemProvider
### Version

master

### Bug description

The newInputStream and newOutputStream methods of SftpFileSystemProvider call the SftpFileSystem#getClient() method without call the close() method:
image

### Actual behavior

I have a thread that uploads a file to the server every 10 seconds, as time goes by, the heap memory will get bigger and bigger, and eventually the heap memory will occur out of memory:
image

When I exported the heap memory, I found that there are a lot of objects of org.apache.sshd.sftp.client.fs.SftpFileSystem$Wrapper
image

### Expected behavior

When the program is running normally, the heap memory will not overflow

### Relevant log output

_No response_

### Other information

_No response_

tomaswolf:
SftpFileSystem.getClient() returns reference-counted wrapper instances that need to be closed to avoid a memory leak via ThreadLocals. Make sure that the streams returned by SftpFileSystemProvider.newInputStream() and newOutputStream() do close the client. Also fix SftpFileSystemProvider.copy() to close the SftpClient it uses.

SftpFileSystemProvider.newDirectoryStream() and newFileChannel() already do close the SftpClient used.

Fix the SftpFileSystemTest to properly close SftpClients and DirectoryStreams.

Fixes #294.

### DIFF
--- a/sshd-sftp/src/test/java/org/apache/sshd/sftp/client/fs/SftpFileSystemTest.java
+++ b/sshd-sftp/src/test/java/org/apache/sshd/sftp/client/fs/SftpFileSystemTest.java
@@ -19,6 +19,8 @@
 package org.apache.sshd.sftp.client.fs;
 
 import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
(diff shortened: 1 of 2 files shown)


sshd-sftp/src/test/java/org/apache/sshd/sftp/client/fs/SftpFileSystemTest.java
```diff
@@ -87,6 +89,7 @@
 import org.junit.runners.MethodSorters;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.testcontainers.shaded.org.apache.commons.io.output.ByteArrayOutputStream;
```
tomaswolf, line 92: Wrong import.

tomaswolf: Fixed.

PR #314, opened 2023-01-19
Report exceptions in logs

### LINKED ISSUE #313: SftpSubsystem eats exceptions
### Version

2.9.3

### Bug description

While trying to implement a virtual file system for SFTPD, I ran into a condition in which the connection would close but there was nothing in the logs.   After 5 hours of trying to configure the logger all I could see is DEBUG messages, but no log of the exception or error.   I eventually traced it down that although there is a master catch in SftpSubsystem.run, it was never being reached.    It appears that AbstractSftpSubsystemHelper contains a lot of code with this sort of pattern...

```
      } catch (IOException | RuntimeException e) {
            sendStatus(prepareReply(buffer), id, e, SftpConstants.SSH_FXP_OPEN, path);
            return;
        }
```

It seems like poor practice to not log something when an exception occurs.   

To reproduce this I implemented a custom FileSystemFactory which returned MyFileSystem in which getSeparator() throw an exception.  To hit multiple points I had to proxy Path, FileSystem, and FileSystemProvider to the normal versions.  The exception must happen in the Path, FileSystem, or FileSystemProvider.   Throws in FileSystemFactory are caught properly as SftpSubsystem catches during prepare, but not in doProcess.

### Actual behavior

Nothing was logged.

### Expected behavior

A log message with the source of the exception.   If this is addressed, perhaps this should be logged at the  DEBUG level as this was not normally logged in the past.   After all there are a lot of tolerable exceptions (missing file, permissions, etc) that are not errors for normal operation.

### Relevant log output

_No response_

### Other information

Logging was configured with log4j with DEBUG to console.   But the lack of output is programmatic rather than the logger, so it should be universal.

Thrameos:
Fixes #313

### DIFF
--- a/sshd-sftp/src/main/java/org/apache/sshd/sftp/server/AbstractSftpSubsystemHelper.java
+++ b/sshd-sftp/src/main/java/org/apache/sshd/sftp/server/AbstractSftpSubsystemHelper.java
@@ -2965,6 +2965,9 @@ public abstract class AbstractSftpSubsystemHelper
      */
     protected void sendStatus(Buffer buffer, int id, Throwable e, int cmd, Object... args)
             throws IOException {
+        if (log.isDebugEnabled()) {
+            log.debug("doSendStatus[{}][id={},cmd={}] exception", getServerSession(), id, cmd, e);
+        }
         SftpErrorStatusDataHandler handler = getErrorStatusDataHandler();
         int subStatus = handler.resolveSubStatus(this, id, e, cmd, args);
         String message = handler.resolveErrorMessage(this, id, e, subStatus, cmd, args);
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,7 @@
 * [GH-297](https://github.com/apache/mina-sshd/issues/297) Auto-configure file password provider for reading encrypted SSH keys.
 * [GH-298](https://github.com/apache/mina-sshd/issues/298) Server side heartbeat not working.
 * [GH-300](https://github.com/apache/mina-sshd/issues/300) Read the channel id in `SSH_MSG_CHANNEL_OPEN_CONFIRMATION` as unsigned int.
+* [GH-313](https://github.com/apache/mina-sshd/issues/313) Log exceptions in the SFTP subsystem before sending a failure status reply.
 
 
 * [SSHD-1315](https://issues.apache.org/jira/browse/SSHD-1315) Do not log sensitive data at log level `TRACE`.


sshd-sftp/src/main/java/org/apache/sshd/sftp/server/AbstractSftpSubsystemHelper.java
```diff
@@ -338,6 +338,9 @@ protected void process(Buffer buffer) throws IOException {
             if (type == SftpConstants.SSH_FXP_INIT) {
                 throw e;
             }
+            if (log.isDebugEnabled()) {
+                log.debug("process({}) Exception", session, e);
+            }
```
tomaswolf, line 343: Why don't you do this _inside_ `sendStatus(Buffer buffer, int id, Throwable e, int cmd, Object... args)`? There would be only a single place to change. The exception stack trace that would be logged would identify where it occurred.

Thrameos: The method name will get lost.    As it should be on the stack trace it should be okay.    I made the change,  please review.

Thrameos:
For reference here is the log with this code being exercised.

```
2023/01/19 13:52:38,097 [sshd-SftpSubsystem-61352-thread-1] DEBUG SftpSubsystem  - doRealPath(ServerSessionImpl[test@/127.0.0.1:51365])[id=1] SSH_FXP_REALPATH (path=.)
2023/01/19 13:52:38,187 [sshd-SftpSubsystem-61352-thread-1] DEBUG SftpSubsystem  - doSendStatus[ServerSessionImpl[test@/127.0.0.1:51365]][id=1,cmd=16] exception
java.lang.RuntimeException
	at org.testing.MyFileSystem.getSeparator(MyFileSystem.java:68)
	at org.apache.sshd.common.util.SelectorUtils.translateToLocalFileSystemPath(SelectorUtils.java:545)
	at org.apache.sshd.sftp.server.SftpFileSystemAccessor.resolveLocalFilePath(SftpFileSystemAccessor.java:113)
	at org.apache.sshd.sftp.server.AbstractSftpSubsystemHelper.resolveFile(AbstractSftpSubsystemHelper.java:3029)
	at org.apache.sshd.sftp.server.AbstractSftpSubsystemHelper.doRealPath(AbstractSftpSubsystemHelper.java:1444)
	at org.apache.sshd.sftp.server.AbstractSftpSubsystemHelper.doProcess(AbstractSftpSubsystemHelper.java:392)
	at org.apache.sshd.sftp.server.SftpSubsystem.doProcess(SftpSubsystem.java:327)
	at org.apache.sshd.sftp.server.AbstractSftpSubsystemHelper.process(AbstractSftpSubsystemHelper.java:344)
	at org.apache.sshd.sftp.server.SftpSubsystem.run(SftpSubsystem.java:303)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
	at java.base/java.lang.Thread.run(Thread.java:832)
2023/01/19 13:52:38,191 [sshd-SftpSubsystem-61352-thread-1] DEBUG SftpSubsystem  - doSendStatus(ServerSessionImpl[test@/127.0.0.1:51365])[id=1] SSH_FXP_STATUS (substatus=SSH_FX_FAILURE, lang=, msg=General failure)
```

sshd-sftp/src/main/java/org/apache/sshd/sftp/server/AbstractSftpSubsystemHelper.java
```diff
@@ -1250,6 +1250,9 @@ protected void doRename(int id, String oldPath, String newPath, Collection<CopyO
             SftpFileSystemAccessor accessor = getFileSystemAccessor();
             accessor.renameFile(this, o, n, opts);
         } catch (IOException | RuntimeException | Error e) {
+            if (log.isDebugEnabled()) {
+                log.debug("doRename({})[id={}] Exception", getServerSession(), id, e);
+            }
```
tomaswolf, line 1255: Hmmm. Log-then-throw is in most cases an anti-pattern. It seems that it would not be necessary to log here: just throw and let it be handled in the sendStatus() call in doRename(Buffer, int) or doPosixRename(Buffer, int).

Except that here Error is caught, too (another anti-pattern), and then re-thrown. I would expect there to be some exception quite high up in the call hierarchy logging an Error, though. So in summary I don't think this extra logging here is necessary.

sshd-sftp/src/main/java/org/apache/sshd/sftp/server/SftpSubsystem.java
```diff
@@ -893,6 +893,9 @@ protected void doClose(int id, String handle) throws IOException {
             nodeHandle.close();
             listener.closed(session, handle, nodeHandle, null);
         } catch (IOException | RuntimeException | Error e) {
+            if (log.isTraceEnabled()) {
+                log.debug("doClose({})[id={}] Exception", session, id, e);
+            }
```
tomaswolf, line 898: This is a similar case as the doRename() over in AbstractSftpSubsystemHelper. Log-then-throw should not be necessary; IOException and RuntimeException will be logged in the sendStatus() call in doClose(Buffer, int), and for Error there should be last-resort handler somewhere higher up that logs it.

Besides, it should be log.isDebugEnabled(), if at all.

tomaswolf (commented):
I think only the change in sendStatus() is really needed.

sshd-sftp/src/main/java/org/apache/sshd/sftp/server/SftpSubsystem.java
Thrameos: done.

sshd-sftp/src/main/java/org/apache/sshd/sftp/server/AbstractSftpSubsystemHelper.java
Thrameos: done

tomaswolf:
Thanks! Is it OK for you if I squash this into a single commit before merging?

Thrameos:
Sure.   Please squash away.

PR #326, opened 2023-03-02
Versions update

lgoldstein:


tomaswolf:
In any case I see no reason to bump the minimum version of slf4j-api. Why not leave it at 1.7.32? Compare also #336.

PR #336, opened 2023-03-18
Adapt OSGi metadata of sshd.osgi to work with slf4j-api 1 and 2

HannesWell:
SSHD-Mina seems to be compatible with slf4j 1.7.x and 2.0.x. At least my local build (`mvn clean verify`) succeed without errors. I only had to disable the license check for slf4j.api.
Please let me know if my assumption about compatibility in regards of slf4j-2 is wrong.

This PR adjusts the OSGi `Import-Package` entries in the MANIFEST.MF to make sshd-mina work with both versions of the slf4j.api bundle in an OSGi runtime.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -107,6 +107,7 @@
         <groovy.version>3.0.11</groovy.version>
         <bouncycastle.version>1.70</bouncycastle.version>
         <slf4j.version>1.7.32</slf4j.version>
+        <slf4j.upper.bound>3</slf4j.upper.bound>
         <logback.version>1.2.11</logback.version>
         <spring.version>5.3.20</spring.version>
             <!-- NOTE: upgrading to 6.x requires Java 11 -->
@@ -1464,7 +1465,11 @@
                         </goals>
                         <configuration>
                             <instructions>
-                                <Import-Package>org.apache.sshd*;version="[$(version;==;${sshd.osgi.version.clean}),$(version;=+;${sshd.osgi.version.clean}))",*</Import-Package>
+                                <Import-Package><![CDATA[
+                                  org.apache.sshd*;version="[$(version;==;${sshd.osgi.version.clean}),$(version;=+;${sshd.osgi.version.clean}))",
+                                  org.slf4j*;version="${range;[==,${slf4j.upper.bound})}",
+                                  *
+                                  ]]></Import-Package>
                                 <Export-Package>*;-noimport:=true</Export-Package>
                             </instructions>
                             <noWarningProjectTypes>pom</noWarningProjectTypes>


HannesWell:
@tomaswolf or @gnodet could you please review this PR.

tomaswolf:
_Running_ against slf4j 2.0.x might work, but we do have a problem building against it: we always get an NPE from the maven-checkstyle-plugin. For instance https://github.com/apache/mina-sshd/actions/runs/4462053052/jobs/7836323807?pr=326 . I get the same locally also with maven-checkstyle-plugin 3.1.2 and slf4j 2.0.7.

HannesWell:
Yes I also could not build sshd with slf4j-2 with all checks enabled and it failed with the same error.

Using Java-17 (with Helpfull NPEs) showed that `licenseArtifact` is null:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.2:check (verify-style) on project sshd:
Execution verify-style of goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.2:check failed: Cannot invoke
"org.apache.maven.artifact.Artifact.getFile()" because "licenseArtifact" is null -> [Help 1]
```

But after disabling the check every class compiled and every test passed locally for me.

This change is only about widening the Import-Package version for the runtime, so it could be ignored for now that the full build does not work for slf4j-2. But if you if want to build against both versions in the CI or only want to build against slf4j-2, then this should of course be fixed. But I don't know how. What even is a licenseArtitfact? Can this be worked around by some configuration or is that potentially a bug in the checkstylte plugin?

HannesWell:
So do you plan to fix the build with slf4j-2 first bevor considering this?
Because, at least for me, it is not a strict prerequiste.
You can verify that SLF4J 2.0.7 works by temporarliy/locally disabling the Checkstyle Plugin for a build.

tomaswolf:
Actually, I think we should not bump the slf4j version in our pom. We should not force users of this library to use slf4j 2.

HannesWell:
> > So do you plan to fix the build with slf4j-2 first bevor considering this?
> 
> IMO, not a priority at this time - 1.7.32 works for us - the users may use 2.x at runtime if they wish to - fully compatible with 1.7.32. Perhaps we should define the scope of 1.7.32 as _compile_ ...

Yes I'm totally fine that you build with 1.7.32, I just want to use it with slf4j-2 in OSGi runtimes, which the metadata adjusted in this PR prevented. :)

Thank you @tomaswolf for your review and merging this.
You don't publish snapshots, do you?
Do you already have plans for the next release that includes this change?

PR #343, opened 2023-03-21
Simplify maven-bundle-plugin configuration

HannesWell:
Instead of the 'maven-bundle-plugin:cleanVersions' goal, use BND-lib's range macro:
https://bnd.bndtools.org/macros/range.html
@tomaswolf maybe you are interested in this as well?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1446,30 +1446,19 @@
                 <extensions>true</extensions>
                 <executions>
                     <execution>
-                        <id>versions</id>
-                        <phase>validate</phase>
-                        <goals>
-                            <goal>cleanVersions</goal>
-                        </goals>
-                        <configuration>
-                            <versions>
-                                <sshd.osgi.version.clean>${project.version}</sshd.osgi.version.clean>
-                            </versions>
-                        </configuration>
-                    </execution>
-                    <execution>
                         <id>bundle-manifest</id>
                         <phase>process-classes</phase>
                         <goals>
                             <goal>manifest</goal>
                         </goals>
                         <configuration>
+                            <supportIncrementalBuild>true</supportIncrementalBuild>
                             <instructions>
                                 <Import-Package><![CDATA[
-                                  org.apache.sshd*;version="[$(version;==;${sshd.osgi.version.clean}),$(version;=+;${sshd.osgi.version.clean}))",
-                                  org.slf4j*;version="${range;[==,${slf4j.upper.bound})}",
+                                  org.apache.sshd*;version="$<range;[===,=+);$<maven_version;${project.version}>>",
+                                  org.slf4j*;version="$<range;[==,${slf4j.upper.bound})>",
                                   *
-                                  ]]></Import-Package>
+                                ]]></Import-Package>
                                 <Export-Package>*;-noimport:=true</Export-Package>
                             </instructions>
                             <noWarningProjectTypes>pom</noWarningProjectTypes>


pom.xml
```diff
@@ -1445,31 +1445,20 @@
...
                         <phase>process-classes</phase>
                         <goals>
                             <goal>manifest</goal>
                         </goals>
                         <configuration>
+                            <supportIncrementalBuild>true</supportIncrementalBuild>
                             <instructions>
                                 <Import-Package><![CDATA[
-                                  org.apache.sshd*;version="[$(version;==;${sshd.osgi.version.clean}),$(version;=+;${sshd.osgi.version.clean}))",
+                                  org.apache.sshd*;version="${range;[==,=+);${maven_version;${project.version}}}",
```
tomaswolf, line 1458: Nice.

However: I just noticed that here (and in the slf4j line below) we have an unholy mixture of bnd macro references and maven property references. What happens if someone defines a maven property named "range"?

bnd allows other brackets to be used. Can we please use `$(...)` for bnd, and reserve `${...}` for maven? Or use `$${macro...}` for bnd references; see [the maven-bundle-plugin documentation](https://felix.apache.org/documentation/_attachments/components/bundle-plugin/manifest-mojo.html#instructions).

Only tangentially related to this change: I wonder if ==,=+ is good. Apache MINA sshd has an unfortunate habit of breaking API in minor versions, so it might actually be safer to require exact matches, and not a range at all... something like `$(version;===;$(maven_version;${project.version}))`. Or at least use a range [===;=+).

HannesWell: > However: I just noticed that here (and in the slf4j line below) we have an unholy mixture of bnd macro references and maven property references. What happens if someone defines a maven property named "range"?

Good point and goo question. But because after range there is some more I think the maven properties substitution would not interpret this as range variable. But I'm not sure about that and better save than sorry. :)

> bnd allows other brackets to be used. Can we please use `$(...)` for bnd, and reserve `${...}` for maven? Or use `$${macro...}` for bnd references; see [the maven-bundle-plugin documentation](https://felix.apache.org/documentation/_attachments/components/bundle-plugin/manifest-mojo.html#instructions).

I changed it to angular brackets now so that the variable/macro end is not confused with a inklusive/exclusive range bound `)` or `]`. This is the list of all allowed types:
https://bnd.bndtools.org/chapters/850-macros.html section 35.1 Macro patterns

> Only tangentially related to this change: I wonder if ==,=+ is good. Apache MINA sshd has an unfortunate habit of breaking API in minor versions, so it might actually be safer to require exact matches, and not a range at all... something like `$(version;===;$(maven_version;${project.version}))`. Or at least use a range [===;=+), which is what the Eclipse Orbit re-bundling uses.

As you prefer :) But a like `bundle-version="1.2.3"` only defines a lower bound, what would make it even worse if MINA does not follow Semantic Versioning. I now changed it to `[===;=+)`, but if you only want to allow that exact version `[===;===]` would probably the way to go, which leads to ranges like `bundle-version="[1.2.3,1.2.3]"`. Theoretically there is still some space for different qualifiers, but if you don't use them practicality it does not allow something else.

By the way in the Eclipse Platform we are now using Maven-Targets to obtain artifacts from Maven-Central directly that already have proper OSGi metadata built in, like MINA SSHD, and try to fade out Orbit:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/d2ebbe9ee4f98f9b0c8423b3df1e5a42544ff925/eclipse.platform.releng.prereqs.sdk/eclipse-sdk-prereqs.target#L171-L176

That's the reason why I reached out to you with my previous PR. :)
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/981

pom.xml
```diff
@@ -1445,31 +1445,20 @@
...
                             <goal>manifest</goal>
                         </goals>
                         <configuration>
+                            <supportIncrementalBuild>true</supportIncrementalBuild>
                             <instructions>
                                 <Import-Package><![CDATA[
-                                  org.apache.sshd*;version="[$(version;==;${sshd.osgi.version.clean}),$(version;=+;${sshd.osgi.version.clean}))",
-                                  org.slf4j*;version="${range;[==,${slf4j.upper.bound})}",
+                                  org.apache.sshd*;version="$<range;[===,=+);$<maven_version;${project.version}>>",
+                                  org.slf4j*;version="$<range;[==,$<slf4j.upper.bound>)>",
```
tomaswolf, line 1459: `${slf4j.upper.bound}`. That is a maven property, not a bnd macro!

HannesWell: That's right. I did more than I wanted. But since all properties of the project are injected into the BND-Analyser that property would have been substituted there and the result would have been the same. Now it is again done by maven.

HannesWell:
Thank you @tomaswolf.

Can you tell me if there are already plans for the next release of sshd or can you give a rough estimate?

tomaswolf:
I had seen that question in the other PR; was getting around to it :-)

We have no release plan; we release whenever we feel it's needed and our release manager has the time to do so.

Luckily we don't have a lot in the pipeline currently, and no blocker bugs.

If we want to have that new release used in Eclipse 2023-06, we should target a new Apache MINA sshd release for Eclipse 2023-06 M3 (2023-05-26). I'll need some time to prepare a JGit version using the new release. (First, I run the JGit SSH tests as part of my approval of an Apache MINA sshd release, and second, we'd want the new JGit in that Eclipse to use the new Apache MINA sshd release anyway.) I don't think I'll get to that before April 15. So a reasonable timeframe for a new Apache MINA sshd release would be "first half of May". That would give me a little time to get the JGit changes in before M3.

I suggest we go to version 2.10; there are some API changes.

@gnodet, would you be available beginning of May to do a new release?

HannesWell:
> Luckily we don't have a lot in the pipeline currently, and no blocker bugs.

Glad to hear that.

> If we want to have that new release used in Eclipse 2023-06, we should target a new Apache MINA sshd release for Eclipse 2023-06 M3 (2023-05-26). I'll need some time to prepare a JGit version using the new release. (First, I run the JGit SSH tests as part of my approval of an Apache MINA sshd release, and second, we'd want the new JGit in that Eclipse to use the new Apache MINA sshd release anyway.) I don't think I'll get to that before April 15. So a reasonable timeframe for a new Apache MINA sshd release would be "first half of May". That would give me a little time to get the JGit changes in before M3.

It would be really great and highly appreciated to have the mentioned changes released in time for Eclipse 2023-06 and the previous version of sshd replaced by the new one, so that the Eclipse SimRel can complete the shift to sfl4j-2.
In case I can assist you with anything in this regard, please let me know.

tomaswolf:
I've created [JGit issue 581770](https://bugs.eclipse.org/bugs/show_bug.cgi?id=581770) to track the JGit update.

HannesWell:
> I've created [JGit issue 581770](https://bugs.eclipse.org/bugs/show_bug.cgi?id=581770) to track the JGit update.

Great.

Would it be possible to deploy a snapshot version of the current state of sshd to any snapshot repo, e.g. the one of Apache? That would allow to test the current state in verification builds.

gnodet:
> I had seen that question in the other PR; was getting around to it :-)
> 
> We have no release plan; we release whenever we feel it's needed and our release manager has the time to do so.
> 
> Luckily we don't have a lot in the pipeline currently, and no blocker bugs.
> 
> If we want to have that new release used in Eclipse 2023-06, we should target a new Apache MINA sshd release for Eclipse 2023-06 M3 (2023-05-26). I'll need some time to prepare a JGit version using the new release. (First, I run the JGit SSH tests as part of my approval of an Apache MINA sshd release, and second, we'd want the new JGit in that Eclipse to use the new Apache MINA sshd release anyway.) I don't think I'll get to that before April 15. So a reasonable timeframe for a new Apache MINA sshd release would be "first half of May". That would give me a little time to get the JGit changes in before M3.
> 
> I suggest we go to version 2.10; there are some API changes.
> 
> @gnodet, would you be available beginning of May to do a new release?

I'll try to do that next week.  Is there any remaining issue/PR to look at ?

tomaswolf:
> Is there any remaining issue/PR to look at ?

@gnodet: none except your own #362, which I think should also go in that release. (But it has to pass the CI build first. Looks like there's still some work needed in the tests. Plus it should be mentioned in CHANGES.md when it goes in.)

Otherwise I think we're good to go.

There's also the closed issue #294, which perhaps wasn't quite fixed by  PR #295 according to comments, but for which I have no reproducer. It's about a memory leak involving ThreadLocals in SftpFileSystemProvider, which might be your area of expertise :-).

HannesWell:
> > @gnodet, would you be available beginning of May to do a new release?
> 
> I'll try to do that next week. Is there any remaining issue/PR to look at ?

Thanks!

PR #347, opened 2023-03-25
Update docs

thachlp:
Motivation:
There is no constructor like: `ProcessShellFactory(String[] command)`
The docs should be update

### DIFF
--- a/docs/server-setup.md
+++ b/docs/server-setup.md
@@ -45,8 +45,7 @@ implementation that you can use if you want. This implementation will create a p
 so it's mostly useful to launch the OS native shell. E.g.,
 
 ```java
-sshd.setShellFactory(new ProcessShellFactory(new String[] { "/bin/sh", "-i", "-l" }));
-
+sshServer.setShellFactory(new ProcessShellFactory("/bin/sh", "-i", "-l" ));
 ```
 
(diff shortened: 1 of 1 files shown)


docs/server-setup.md
```diff
@@ -45,8 +45,7 @@ implementation that you can use if you want. This implementation will create a p
 so it's mostly useful to launch the OS native shell. E.g.,
 
 ```java
-sshd.setShellFactory(new ProcessShellFactory(new String[] { "/bin/sh", "-i", "-l" }));
-
+sshServer.setShellFactory(new ProcessShellFactory("/bin/sh -i -l", "-i", "-l" ));
 ```
```
thachlp, line 49: Don't fully understand but this configuration works. I also need to improve this too.

thachlp: So we should fix the implementation. After trying what you suggested, it didn't work for me.

tomaswolf: It's worse, but understandable if you look at existing uses. `command` is supposed to be the full command as it was received, `elements` is supposed to be the individual parts. The correct invocation is
```
new ProcessShellFactory("/bin/sh -i -l", "/bin/sh", "-i", "-l")
```
(This is from memory; I'm not at my development machine where I could easily double-check this.)

PR #368, opened 2023-05-04
[SSHD-1259] Match key type in known_hosts lookup

FliegenKLATSCH:


tomaswolf:
I don't think this is correct.

First, the client has already sent its KEX proposal, with a different list. The server-sig-algs KEX extension message can arrive only after the server has received the client's KEX proposal, and that includes ext-info-c.

Second, server-sig-algs is signature algorithms, not key types. But known_hosts records key types. So for RSA keys, they might not match. (Key type "ssh-rsa", signature algorithm for instance "rsa-sha2-512".)

In JGit, we've implemented this "prefer already known keys" in `ClientSession.resolveAvailableSignaturesProposal()`, and we make sure to add all three RSA signature algorithm names at the front if ssh-rsa is a known host key type.

Getting the order wrong in the KEX proposal may lead to hard to track down negotiation failures such as [Eclipse bug 574635](https://bugs.eclipse.org/bugs/show_bug.cgi?id=574635).

FliegenKLATSCH:
I removed that part, was just "sugar" because I noticed openssh is a bit more clever while testing around a bit..

sshd-core/src/test/java/org/apache/sshd/client/keyverifier/KnownHostsServerKeyVerifierTest.java
```diff
@@ -62,16 +46,33 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.security.KeyPair;
+import java.security.PublicKey;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
```
tomaswolf, line 65: Please make sure you have the same ordering as before.

sshd-core/src/main/java/org/apache/sshd/client/keyverifier/KnownHostsServerKeyVerifier.java
```diff
@@ -60,6 +39,22 @@
...
+import java.nio.file.Files;
+import java.nio.file.LinkOption;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.security.GeneralSecurityException;
+import java.security.PublicKey;
+import java.util.*;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
+
```
tomaswolf, line 57: Please don't change the order of import blocks. If necessary, adjust your IDE's settings.

sshd-core/src/main/java/org/apache/sshd/client/keyverifier/KnownHostsServerKeyVerifier.java
```diff
@@ -60,6 +39,22 @@
...
+import java.io.Writer;
+import java.net.SocketAddress;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.LinkOption;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.security.GeneralSecurityException;
+import java.security.PublicKey;
+import java.util.*;
```
tomaswolf, line 54: We don't use wildcard imports.

sshd-core/src/main/java/org/apache/sshd/client/keyverifier/KnownHostsServerKeyVerifier.java
```diff
@@ -491,43 +486,68 @@ protected boolean acceptKnownHostEntry(
         return true;
     }
 
-    protected HostEntryPair findKnownHostEntry(
-            ClientSession clientSession, SocketAddress remoteAddress, Collection<HostEntryPair> knownHosts) {
+    protected List<HostEntryPair> getAllEntriesForHost(ClientSession clientSession, SocketAddress remoteAddress, Collection<HostEntryPair> knownHosts) {
```
tomaswolf, line 489: Return `Collections.emptyList()` instead of `null`.

sshd-core/src/main/java/org/apache/sshd/client/keyverifier/KnownHostsServerKeyVerifier.java
```diff
@@ -491,43 +486,68 @@ protected boolean acceptKnownHostEntry(
...
+        return matches;
+    }
+
+    protected HostEntryPair findKnownHostEntry(
+            ClientSession clientSession, SocketAddress remoteAddress, Collection<HostEntryPair> knownHosts, String serverKeyType) {
+
+        List<HostEntryPair> filtered = getAllEntriesForHost(clientSession, remoteAddress, knownHosts);
+        if (filtered == null) {
+            return null;
+        }
```
tomaswolf, line 532: This `if` is not needed if the method returns an empty list, not `null`.

sshd-core/src/main/java/org/apache/sshd/client/keyverifier/KnownHostsServerKeyVerifier.java
```diff
@@ -267,36 +268,55 @@ protected PublicKeyEntryResolver getFallbackPublicKeyEntryResolver() {
...
         }
 
-        try {
-            if (!acceptModifiedServerKey(clientSession, remoteAddress, entry, expected, serverKey)) {
-                return false;
+        if (!keyMatches.isEmpty()) {
+            return true;
+        }
+
+        for (HostEntryPair match : hostMatches) {
```
tomaswolf, line 293: This loop again goes over revoked entries. Those should probably never be fed to `acceptModifiedServerKey()`. Because if we have only revoked keys, it's not a "modified server key" case, but an "unknown server key".

sshd-core/src/main/java/org/apache/sshd/client/keyverifier/KnownHostsServerKeyVerifier.java
```diff
@@ -267,36 +268,55 @@ protected PublicKeyEntryResolver getFallbackPublicKeyEntryResolver() {
...
+            KnownHostEntry entry = match.getHostEntry();
+            PublicKey expected = match.getServerKey();
+
+            try {
+                if (acceptModifiedServerKey(clientSession, remoteAddress, entry, expected, serverKey)) {
+                    updateModifiedServerKey(clientSession, remoteAddress, serverKey, knownHosts, match);
+                    return true;
+                }
```
tomaswolf, line 301: I presume you can return false here after the `if`. No need to ask several times whether to accept a modified key -- and we know that none of the entries match the server key.

sshd-core/src/main/java/org/apache/sshd/client/keyverifier/KnownHostsServerKeyVerifier.java
```diff
@@ -267,36 +268,55 @@ protected PublicKeyEntryResolver getFallbackPublicKeyEntryResolver() {
...
+                warn("acceptKnownHostEntries({})[{}] failed ({}) to accept modified server key: {}",
+                        clientSession, remoteAddress, t.getClass().getSimpleName(), t.getMessage(), t);
             }
-        } catch (Throwable t) {
-            warn("acceptKnownHostEntries({})[{}] failed ({}) to accept modified server key: {}",
-                    clientSession, remoteAddress, t.getClass().getSimpleName(), t.getMessage(), t);
-            return false;
         }
 
+        return false;
```
tomaswolf, line 308: If the above loop skips revoked entries (or `acceptModifiedServerKey` skipped them and returned `false`), then we can get here if we have for instance only revoked keys, all different from the actual server key. So here we should then call `return acceptUnknownHostKey(...);`.

tomaswolf (commented):
So now this also incorporates "revoked" handling. Thanks for your efforts; it's greatly appreciated!

I think it may still not be quite correct if there are only revoked keys for a host.

Now that this handles revocations, there should also be test cases with revocations. At least three:

- a revocation of the test server key &rarr; should not be accepted.
- one host with a single revocation of some key that is not the test server key &rarr; should go into the "accept unknown key?" case.
- one host with a revocation for some key, plus a valid entry for the test server key &rarr; should be accepted.

sshd-core/src/main/java/org/apache/sshd/client/keyverifier/KnownHostsServerKeyVerifier.java
FliegenKLATSCH: My first thought was to change the signature and pass a list (and adding another list of all names under which the real key is known, similar to openssh, which shows that with the warning), but I assume that "big" change is not ideal - assuming a lot of implementations overwrite this function.
In the end, it may be (too) special cases where someone relies on an specific existing entry, so you are probably right - but on the other end, if someone accepts it right away, it will also not ask again, so it doesn't really hurt?
I am wondering if some implementations would do an external lookup with the old key..

FliegenKLATSCH:
Regarding tests; I think all cases are covered by KnownHostsServerKeyVerifierTest.java

sshd-core/src/main/java/org/apache/sshd/client/keyverifier/KnownHostsServerKeyVerifier.java
```diff
@@ -267,36 +268,63 @@ protected PublicKeyEntryResolver getFallbackPublicKeyEntryResolver() {
...
-        try {
-            if (!acceptModifiedServerKey(clientSession, remoteAddress, entry, expected, serverKey)) {
-                return false;
+        if (!keyMatches.isEmpty()) {
+            return true;
+        }
+
+        List<HostEntryPair> acceptedHostMatches = hostMatches.stream()
+                .filter(k -> acceptKnownHostEntry(clientSession, remoteAddress, serverKey, k.getHostEntry()))
+                .collect(Collectors.toList());
```
tomaswolf, line 295: A `findAny` would be sufficient instead of collecting the list. If none is found, call `acceptUnknownHostKey`, otherwise call `acceptModifiedServerKey()` _once_ with the found entry.

The normal operation with `acceptModifiedServerKey()` is to ask the user. And if the user says "no", it'll return false. There is no need for a loop here going over all the other keys that we know and that the server didn't present. This will only result in the user having to say "no" multiple times.

Also: we know here that `hostMatches` contains no matching key here. All this does is filter out revoked keys. Please name this variable `nonRevokedMatch`.

It might be even clearer to use `.filter(k -> !"revoked".equals(k.getMarker()))` directly here.

sshd-core/src/main/java/org/apache/sshd/client/keyverifier/KnownHostsServerKeyVerifier.java
```diff
@@ -267,36 +268,63 @@ protected PublicKeyEntryResolver getFallbackPublicKeyEntryResolver() {
...
-            return acceptKnownHostEntry(clientSession, remoteAddress, serverKey, entry);
+        String serverKeyType = KeyUtils.getKeyType(serverKey);
+
+        List<HostEntryPair> keyMatches = hostMatches.stream()
+                .filter(entry -> serverKeyType.equals(entry.getHostEntry().getKeyEntry().getKeyType()))
+                .filter(k -> KeyUtils.compareKeys(k.getServerKey(), serverKey))
+                .collect(Collectors.toList());
+
+        if (keyMatches.stream()
+                .anyMatch(k -> !acceptKnownHostEntry(clientSession, remoteAddress, serverKey, k.getHostEntry()))) {
```
tomaswolf, line 285: This determines whether any of the matching entries is revoked. It might be clearer to use `.anyMatch(k -> "revoked".equals(k.getMarker()))` directly.

PR #372, opened 2023-05-16
READ (only) default mode in SftpFileSystemProvider.newFileChannel()

hannibal218bc:
According to [upstream javadoc](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/FileChannel.html#open-java.nio.file.Path-java.util.Set-java.nio.file.attribute.FileAttribute...-), the default mode is READ (only) when there is no mode specified:

> The [READ](https://docs.oracle.com/javase/8/docs/api/java/nio/file/StandardOpenOption.html#READ) and [WRITE](https://docs.oracle.com/javase/8/docs/api/java/nio/file/StandardOpenOption.html#WRITE) options determine if the file should be opened for reading and/or writing. If neither option (or the [APPEND](https://docs.oracle.com/javase/8/docs/api/java/nio/file/StandardOpenOption.html#APPEND) option) is contained in the array then the file is opened for reading

SftpFileSystemProvider used `READ, WRITE` which leads to `Permission denied` errors when trying to access read-only files (see [this Stackoverflow thread](https://stackoverflow.com/questions/49235417/sshd-get-content-of-readonly-file)).

This MR suggests to use a standard-complaint setting; however, this might cause regressions...

### DIFF
--- a/sshd-sftp/src/main/java/org/apache/sshd/sftp/client/fs/SftpFileSystemProvider.java
+++ b/sshd-sftp/src/main/java/org/apache/sshd/sftp/client/fs/SftpFileSystemProvider.java
@@ -506,7 +506,7 @@ public class SftpFileSystemProvider extends FileSystemProvider {
             throws IOException {
         Collection<OpenMode> modes = OpenMode.fromOpenOptions(options);
         if (modes.isEmpty()) {
-            modes = EnumSet.of(OpenMode.Read, OpenMode.Write);
+            modes = EnumSet.of(OpenMode.Read);
         }
         // TODO: process file attributes
         SftpPath p = toSftpPath(path);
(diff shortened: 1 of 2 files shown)


tomaswolf:
Thanks for this fix. Unfortunately tests are failing.

A quick fix might be to open the channel for writing in that failing test method.

But it seems that the locking implementation doesn't map exceptions correctly when channels are opened for reading only (server just returns a "General Failure"); surely that could be improved. And moreover, the javadoc on `FileChannel.lock()` says a `NonWritableChannelException` should be thrown if the channel wasn't writeable. It should be possible to check that on the client side before even making any remote call.

We have no tests for other combinations (read lock on a write-only channel).

tomaswolf:
Created bug #383 for this.

PR #404, opened 2023-08-10
Add support for Proxy Protocol V2

fherbreteau:
Proposed implementation for PROXY Protocol V2 for use in SSHD.

### DIFF
--- a/sshd-common/src/main/java/org/apache/sshd/common/util/buffer/Buffer.java
+++ b/sshd-common/src/main/java/org/apache/sshd/common/util/buffer/Buffer.java
@@ -292,6 +292,10 @@ public abstract class Buffer implements Readable {
         return v;
     }
 
+    public int getUShort() {
+        return getShort() & 0xFFFF;
+    }
+
     public int getInt() {
         return (int) getUInt();
     }
--- a/sshd-contrib/pom.xml
+++ b/sshd-contrib/pom.xml
@@ -80,6 +80,16 @@
             <type>test-jar</type>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>3.24.2</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/sshd-contrib/src/main/java/org/apache/sshd/contrib/server/session/proxyprotocolv2/utils/ProxyUtils.java
+++ b/sshd-contrib/src/main/java/org/apache/sshd/contrib/server/session/proxyprotocolv2/utils/ProxyUtils.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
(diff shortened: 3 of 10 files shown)


fherbreteau:
Proposed implementation for PROXY Protocol V2 for use in SSHD.

sshd-contrib/src/main/java/org/apache/sshd/contrib/server/session/proxyprotocolv2/utils/ProxyUtils.java
```diff
@@ -0,0 +1,70 @@
...
+     */
+    public static int readUnsignedShort(Buffer buffer) {
+        // Read the expected bytes from the buffer
+        byte[] datas = new byte[2];
+        buffer.getRawBytes(datas);
+        // Extends the array with 2 more bytes
+        datas = Arrays.copyOf(datas, 4);
+        // Extract the value
+        int value = (int) (BufferUtils.getUInt(datas, 0, Integer.BYTES) >> Short.SIZE);
+        return value & 0xFFFF;
```
tomaswolf, line 53: This is a bit strange. Why not simply `return Buffer.getShort() & 0xFFFF;` ? (Compare commit c411efb83)

fherbreteau: You're right. This way is much better. I'll fix this.

tomaswolf: We could add one. Feel free to do so :-) It wasn't needed so far; the SSH protocol doesn't use 16bit values. (I think there is one extension that uses a 16bit value somewhere, check the call hierarchy of getShort() -- which is also why nobody noticed until recently that the implementation was completely broken.)

fherbreteau: Thx for the info.

I will add a `Buffer.getUShort()` to improve the code readability of my implementation.

tomaswolf (commented):
Looks good; thanks. Just some nits.

I'm a strong believer in meaningful commit messages. They help so much when one tries to follow git history, or find out in a few years from the git history why certain things were done in a particular way.

I suggest to squash these three commits into one and give it a commit message that gives just a little bit more context. Perhaps
```
Add support for Proxy Protocol V2

HAProxy protocol V2 has several improvements over the V1 protocol. Add an
acceptor and parsing support for the protocol as described at [1].

Also add a Buffer.getUShort() method to read an unsigned 16bit value from
a Buffer. The proxy protocol V2 uses such values.

[1] https://www.haproxy.org/download/2.7/doc/proxy-protocol.txt
```
or some such.

sshd-common/src/main/java/org/apache/sshd/common/util/buffer/Buffer.java
```diff
@@ -292,6 +292,10 @@ public short getShort() {
         return v;
     }
 
+	public int getUShort() {
+		return getShort() & 0xFFFF;
+	}
+
```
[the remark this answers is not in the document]
fherbreteau: Indentation fixed. 
Sorry for that.

sshd-contrib/pom.xml
```diff
@@ -80,6 +80,16 @@
...
+		<dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-core</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.assertj</groupId>
+			<artifactId>assertj-core</artifactId>
+			<version>3.24.2</version>
+		</dependency>
```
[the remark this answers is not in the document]
fherbreteau: Fixed.

fherbreteau:
Indentation fixed.
Commit squashed with new commit message.
btw I've reused your proposal as it fits the intent of the MR.

PR #416, opened 2023-09-18
[SSHD-1330] Using common keep-alive global request handler in client as well as server

lgoldstein:


tomaswolf:
Moving the `KeepAliveHandler` to a different package will break any code that currently references it using the old package. Not sure how to deal with this.

Issue #445, opened 2023-12-19
Terrapin Mitigation: "strict-kex"


ecki:
### Description

Hello,

is Mina or any contributor planning to work on adding the new OpenSSH protocol extension "strict-KEX" for mitigating Terrapin attacks?

Also did somebody check for the counter overflow conditions mentioned in the advanced counter manupulation section of the paper?

BTW: when implementing config options, I would do it like jssh which allows to define a "required" mode, so you can set up a listener which rejects any handshakes without this protection. (if you want to make "supported" configurable I dont care, seems to be not a big compat problem if implemented correctly)

In addition to resetting the counters the strict mode probbaly also should reject the "filler" debug and ignore mesages - I hope PMC received detailed guidance from the Terrapin team?

### Motivation

Users want to mitigate the new protocol attac which can only work if client and server are extended.

### Alternatives considered

Turning off the ciphers is an interop problem.

### Additional context

https://terrapin-attack.com

TrueSkrillor:
> According to [OpenSSH PROTOCOL](https://github.com/openssh/openssh-portable/blob/master/PROTOCOL) - section `1.9 transport: strict key exchange extension`
> 
> > Ater sending or receiving a SSH2_MSG_NEWKEYS message, reset the packet sequence number to zero
> 
> However, it does not specify **which** sequence number to reset - the incoming our outgoing. Bear in mind that the NEWKEYS message is "symmetrical" - if we sent one, then an incoming one is due any time and vice versa. Since the sequence number is part of the encryption (if this is not the 1st NEWKEYS) then we need to know whether we sent the request or are responding to it.

Strict key exchange resets the sequence number of the corresponding direction after it has received a SSH_MSG_NEWKEYS message. I. e. when receiving SSH_MSG_NEWKEYS from the remote peer reset the incoming sequence number, when sending SSH_MSG_NEWKEYS reset the outgoing sequence number. The SSH_MSG_NEWKEYS itself is handled under the old sequence numbers (sequence number 0 will be the first message after SSH_MSG_NEWKEYS).

TrueSkrillor:
Yeah, I expected that kind of response, given that I replied 6 hours later. If any additional questions pop up during the implementation of "strict kex", just let me know. I'm most likely to be able to answer them.

ecki:
Lyle, let me know when you think I can have a look. We are using it heavily inside karaf, so we are happy to help.

cywt123:
[...]

PR #446, opened 2023-12-22
[GH-445] Implemented OpenSSH strict KEX protocol

lgoldstein:
@tomaswolf @gnodet I value your opinion so I would appreciate it if you could take a look at this. I had some doubts as to some pieces of code (don't want to bias your review so I won't say where) so I would really appreciate a fresh set of eyes on this. I apologize for the seemingly large diff on `AbstractSession`, but it is **not** my doing - the Maven formatting plugin we use did this (for whatever reason...). However, I believe it is still quite readable.

As far as testing goes, I have tested it with `OpenSSH_8.9p1 Ubuntu-3ubuntu0.5, OpenSSL 3.0.2 15 Mar 2022`. I do plan on adding unit tests for the code, but before I do that I prefer having your feedback, in case more changes are necessary. In this context, I do encourage you to test **both** client and server interoperability with this code as well - perhaps on a different OpenSSH version. Note that the feature requires turning it on **explicitly** - I have added a custom CLI option that you can use for this purpose with the CLI code  - `-o UseStrictKex=yes`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -38,7 +38,7 @@
 
 ### [GH-445 - Terrapin attack mitigation](https://github.com/apache/mina-sshd/issues/429)
 
-There is a **new** `CoreModuleProperties` property that controls the mitigation for the [Terrapin attach](https://terrapin-attack.com/) via what is known as
+There is a **new** `CoreModuleProperties` property that controls the mitigation for the [Terrapin attack](https://terrapin-attack.com/) via what is known as
 "strict-KEX" (see [OpenSSH PROTOCOL - 1.9 transport: strict key exchange extension](https://github.com/openssh/openssh-portable/blob/master/PROTOCOL)).
 It is **disabled** by default due to its experimental nature and possible interoperability issues, so users who wish to use this feature must turn it on *explicitly*.
 
@@ -56,7 +56,18 @@ Provide (read-only) public access to internal session state values related to KE
 * *getSessionKexDetails*
 * *getSessionCountersDetails*
 
+### Added `SessionListener#sessionPeerIdentificationSent` callback
+
+Invoked after successful or failed attempt to send client/server identification to peer. The callback provides the failure error if such occurred.
+
 ## Potential compatibility issues
+
+### Added finite wait time for default implementation of `ClientSession#executeRemoteCommand`
+
+* `CoreModuleProperties#EXEC_CHANNEL_OPEN_TIMEOUT` - default = 30 seconds.
+* `CoreModuleProperties#EXEC_CHANNEL_CMD_TIMEOUT` - default = 30 seconds.
+
+This may cause failures for code that was running long execution commands using the default method implementations.
 
 ## Major Code Re-factoring
 
--- a/docs/standards.md
+++ b/docs/standards.md
@@ -1,6 +1,7 @@
 # Supported standards
 
 ## Reference implementation documentation
+
 * [RFC 4251 - The Secure Shell (SSH) Protocol Architecture](https://tools.ietf.org/html/rfc4251)
 * [RFC 4252 - The Secure Shell (SSH) Authentication Protocol](https://tools.ietf.org/html/rfc4252)
 * [RFC 4253 - The Secure Shell (SSH) Transport Layer Protocol](https://tools.ietf.org/html/rfc4253)
@@ -31,10 +32,26 @@
 * [Key Exchange (KEX) Method Updates and Recommendations for Secure Shell](https://tools.ietf.org/html/draft-ietf-curdle-ssh-kex-sha2-03)
 
 ## *OpenSSH*
+
 * [OpenSSH support for U2F/FIDO security keys](https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.u2f)
     * **Note:** the server side supports these keys by default. The client side requires specific initialization
 * [OpenSSH public-key certificate authentication system for use by SSH](https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.certkeys)
 * [OpenSSH 1.9 transport: strict key exchange extension](https://github.com/openssh/openssh-portable/blob/master/PROTOCOL)
+* [(Some) OpenSSH SFTP extensions](https://github.com/openssh/openssh-portable/blob/master/PROTOCOL)
+
+**Note:** some implementations may be limited to client-side - i.e., we provide a capability for the client to detect if the server
+supports the extension and then use it, but our server does not publish it as being supported.
+
+| Section | Extension                  | Client | Server |
+| ------- | -------------------------- | ------ | ------ |
+| 4.3     | `posix-rename@openssh.com` | Yes    | Yes    |
+| 4.4     | `statvfs@openssh.com`      | Yes    | Yes    |
+| 4.4     | `fstatvfs@openssh.com`     | Yes    | Yes    |
+| 4.5     | `hardlink@openssh.com`     | Yes    | Yes    |
+| 4.6     | `fsync@openssh.com`        | Yes    | Yes    |
+| 4.7     | `lsetstat@openssh.com`     | Yes    | Yes    |
+| 4.8     | `limits@openssh.com`       | Yes    | Yes    |
+| 4.10    | `copy-data`                | Yes    | Yes    |
 
 ## SFTP version 3-6 + extensions
 
@@ -49,7 +66,6 @@
 * `copy-file`, `copy-data` - [DRAFT 00 - sections 6, 7](https://tools.ietf.org/id/draft-ietf-secsh-filexfer-extensions-00.txt)
 * `space-available` - [DRAFT 09 - section 9.2](https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-09#section-9.2)
 * `filename-charset`, `filename-translation-control` - [DRAFT 13 - section 6](https://tools.ietf.org/html/draft-ietf-secsh-filexfer-13#section-6) - only client side
-* Several [OpenSSH SFTP extensions](https://github.com/openssh/openssh-portable/blob/master/PROTOCOL)
 	
 ## Miscellaneous
 
--- a/sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
@@ -113,7 +113,7 @@ import org.apache.sshd.core.CoreModuleProperties;
  *
  * @author <a href="mailto:dev@mina.apache.org">Apache MINA SSHD Project</a>
  */
-@SuppressWarnings("checkstyle:MethodCount")
+@SuppressWarnings("checkstyle:MethodCount") // Number of methods exceeds max. allowed
 public abstract class AbstractSession extends SessionHelper {
     /**
      * Name of the property where this session is stored in the attributes of the underlying MINA session. See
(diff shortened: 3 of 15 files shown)


CHANGES.md
```diff
@@ -36,13 +36,26 @@
 
 ## Behavioral changes and enhancements
 
+### [GH-445 - Terrapin attack mitigation](https://github.com/apache/mina-sshd/issues/429)
+
+There is a **new** `CoreModuleProperties` property that controls the mitigation for the [Terrapin attach](https://terrapin-attack.com/) via what is known as
+"strict-KEX" (see [OpenSSH PROTOCOL - 1.9 transport: strict key exchange extension](https://github.com/openssh/openssh-portable/blob/master/PROTOCOL)).
+It is **disabled** by default due to its experimental nature and possible interoperability issues, so users who wish to use this feature must turn it on *explicitly*.
```
sberyozkin, line 43: IMHO it makes sense to enable it by default - otherwise the mitigation will likely not be very effective in practice, for example, users may not be aware of the risk or it may not be possible to enable it at the level of other libraries depending on sshd-core

CHANGES.md
```diff
@@ -36,14 +36,38 @@
 
 ## Behavioral changes and enhancements
 
+### [GH-445 - Terrapin attack mitigation](https://github.com/apache/mina-sshd/issues/429)
+
+There is a **new** `CoreModuleProperties` property that controls the mitigation for the [Terrapin attach](https://terrapin-attack.com/) via what is known as
```
gnodet, line 41: `[Terrapin attach]` -> `[Terrapin attack]`

docs/standards.md
```diff
@@ -29,23 +29,31 @@
...
+* `vendor-id` - [DRAFT 09 - section 4.4](https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-09#section-4.4)
+* `acl-supported` - [DRAFT 11 - section 5.4](https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-11#section-5.4)
+* `newline` - [DRAFT 09 Section 4.3](https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-09#section-4.3)
+* `md5-hash`, `md5-hash-handle` - [DRAFT 09 - section 9.1.1](https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-09#section-9.1.1)
+* `check-file-handle`, `check-file-name` - [DRAFT 09 - section 9.1.2](https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-09#section-9.1.2)
+* `copy-file`, `copy-data` - [DRAFT 00 - sections 6, 7](https://tools.ietf.org/id/draft-ietf-secsh-filexfer-extensions-00.txt)
+* `space-available` - [DRAFT 09 - section 9.2](https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-09#section-9.2)
+* `filename-charset`, `filename-translation-control` - [DRAFT 13 - section 6](https://tools.ietf.org/html/draft-ietf-secsh-filexfer-13#section-6) - only client side
+* Several [OpenSSH SFTP extensions](https://github.com/openssh/openssh-portable/blob/master/PROTOCOL)
```
gnodet, line 52: I think that line should be removed, as you create a specific section for openssh.

CHANGES.md
ecki: Yes, sending it in client should be default, server can send it only if requested, and the additional “make it mandatory” off by default on both.

sshd-common/src/main/java/org/apache/sshd/common/config/ConfigFileReaderSupport.java
```diff
@@ -89,6 +89,8 @@ public final class ConfigFileReaderSupport {
     public static final String MACS_CONFIG_PROP = "MACs";
     // see http://manpages.ubuntu.com/manpages/precise/en/man5/sshd_config.5.html
     public static final String KEX_ALGORITHMS_CONFIG_PROP = "KexAlgorithms";
+    // see https://github.com/openssh/openssh-portable/blob/master/PROTOCOL - section 1.9
+    public static final String STRICT_KEX_CUSTOM_CONFIG_PROP = "UseStrictKex";
```
ecki, line 93: I would use true/support and require and false with supported (which means offered and accepted) as the default - once it has stabilized.

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
```diff
@@ -151,24 +156,34 @@ public abstract class AbstractSession extends SessionHelper {
...
     // synchronized on kexState.
     protected DefaultKeyExchangeFuture kexInitializedFuture;
 
     /*
      * SSH packets encoding / decoding support
      */
+    /** Input packet sequence number. */
+    protected long seqi;
+    /** Output packet sequence number. */
+    protected long seqo;
```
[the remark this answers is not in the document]
ecki: Unfortunatelly zour code does touch unrelated things, it would be good not to have to review them. Besides that it is relevant to the patch because you introduced a synchronized and it was not clear to me if it’s the right one (since it looks like the increase of those variables do not know about it).

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/AbstractSession.java
```diff
@@ -1741,6 +1979,143 @@ protected byte[] sendKexInit(Map<KexProposalOption, String> proposal) throws Exc
...
+        }
+
+        String peerValue = isServerSession()
+                ? KexExtensions.STRICT_KEX_CLIENT_EXTENSION
+                : KexExtensions.STRICT_KEX_SERVER_EXTENSION;
+        if (!value.contains(peerValue)) {
+            return value;
+        }
+
+        // Just a bit of paranoia...
```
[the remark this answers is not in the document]
ecki: When you check if the peer value is present you can also check that the own (I.e. c instead of s for a client connection) is not present, because that would not be allowed. RFC8303 (which is analogous to the strict kex) says “may disconnect” in 2.2:

>  Implementations MUST NOT send an incorrect indicator name for their role.  Implementations MAY disconnect if the counterparty sends an incorrect indicator.  If "ext-info-c" or "ext-info-s" ends up being negotiated as a key exchange method, the parties MUST disconnect.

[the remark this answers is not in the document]
ecki: It’s not about your own code it’s about malformed peer messages

CHANGES.md
```diff
@@ -36,14 +36,38 @@
 
 ## Behavioral changes and enhancements
 
+### [GH-445 - Terrapin attack mitigation](https://github.com/apache/mina-sshd/issues/429)
+
+There is a **new** `CoreModuleProperties` property that controls the mitigation for the [Terrapin attack](https://terrapin-attack.com/) via what is known as
+"strict-KEX" (see [OpenSSH PROTOCOL - 1.9 transport: strict key exchange extension](https://github.com/openssh/openssh-portable/blob/master/PROTOCOL)).
+It is **disabled** by default due to its experimental nature and possible interoperability issues, so users who wish to use this feature must turn it on *explicitly*.
```
sberyozkin, line 43: @lgoldstein sorry for the noise as I could not open the earlier discussion. So, it is a security hardening fix and as such keeping such fixes disabled is not great IMHO, the affected cipher has been marked as deprecated in the list of built in parsers for a while AFAIK. 
In any case, please feel free to close this comment too, thanks
This PR is great otherwise.

tomaswolf:
I find this very hard to review.

1. That reformatting should be avoided. It indicates some problem with the configuration in your IDE or in your maven build. Please ensure that your IDE uses the same settings as the maven build, then such spurious reformatting should not happen.
2. There is completely unrelated stuff in this change. These things must not be in this change at all.
    - The command execution timeout looks suspicious to me. At least the timeout for running a command should by default be zero, as it was before.
    - There was some refactoring grouping together isolated variables related to KEX (SessionCountersDetails/SessionKexDetails). Probably not needed to implement strict KEX.
3. I don't think a custom SSH config property "UseStrictKex" is needed.
4. In fact I don't think any customization flag is needed at all. (Also not in CoreModuleProperties.)
5. This needs tests against `OpenSSH 9.6`. A container test once that OpenSSH version is available in some Linux repositories; until then, at least manual testing. We do have various container tests using older OpenSSH versions, so that bit of interoperability should already be covered.

As I understand it:

Both parties can unconditionally include the "strict kex" flag in their KEX_INIT message. "strict kex" is active if and only if both parties announce it in their initial KEX_INIT proposal. So both parties know whether strict kex is active once they have received the peer's proposal. If a party concludes that strict kex is active, it can:

- check that the receive sequence number of the peer's KEX_INIT is 1. If not, there were earlier messages, and they disconnect.
    - It does not seem necessary to actively avoid sending IGNORE/DEBUG messages before one's own KEX_INIT. Normally that shouldn't happen anyway. If it is needed, it could be done unconditionally (i.e., irrespective of whether strict kex will be active).
- as long as `initialKexDone == false`, only allow KEX messages. Reception of any other message causes the party to disconnect.
- When a party sends its own NEW_KEYS message, it resets the send sequence number to zero after having encoded the NEW_KEYS message itself. (So basically where it installs the new keys it also resets the message sequence counter.)
- When a party receives a NEW_KEYS message, it resets the receive sequence number to zero after having decoded the message. (Same here: where we install the new keys, we also reset the counter.)

Protocol-wise, this should be a fairly simple change. I would refrain from any beautification or not strictly needed refactoring; such things can be done in later changes once the basic modified protocol works. Probably unit tests would be the bulk of the change.

tomaswolf:
As you're a committer and PMC member I cannot stop you from merging this, but I fundamentally disagree with this change.

A PR should have one concern ("provide strict KEX"), and commits in a PR also should have one concern ("implement strict KEX functionality", "provide unit tests for strict KEX", "provide container tests (integration tests) for strict KEX").

Please note that OpenSSH 8.9 does **not** have the strict KEX feature. The first public version that does have it is OpenSSH 9.6. See the [OpenSSH release notes](https://www.openssh.com/releasenotes.html). Or did you build your own version of 8.9 with the strict KEX patch applied?

As for the reformatting: of course I don't _know_ how it happened... it looks to me as if your IDE reformatted the whole file to line length 80 on save. We normally use a line length of 120 characters.

Implementing strict KEX is about a 20, maybe 30-line change in AbstractSession. And then there may be hundreds of lines for tests. I'll try to put together a simpler PR over New Year.

ecki:
Debian bookworm backported the strictkex fix, so you could use that container (or maybe pinning Sid, it has the upstream fix, but that’s harder to keep available in the long run)

tomaswolf:
> > Please note that OpenSSH 8.9 does not have the strict KEX feature. The first public version that does have it is OpenSSH 9.6. See the [OpenSSH release notes](https://www.openssh.com/releasenotes.html). Or did you build your own version of 8.9 with the strict KEX patch applied?
> 
> I don't know what to tell you - when i run `sshd -V` it claims to be 8.9, and when I run the code it claims to have strict KEX... Maybe there is a subtle bug in my code - will try to look at it again

Seems like the Linux distro package maintainers have backported the strict kex patch also to older versions. For [Ubuntu](https://launchpad.net/ubuntu/+source/openssh/+changelog), I see it having been applied to OpenSSH versions 8.2, 8.9, 9.3, and in [debian](https://tracker.debian.org/pkg/openssh) for 8.4 and 9.2.

tomaswolf:
An alternate single-purpose PR that _only_ implements strict KEX and nothing else is available and ready for review now at #449. Not claiming it was "better", but it's much more focused. It includes interoperability tests with OpenSSH with and without strict KEX.

PR #459, opened 2024-01-26
Use a singleton threadpool for kex message handler flushing

### LINKED ISSUE #458: KeyExchangeMessageHandler threads pilling up
### Version

2.11.0

### Bug description

`KeyExchangeMessageHandler:89` creates an unnamed SingleThreadExecutor, which is kept open for each session.
In our use case we create connections to several thousand servers and keep them open.
Would it be possible to close this pool more eagerly?
If not or too complicated, a single instance pool, which ideally could be configured, would probably also work.

### Actual behavior

Threads stay alive

### Expected behavior

Pool is closed more eagerly

### Relevant log output

```Shell
2024-01-24 15:18:42,094 INFO  session.ClientSessionImpl: flushQueue(ClientSessionImpl[...:22]): 3 pending packets to flush [pool-3231-thread-1]
```

### Other information

_No response_

FliegenKLATSCH:
fixes #458
Threads will expire after 60 seconds.

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/common/session/helpers/KeyExchangeMessageHandler.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/session/helpers/KeyExchangeMessageHandler.java
@@ -28,7 +28,6 @@ import java.util.Objects;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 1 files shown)


tomaswolf:
While the task submitted in `flushQueue()` quits when a new KEX has started while flushing or when the session has been closed while flushing, I wonder if we now need something to prevent a second task being submitted by the same session before the first one has ended.

garydgregory:
Should the threads be daemon threads?

FliegenKLATSCH:
> While the task submitted in `flushQueue()` quits when a new KEX has started while flushing or when the session has been closed while flushing, I wonder if we now need something to prevent a second task being submitted by the same session before the first one has ended.

Good point, I'd tend to say no because of the locking in the loop.

> Should the threads be daemon threads?

Actually not really, but the same is probably true for other usages as well :/ Wondering if this would be a real world issue, since implementations would anyway wait for a response. Could rather be an issue on the server side.
Should we make it configurable? I am wondering if someone knows some good framework around thread pools to be used?

gnodet (approved):

PR #462, opened 2024-01-30
RAIN-4910 - Removed secret keys from old archived repositories.

akhileshpatil06:


ff-sdesai (approved):

ff-cviradiya (approved):

tomaswolf:
This is based on commit 95f75c4dd44 from 2016. In the current main branch, these files are in other directories and are used.

PR #472, opened 2024-03-11
change client start condition in sshd-spring-sftp

alwaystom:
The condition for starting the sshclient should be whether the client is started, rather than whether it is opened，


tomaswolf:
Thank you.

PR #475, opened 2024-03-14
Switch uses of JSch library to the com.github.mwiede:jsch fork

Alex-Vol-Amz:
The original Jcraft version of the JSch library has not seen updates in over a decade. The current fork with the most active maintenance is the com.github.mwiede:jsch library and it is largely compatible with the original but with secure defaults. It also absorbed the Jzlib implementation in its own code base and eliminated the need to have the extra dependencies.

As this is only a test dependency for this code base I feel adding this update is an improvement.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -504,14 +504,9 @@
             </dependency>
 
             <dependency>
-                <groupId>com.jcraft</groupId>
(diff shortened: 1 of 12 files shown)


tomaswolf:
Thank you.

PR #476, opened 2024-03-18
Fix Android detection false negative

wh0:
`p` is the property name like `java.vendor`, and this wasn't matching


tomaswolf:
Thank you.

PR #492, opened 2024-04-27
[SSHD-1237] Handle keep-alive channel requests

tomaswolf:
OpenSSH server sends keep-alive requests as channel requests if a channel is open. Handle these channel requests in the "last-resort" channel request handler.


gnodet:
> OpenSSH server sends keep-alive requests as channel requests if a channel is open. Do not log a warning about such requests, simply send back a failure reply. OpenSSH just expects any reply, whether success or failure.

Not displaying the warning looks weird. Wouldn’t it be cleaner to « support » them instead and reply with a success ?

tomaswolf:
> Not displaying the warning looks weird. Wouldn’t it be cleaner to « support » them instead and reply with a success or failure ?

Done.

PR #494, opened 2024-05-02
Nio2Session improvements

evgeny-pasynkov:
The proposed improvements for Nio2Session class are based on production experience and fixing some corner cases.

- When the buffer fails to be written, mark the corresponding Future as finished exceptionally
- When underlying network infrastructure splits buffer to multiple write chunks, the second and next chunks were written without respect to configured write timeout
- Minor consistency improvement: use better data structure for write queues which honours possibly concurrent access


tomaswolf:
Thank you, Evgeny! That looks great!

Issue #510, opened 2024-05-30
Fix mina full class name in BuiltinIoServiceFactoryFactories


JinHeap:
https://github.com/apache/mina-sshd/blob/37d239e792edabe2008df8e98bf00eef167a147f/sshd-core/src/main/java/org/apache/sshd/common/io/BuiltinIoServiceFactoryFactories.java#L37

full class name in BuiltinIoServiceFactoryFactories for mina is unavailable. I have to modified the code from   org.apache.sshd.common.io.mina.MinaServiceFactoryFactory to org.apache.sshd.mina.MinaServiceFactoryFactory in order to run SshClientMain with mina

> mina-sshd> java -cp * org.apache.sshd.cli.client.SshClientMain -w 1 root@127.0.0.1 -io mina
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
java.lang.RuntimeException: java.lang.ClassNotFoundException: org.apache.sshd.common.io.mina.MinaServiceFactoryFactory
        at org.apache.sshd.common.io.BuiltinIoServiceFactoryFactories.getFactoryClass(BuiltinIoServiceFactoryFactories.java:76)
        at org.apache.sshd.common.io.BuiltinIoServiceFactoryFactories.create(BuiltinIoServiceFactoryFactories.java:88)
        at org.apache.sshd.cli.CliSupport.setupIoServiceFactory(CliSupport.java:149)
        at org.apache.sshd.cli.client.SshClientCliSupport.setupDefaultClient(SshClientCliSupport.java:407)
        at org.apache.sshd.cli.client.SshClientCliSupport.setupClient(SshClientCliSupport.java:447)
        at org.apache.sshd.cli.client.SshClientCliSupport.setupClientSession(SshClientCliSupport.java:244)
        at org.apache.sshd.cli.client.SshClientMain.main(SshClientMain.java:134)
Caused by: java.lang.ClassNotFoundException: org.apache.sshd.common.io.mina.MinaServiceFactoryFactory
        at java.net.URLClassLoader.findClass(Unknown Source)
        at java.lang.ClassLoader.loadClass(Unknown Source)
        at sun.misc.Launcher$AppClassLoader.loadClass(Unknown Source)
        at java.lang.ClassLoader.loadClass(Unknown Source)
[...]

PR #511, opened 2024-05-30
Fix mina full class name in BuiltinIoServiceFactoryFactories

JinHeap:
Addresses https://github.com/apache/mina-sshd/issues/510


tomaswolf:
Thanks. The change is good, but now that this is fixed `DefaultIoServiceFactoryFactoryTest` hangs for MINA during disposal of the partially set up (with mocks) `MinaServiceFactory`. Previously it just skipped the non-existing class.

Also, `BuiltinIoServiceFactoryFactoriesTest` should have a test that fails if one of the enum constants has a wrong class name. I.e., a test that would have uncovered the original problem (wrong string).

I'll see if I can amend this commit to fix these test issues.

JinHeap:
thanks. then if need me do something, please tell me.

tomaswolf:
Merged as commit 82e0ed1f.

PR #537, opened 2024-07-23
Require JDK 17 at build time (fixes #536)

### LINKED ISSUE #536: Require JDK 17 for building
### Description

Bump the minimal JDK version required for building to java 17.

### Motivation

This will streamline the build, allow upgrading to more recent plugins.
The minimal JDK requirement for runtime will be unchanged at JDK 8.

gnodet:
Fixes #536

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,18 +79,15 @@
     </scm>
 
     <properties>
-        <projectRoot>${project.basedir}</projectRoot>
+        <minimalJavaBuildVersion>17</minimalJavaBuildVersion>
+        <jdkToolchainVersion>[${minimalJavaBuildVersion},)</jdkToolchainVersion>
+        <minimalMavenBuildVersion>3.9.8</minimalMavenBuildVersion>
+
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.build.outputTimestamp>2024-06-20T20:44:27Z</project.build.outputTimestamp>
 
-        <!-- NOTE: actual workspace root dir value updated by gmavenplus-plugin on initialization phase -->
-        <!-- workspace.root.dir>${project.basedir}</workspace.root.dir -->
-
-        <java.major.version>1</java.major.version>
-        <java.minor.version>8</java.minor.version>
         <java.sdk.version>8</java.sdk.version>
-        <sniffer.signatures.version>1.0</sniffer.signatures.version>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -137,10 +134,38 @@
...
+                            <jdkToolchain>
+                                <version>${jdkToolchainVersion}</version>
+                            </jdkToolchain>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+
+        <profile>
```
tomaswolf, line 158: What are these profiles for? They don't seem to be needed.

gnodet: The only purpose is the following line:
 https://github.com/apache/mina-sshd/blob/38c19dc435a875e3cd98f17422f1441559632d28/sshd-ldap/pom.xml#L150
 
 LDAP testing requires [JVM specific args](https://github.com/apache/mina-sshd/blob/38c19dc435a875e3cd98f17422f1441559632d28/sshd-ldap/pom.xml#L36), but those are only available on JDK >= 11.  We can't use a simple profile activation, as those can only be activated based on the runtime JDK, not on the toolchain selected one.  Property based activation is also limited to presence or absence of a property. 

 But I think it should be possible to get rid of those by creating a property, let me simplify it.

tomaswolf:
@gnodet: snapshot deployment fails with
> Error:  Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.3.0:check (verify-style) on project sshd: Failed during checkstyle execution: Unable to find configuration file at location: ${session.rootDirectory}/sshd-checkstyle.xml: Could not find resource '${session.rootDirectory}/sshd-checkstyle.xml'. -> [Help 1]

gnodet:
> @gnodet: snapshot deployment fails with
> 
> > Error:  Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.3.0:check (verify-style) on project sshd: Failed during checkstyle execution: Unable to find configuration file at location: ${session.rootDirectory}/sshd-checkstyle.xml: Could not find resource '${session.rootDirectory}/sshd-checkstyle.xml'. -> [Help 1]

It requires maven 3.9.x.  Where is the job configured ?

tomaswolf:
Deployment is in master-build.yaml. Does that need the extra maven-setup step now, too? Where is that session.rootDirectory coming from?

PR #538, opened 2024-07-25
Add dependabot

gnodet:
I'd like to add dependabot to the repo, @tomaswolf any problem with that ?


tomaswolf:
No. We'll see how it goes. But we used to get dependabot updates before without this; for instance #457.

garydgregory:
I would schedule it weekly personally, to make it a bit less noisy.

PR #546, opened 2024-07-25
Switch to JUnit 5

gnodet:


tomaswolf:
It appears that with JUnit5 there needs to be at least one test class in the project itself. sshd-netty has all tests in the sshd-core dependency. sshd-mina worked fine, but sshd-mina does have one test class.

sshd-common/src/test/java/org/apache/sshd/client/auth/password/PasswordIdentityProviderTest.java
```diff
@@ -69,7 +70,7 @@ public void testMultiProvider() throws IOException, GeneralSecurityException {
 
     private static void assertProviderContents(String message, PasswordIdentityProvider p, Iterable<String> expected)
             throws IOException, GeneralSecurityException {
-        assertNotNull(message + ": no provider", p);
+        assertNotNull(p, message + ": no provider");
         assertEquals(message, expected, p.loadPasswords(null));
```
tomaswolf, line 74: Shouldn't the message be at the end here?

tomaswolf: Oh no, wait. This is calling our own assertEquals!

tomaswolf:
Rebased, squashed, migrated the one test added in the meantime from JUnit4 to JUnit5. Works well building locally for me.

If the CI build succeeds, I'll merge this. Before other new tests get added based on JUnit4.

garydgregory:
Nice!

PR #552, opened 2024-07-26
Test with jdk 21

gnodet:


tomaswolf:
sshd-ldap is failing. Apache DS appears to be a dead end; see https://github.com/apache/cxf/pull/1042 .

PR #565, opened 2024-07-26
[GH-539] Implement no-flow-control extension

### LINKED ISSUE #539: Implement  no-flow-control extension
### Description

Implement support for the "no-flow-control" SSH extension, per [RFC 8308](https://tools.ietf.org/html/rfc8308#section-3.3).

### Motivation

This may bring some improvements on networks with high latency.

### Alternatives considered

Growing the window size is an alternative, but requires a global modification on the server side and can not only affect file transfers.

### Additional context

_No response_

gnodet:
- **Support for new client side extension**
- **Implement no-flow-control extension (fixes #539)**
- **Minor tweaks**


sshd-core/src/main/java/org/apache/sshd/common/kex/extension/DefaultServerKexExtensionHandler.java
```diff
@@ -130,6 +136,23 @@ public void sendKexExtensions(Session session, KexPhase phase) throws Exception
...
+        if (NoFlowControl.NAME.equals(name)) {
+            String o = NoFlowControl.INSTANCE.parseExtension(data);
+            Optional<Boolean> nfc = CoreModuleProperties.NO_FLOW_CONTROL.get(session);
+            if (NoFlowControl.PREFERRED.equals(o) && nfc.orElse(Boolean.TRUE)
+                    || NoFlowControl.SUPPORTED.equals(o) && nfc.orElse(Boolean.FALSE)) {
+                AbstractSession abstractSession
+                        = ValidateUtils.checkInstanceOf(session, AbstractSession.class, "Not a supported session: %s", session);
+                abstractSession.activateNoFlowControl();
+            }
+        }
```
tomaswolf, line 152: Can we avoid this code duplication between client & server?

tomaswolf (commented):
Several findings, see inline comments.

In general I wonder how much this brings in terms of throughput. The point of it all is to avoid that the channel remote window ever gets smaller than the packet size. I think we already have that. A connection would need to have a really huge latency to have the peer's window adjustments arrive so late that a sender would be blocked by a small channel window. Or the receiver would need to be much slower in processing data received than the sender sending it. In either case a sender would be blocked probably anyway by TCP/IP send or receive buffers being full.

How could we do an integration test? Is there a publicly available third-party SSH implementation besides OpenSSH that we could run in a container? AFAIK OpenSSH does not implement this extension.

sshd-core/src/main/java/org/apache/sshd/common/session/helpers/KeyExchangeMessageHandler.java
```diff
@@ -236,7 +237,7 @@ public IoWriteFuture writePacket(Buffer buffer, long timeout, TimeUnit unit) thr
         int cmd = bufData[buffer.rpos()] & 0xFF;
         boolean enqueued = false;
         boolean isLowLevelMessage = cmd <= SshConstants.SSH_MSG_KEX_LAST && cmd != SshConstants.SSH_MSG_SERVICE_REQUEST
-                && cmd != SshConstants.SSH_MSG_SERVICE_ACCEPT;
+                && cmd != SshConstants.SSH_MSG_SERVICE_ACCEPT && cmd != KexExtensions.SSH_MSG_EXT_INFO;
```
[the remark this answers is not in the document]
tomaswolf: Crap. I misread the code.

sshd-core/src/main/java/org/apache/sshd/common/kex/extension/DefaultClientKexExtensionHandler.java
```diff
@@ -133,4 +151,39 @@ protected void handleServerSignatureAlgorithms(Session session, Collection<Strin
             session.setSignatureFactories(clientAlgorithms);
         }
     }
+
+    @Override
+    public void sendKexExtensions(Session session, KexPhase phase) throws Exception {
```
tomaswolf, line 156: This must send something only if the server had announced ext-info-s. Otherwise the client must not send any extensions. See the logic in the DefaultServerKexExtensionHandler.

PR #580, opened 2024-08-09
Add Android-specific exception message when finding user home

ShellWen:
In the Android world, there is no concept similar to a home directory.
That is to say `System.getProperty("user.home")` is always `null`.
This PR can guide developer who target Android to add their own home directory implementation by modifying the exception message.


tomaswolf:
Needs formatting (see CI build output).

ShellWen:
Sorry for mistake. Please check again. 😉

tomaswolf:
Still fails. The order of imports is wrong.

If you run a maven build locally it should format as required. Then amend your commit and force push.

tomaswolf:
Thank you!

PR #583, opened 2024-08-14
GH-582: Fix filtering of unsupported algorithms in NamedFactory

### LINKED ISSUE #582: NamedFactory uses wrong filter expression
### Version

2.14.0-SNAPSHOT

### Bug description

Split off from #525:

The implementations of `NamedFactory.setupBuiltinFactories()` and `NamedFactory.setupTransformedFactories()` use incorrect filter expressions to filter out algorithms not supported by the platform.

They use the expression `ignoreUnsupported || f.isSupported()` and expect that to be `true` for supported factories, and `false` for unsupported ones.

However, that is not correct. The parameter `ignoreUnsupported` actually implements the semantics of "include unsupported".

This needs to be fixed. The correct expression would be `ignoreUnsupported ? f.isSupported() : true`, or `!ignoreUnsupported || f.isSupported()`.

This bug exists in the code since 2015.

### Actual behavior

`NamedFactory.setupBuiltinFactories(true, ...)` includes factories for unsupported algorithms in the returned list.

### Expected behavior

`NamedFactory.setupBuiltinFactories(true, ...)` includes only factories for supported algorithms in the returned list.

### Relevant log output

_No response_

### Other information

_No response_

tomaswolf:
Also give the methods some javadoc.

Note that `ignoreUnsupported ? f.isSupported : true` is the same as `!ignoreUnsupported || f.isSupported()`, but I find the former easier to read and understand.

Fixes #582.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -40,6 +40,7 @@
 
 * [GH-524](https://github.com/apache/mina-sshd/issues/524) Performance improvements
 * [GH-533](https://github.com/apache/mina-sshd/issues/533) Fix multi-step authentication
+* [GH-582](https://github.com/apache/mina-sshd/issues/582) Fix filtering in `NamedFactory`
 
 ## New Features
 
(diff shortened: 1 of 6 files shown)


quic-nasserg:
Thank you Thomas!

PR #586, opened 2024-08-26
Fix license name

kelunik:
See https://spdx.org/licenses/Apache-2.0.html.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,7 @@
 
     <licenses>
         <license>
-            <name>Apache 2.0 License</name>
+            <name>Apache License 2.0</name>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -34,7 +34,7 @@
 
     <licenses>
         <license>
-            <name>Apache 2.0 License</name>
+            <name>Apache License 2.0</name>
```
gnodet, line 37: I'd use `Apache-2.0` to be consistent with the ASF parent POM.
However, I think the `<licenses>` tag should be removed completely, as it's already inherited from the ASF parent.  This must be a remaining artifact from before the move to the ASF.

gnodet (changes requested):
Remove tag completely.

Issue #590, opened 2024-08-31
Cannot use Mina SSH client in a FIPS constrained environment


olamy:
### Version

2.13.2

### Bug description

When using using registar created for BCFIPS (which means any BC classes are available within the class path).

I got

```
java.lang.IllegalArgumentException: BouncyCastle not registered
  at PluginClassLoader for mina-sshd-api-common//org.apache.sshd.common.util.ValidateUtils.createFormattedException(ValidateUtils.java:213)
  at PluginClassLoader for mina-sshd-api-common//org.apache.sshd.common.util.ValidateUtils.throwIllegalArgumentException(ValidateUtils.java:179)
  at PluginClassLoader for mina-sshd-api-common//org.apache.sshd.common.util.ValidateUtils.checkTrue(ValidateUtils.java:156)
  at PluginClassLoader for mina-sshd-api-common//org.apache.sshd.common.util.security.SecurityUtils.getBouncycastleEncryptedPrivateKeyInfoDecryptor(SecurityUtils.java:553)
  at PluginClassLoader for mina-sshd-api-common//org.apache.sshd.common.config.keys.loader.pem.PKCS8PEMResourceKeyPairParser.decryptKeyPairs(PKCS8PEMResourceKeyPairParser.java:107)
```

I tried to "trick" it :) using the name BC here https://github.com/jenkinsci/mina-sshd-api-plugin/pull/114/files#diff-5440105bdcdf53b86acce84166b9884f497eb6908da1d68b82ec974aa0fd83e1R45

But turns into:

```
  Caused: java.lang.NoClassDefFoundError: org/bouncycastle/crypto/prng/RandomGenerator
  at PluginClassLoader for mina-sshd-api-common//org.apache.sshd.common.util.security.bouncycastle.BouncyCastleRandomFactory.create(BouncyCastleRandomFactory.java:43)
  at PluginClassLoader for mina-sshd-api-common//org.apache.sshd.common.util.security.bouncycastle.BouncyCastleRandomFactory.create(BouncyCastleRandomFactory.java:28)
  at PluginClassLoader for mina-sshd-api-common//org.apache.sshd.common.random.SingletonRandomFactory.<init>(SingletonRandomFactory.java:38)
with BCFIPS (classic BC class not available in the classpth) this is this class https://javadoc.io/doc/org.bouncycastle/bc-fips/latest/org/bouncycastle/crypto/fips/FipsSecureRandom.html
```
But the idea is to make this random factory more configurable.

[...]

PR #591, opened 2024-08-31
GH-590: Be more careful with Bouncy Castle in FIPS environment

tomaswolf:
Decouple the registrar name from the security provider name. Also check whether the BC RandomGenerator exists at all; in BC-FIPS, it doesn't.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -41,24 +41,47 @@
 * [GH-524](https://github.com/apache/mina-sshd/issues/524) Performance improvements
 * [GH-533](https://github.com/apache/mina-sshd/issues/533) Fix multi-step authentication
 * [GH-582](https://github.com/apache/mina-sshd/issues/582) Fix filtering in `NamedFactory`
+* [GH-590](https://github.com/apache/mina-sshd/issues/590) Better support for FIPS
 
 ## New Features
 
 * New utility methods `SftpClient.put(Path localFile, String remoteFileName)` and
 `SftpClient.put(InputStream in, String remoteFileName)` facilitate SFTP file uploading.
 
+### [GH-590](https://github.com/apache/mina-sshd/issues/590) Better support for FIPS
+
+Besides fixing a bug with bc-fips (the `RandomGenerator` class exists in normal Bouncy Castle,
+but not in the FIPS version, but Apache MINA sshd referenced it even if only bc-fips was present), 
+support was improved for running in an environment restricted by FIPS.
+
+There is a new system property `org.apache.sshd.security.fipsEnabled`. If set to `true`, a number
+of crypto-algorithms not approved by FIPS 140 are disabled:
+
+* key exchange methods sntrup761x25519-sha512, sntrup761x25519-sha512<!-- -->@openssh.com, curve25519-sha256, curve25519-sha256<!-- -->@libssh.org, curve448-sha512.
+* the chacha20-poly1305 cipher.
+* the bcrypt KDF used in encrypted private key files in [OpenSSH format](https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.key).
+* all ed25519 keys and signatures.
+
+Additionally, the new "SunJCEWrapper" `SecurityProviderRegistrar` (see below) and the
+`EdDSASecurityProviderRegistrar` are disabled, and the `BouncyCastleScurityProviderRegistrar`
+looks only for the "BCFIPS" security provider, not for the normal "BC" provider.
+
+If the system property is _not_ set to `true`, FIPS mode can be enabled programmatically
+by calling `SecurityUtils.setFipsMode()` before any other call to Apache MINA sshd.
+
 ## Potential compatibility issues
 
 ### New security provider registrar
 There is a new `SecurityProviderRegistrar` that is registered by default
-if there is a `SunJCE` security provider and that uses the AES and
+if there is a `SunJCE` security provider. It uses the AES and
 HmacSHA* implementations from `SunJCE` even if Bouncy Castle is also
 registered. `SunJCE` has native implementations, whereas Bouncy Castle
 may not.
 
 The new registrar has the name "SunJCEWrapper" and can be configured
 like any other registrar. It can be disabled via the system property
-`org.apache.sshd.security.provider.SunJCEWrapper.enabled=false`.
+`org.apache.sshd.security.provider.SunJCEWrapper.enabled=false`. It is also
+disabled in FIPS mode (see above).
 
 ### [GH-582](https://github.com/apache/mina-sshd/issues/582) Fix filtering in `NamedFactory`
 
--- a/docs/security-providers.md
+++ b/docs/security-providers.md
@@ -107,3 +107,30 @@ In any case, the values are auto-detected by the code but the user can intervene
 * Setting a value of zero indicates a **lazy** auto-detection of the supported range the next time these values are needed.
 
 Furthermore, if all possible primes have been exhausted the code no longer falls back to DH group exchange using SHA-1 unless the `ALLOW_DHG1_KEX_FALLBACK` core module property is set.
+
+## FIPS Mode
+
+Apache MINA sshd supports running in environments restricted by FIPS 140 through a special "FIPS mode".
+In FIPS mode, crypto-algorithms not approved by FIPS-140 are disabled.
+
+FIPS mode can be switched on in two ways:
+* via System property `org.apache.sshd.security.fipsEnabled=true`, or
+* if the system property is not set to `true` also programmatically via a call to `SecurityUtils.setFipsMode()` before any other call to the library.
+
+In FIPS mode, the following algorithms are disabled completely:
+* key exchange methods sntrup761x25519-sha512, sntrup761x25519-sha512<!-- -->@openssh.com, curve25519-sha256, curve25519-sha256<!-- -->@libssh.org, and curve448-sha512.
+* the chacha20-poly1305 cipher.
+* the bcrypt KDF used in encrypted private key files in [OpenSSH format](https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.key).
+* all ed25519 keys and signatures.
+
+The `BouncyCastleSecurityProviderRegistrar` only considers the "BCFIPS" provider from `bc-fips`
+and disregards the "BC" provider from normal Bouncy Castle. The registrar for _EdDSA_ is disabled.
+For random numbers, `SecureRandom.getInstanceStrong()` is used.
+
+This FIPS mode does not automatically make an application using Apache MINA sshd FIPS-140-compliant.
+It only ensures that the Apache MINA sshd library does not provide or use algorithms that should not
+be used in FIPS-140-compliant applications running in approved mode.
+
+It is usually necessary to configure the JVM for full FIPS 140 compliance, for instance, by defining
+appropriate security providers statically. Any such configuration is beyond the scope of the Apache MINA
+sshd library.
--- a/sshd-common/src/main/java/org/apache/sshd/common/cipher/BuiltinCiphers.java
+++ b/sshd-common/src/main/java/org/apache/sshd/common/cipher/BuiltinCiphers.java
@@ -31,6 +31,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeMap;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -39,6 +40,7 @@ import org.apache.sshd.common.NamedResource;
 import org.apache.sshd.common.config.NamedFactoriesListParseResult;
 import org.apache.sshd.common.util.GenericUtils;
 import org.apache.sshd.common.util.ValidateUtils;
+import org.apache.sshd.common.util.security.SecurityUtils;
 
 /**
  * Provides easy access to the currently implemented ciphers
@@ -50,6 +52,11 @@ public enum BuiltinCiphers implements CipherFactory {
         @Override
         public Cipher create() {
             return new CipherNone();
+        }
+
+        @Override
+        public boolean isSupported() {
+            return !SecurityUtils.isFipsMode();
         }
     },
     aes128cbc(Constants.AES128_CBC, 16, 0, 16, "AES", 128, "AES/CBC/NoPadding", 16) {
@@ -149,6 +156,11 @@ public enum BuiltinCiphers implements CipherFactory {
         public Cipher create() {
             return new ChaCha20Cipher();
         }
+
+        @Override
+        public boolean isSupported() {
+            return !SecurityUtils.isFipsMode();
+        }
     },
     /**
      * @deprecated
@@ -175,7 +187,7 @@ public enum BuiltinCiphers implements CipherFactory {
     private final int blkSize;
     private final String algorithm;
     private final String transformation;
-    private final boolean supported;
+    private final AtomicReference<Boolean> supported = new AtomicReference<>();
 
     BuiltinCiphers(
                    String factoryName, int ivsize, int authSize, int kdfSize,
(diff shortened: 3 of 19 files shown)


olamy:
Thanks for the PR but hardcoded "BC" gives me this:

```
java.lang.IllegalArgumentException: BouncyCastle not registered
	at PluginClassLoader for mina-sshd-api-common//org.apache.sshd.common.util.ValidateUtils.createFormattedException(ValidateUtils.java:213)
	at PluginClassLoader for mina-sshd-api-common//org.apache.sshd.common.util.ValidateUtils.throwIllegalArgumentException(ValidateUtils.java:179)
	at PluginClassLoader for mina-sshd-api-common//org.apache.sshd.common.util.ValidateUtils.checkTrue(ValidateUtils.java:156)
	at PluginClassLoader for mina-sshd-api-common//org.apache.sshd.common.util.security.SecurityUtils.getBouncycastleEncryptedPrivateKeyInfoDecryptor(SecurityUtils.java:546)
	at PluginClassLoader for mina-sshd-api-common//org.apache.sshd.common.config.keys.loader.pem.PKCS8PEMResourceKeyPairParser.decryptKeyPairs(PKCS8PEMResourceKeyPairParser.java:107)
```

olamy:
as long as there is this check https://github.com/apache/mina-sshd/blob/40a08a4fc4b469185b1dd4f7c90cdf8949dc0c80/sshd-common/src/main/java/org/apache/sshd/common/util/security/SecurityUtils.java#L396 still using the hardcoded `BC` this will not work

olamy:
There is a test for this with this PR https://github.com/jenkinsci/mina-sshd-api-plugin/pull/114

tomaswolf:
The exception you show above shows that you are trying to load an encrypted PEM key. (Starts with "-----BEGIN ENCRYPTED PRIVATE KEY-----").

AFAIK that is not possible in a FIPS environments as FIPS has no PBE/PBES2.

So the exception is totally correct.

olamy:
Ah. Do you have any link for this?

olamy:
TBI have been always thinking this `openssl genrsa -des3 -out rsa2048 rsa2048 ` would generate a FIPS compatible key.

tomaswolf:
Hm. Actually, I had a link, but can't find it anymore. Anyway, it said bc-fips had no pbe... (not that FIPS didn't have it). Might be technically true since I don't see any such code in bundle bc-fips, but there's also bundle bcpkix-fips which seems to have the classes I was missing. I'll keep trying.

olamy:
that's failing my tests here https://github.com/jenkinsci/mina-sshd-api-plugin/pull/114/

tomaswolf:
@olamy: All right. The original idea _was_ sound after all. With this PR at commit 9518312 and the following patch applied to your that's failing my tests here https://github.com/jenkinsci/mina-sshd-api-plugin/pull/114 your test succeeds.

Patch:
```
 public class FIPSBouncyCastleSecurityProviderRegistar extends AbstractSecurityProviderRegistrar {
 
     public FIPSBouncyCastleSecurityProviderRegistar() {
-        super("BCFIPS");
+        super("BC");
+    }
+
+    @Override
+    public String getProviderName() {
+        return "BCFIPS";
     }
 
     @Override
```

Test run succeeds:
```
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 7.665 s -- in io.jenkins.plugins.mina_sshd_api.core.bouncycaslte_registar.TestBCFIPSRegistar
```

Note that I did not include the customizeable RandomFactory. Not sure it's needed; it'll use `JceRandom` with BCFIPS anyway. (And JceRandom is changed to use `SecureRandom.getInstanceStrong()`.)

Two observations:

1. The typo in the package name I mentioned before is io.jenkins.plugins.mina_sshd_api.core.bouncyca**slte**_registar. That should probably be "bouncyca**stl**e_registrar".
2. I think typical applications will include either the BCFIPS bundles _or_ the normal BC bundles, but not both. Mix and match won't work anyway. In such typical applications, there shouldn't be a need for an extra registrar at all, since the built-in `BouncyCastleSecurityRegistrar` can handle either, depending on what is accessible or what security providers are installed already. I presume no application would have both the BCFIPS _and_ the BC provider installed.

Your use case seems to be in Jenkins and apparently both libraries (BC and BCFIPS) may be present. That use case may still need its special-purpose registrar. Also you mentioned you'd not want to use the SunJCE AES, so a system property to disable the SunJCEWrapper registrar would be needed anyway if you don't override the registrars. (If you do -- as is currently the case -- that SunJCEWrapper won't be registered at all.)

tomaswolf:
And in `FIPSRegistarInitPlugin` two lines can be removed:
```
         if(FIPS140.useCompliantAlgorithms() && Security.getProvider("BCFIPS") != null) {
             LOG.info("register FIPSBouncyCastleSecurityProviderRegistar");
             System.setProperty("org.apache.sshd.security.registrars", "io.jenkins.plugins.mina_sshd_api.core.bouncycaslte_registar.FIPSBouncyCastleSecurityProviderRegistar");
-            System.setProperty("org.apache.sshd.security.defaultProvider", "BCFIPS");
-            System.setProperty("org.apache.sshd.common.util.security.bouncycastle.provider.name", "BCFIPS");
         } else {
             LOG.config("not needed to register FIPSBouncyCastleSecurityProviderRegistar");
         }
```

olamy:
@tomaswolf Thanks a lot!!

> @olamy: All right. The original idea _was_ sound after all. With this PR at commit [9518312](https://github.com/apache/mina-sshd/commit/9518312b76aa4552343c56ad58b54e0ebb0b4950) and the following patch applied to your test [jenkinsci/mina-sshd-api-plugin#114](https://github.com/jenkinsci/mina-sshd-api-plugin/pull/114) that test succeeds.
> 
> Patch:
> 
> ```
>  public class FIPSBouncyCastleSecurityProviderRegistar extends AbstractSecurityProviderRegistrar {
>  
>      public FIPSBouncyCastleSecurityProviderRegistar() {
> -        super("BCFIPS");
> +        super("BC");
> +    }
> +
> +    @Override
> +    public String getProviderName() {
> +        return "BCFIPS";
>      }
>  
>      @Override
> ```
> 
> Test run succeeds:
> 
> ```
> [INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 7.665 s -- in io.jenkins.plugins.mina_sshd_api.core.bouncycaslte_registar.TestBCFIPSRegistar
> ```
> 
> Note that I did not include the customizeable RandomFactory. Not sure it's needed; it'll use `JceRandom` with BCFIPS anyway. (And JceRandom is changed to use `SecureRandom.getInstanceStrong()`.)
> 

Right so it's not needed anymore.

> Two observations:
> 
> 1. The typo in the package name I mentioned before is io.jenkins.plugins.mina_sshd_api.core.bouncyca**slte**_registar. That should probably be "bouncyca**stl**e_registrar".

Haha good catch. I was checking if String constants in System.setProperty were right but the problem was something else. Uhm I need some new glasses... 

> 2. I think typical applications will include either the BCFIPS bundles _or_ the normal BC bundles, but not both. Mix and match won't work anyway. In such typical applications, there shouldn't be a need for an extra registrar at all, since the built-in `BouncyCastleSecurityRegistrar` can handle either, depending on what is accessible or what security providers are installed already. I presume no application would have both the BCFIPS _and_ the BC provider installed.
> 
> Your use case seems to be in Jenkins and apparently both libraries (BC and BCFIPS) may be present. That use case may still need its special-purpose registrar. Also you mentioned you'd not want to use the SunJCE AES, so a system property to disable the SunJCEWrapper registrar would be needed anyway if you don't override the registrars. (If you do -- as is currently the case -- that SunJCEWrapper won't be registered at all.)

No we don't. If Jenkins is running in FIPS mode, the classic BC is not registered (not easy to figure when looking at the poms  only but it's here https://github.com/jenkinsci/bouncycastle-api-plugin/blob/e27176eb46cbe94c1d1de6a9f318e3d538c34a4a/src/main/java/jenkins/bouncycastle/api/BouncyCastlePlugin.java#L27)

olamy:
> And in `FIPSRegistarInitPlugin` two lines can be removed:
> 
> ```
>          if(FIPS140.useCompliantAlgorithms() && Security.getProvider("BCFIPS") != null) {
>              LOG.info("register FIPSBouncyCastleSecurityProviderRegistar");
>              System.setProperty("org.apache.sshd.security.registrars", "io.jenkins.plugins.mina_sshd_api.core.bouncycaslte_registar.FIPSBouncyCastleSecurityProviderRegistar");
> -            System.setProperty("org.apache.sshd.security.defaultProvider", "BCFIPS");
this one ^ is still needed because we want to use BCFIPS.
> -            System.setProperty("org.apache.sshd.common.util.security.bouncycastle.provider.name", "BCFIPS");
>          } else {
>              LOG.config("not needed to register FIPSBouncyCastleSecurityProviderRegistar");
>          }
> ```


Running few tests and this is looking really great. Thanks for your help.

tomaswolf:
> this one ^ is still needed because we want to use BCFIPS.

Right.

However, for FIPS mode I think we may have another problem: we use our own ChaCha20-Poly1305 implementation, and we use the bcyrypt KDF for encrypted OpenSSH "new format" keys.

Probably we should have a way to disable those if you want to run in FIPS mode?

olamy:
> > this one ^ is still needed because we want to use BCFIPS.
> 
> Right.
> 
> However, for FIPS mode I think we may have another problem: we use our own ChaCha20-Poly1305 implementation, and we use the bcyrypt KDF for encrypted OpenSSH "new format" keys.
> 

right as far as I understand ` ChaCha20-Poly1305` is not FIPS compliant. 

Regarding `OpenSSH "new format" keys`, I'm not quite sure.
 

> Probably we should have a way to disable those if you want to run in FIPS mode?

tomaswolf:
> I'm not quite sure. with new do you mean this from 2018? :) https://www.openssh.com/txt/release-7.8

Yes: https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.key

I also think neither ChaCha20-Poly1305 nor bcrypt are covered by FIPS.

For ChaCha20-Poly1305 we have two options: either we say it's the user's responsibility to configure the SshClient appropriately by removing the cipher via configuration (`SshClient.setCipherFactories()`), or we add a flag (with system property and setter) in `SecurityUtils` and then hard-disable this Cipher if the flag is set.

The bcrypt thing is less critical. I don't know which KDF's are FIPS-approved, but I'd be surprised if bcrypt was. If not, anyone using this OpenSSH format for stored keys would deviate from FIPS anyway, so we could say it's entirely the user's responsibility. (As is using only PEM keys with FIPS-approved encryptions and KDFs.) Or we could use the aforementioned flag to also disable bcrypt (with the effect that such keys cannot be read or written in FIPS mode). Or we might only disable writing such files, while still allowing to read them.

jtnord:
> The bcrypt thing is less critical. I don't know which KDF's are FIPS-approved, but I'd be surprised if bcrypt was.

for BouncyCastle FIPS see section 6 of https://downloads.bouncycastle.org/fips-java/docs/BC-FJA-UserGuide-1.0.2.pdf

olamy:
> > I'm not quite sure. with new do you mean this from 2018? :) https://www.openssh.com/txt/release-7.8
> 
> Yes: https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.key
> 
> I also think neither ChaCha20-Poly1305 nor bcrypt are covered by FIPS.
> 
> For ChaCha20-Poly1305 we have two options: either we say it's the user's responsibility to configure the SshClient appropriately by removing the cipher via configuration (`SshClient.setCipherFactories()`), or we add a flag (with system property and setter) in `SecurityUtils` and then hard-disable this Cipher if the flag is set.
>

LGTM
 
> The bcrypt thing is less critical. I don't know which KDF's are FIPS-approved, but I'd be surprised if bcrypt was. If not, anyone using this OpenSSH format for stored keys would deviate from FIPS anyway, so we could say it's entirely the user's responsibility. (As is using only PEM keys with FIPS-approved encryptions and KDFs.) Or we could use the aforementioned flag to also disable bcrypt (with the effect that such keys cannot be read or written in FIPS mode). Or we might only disable writing such files, while still allowing to read them.

what about something such a general flag `fips enabled`? Will turn reponsibility to the library but could be a great user feature?

tomaswolf:
So here's another commit on top of this PR for this single flag. See the commit message.

With this, I think you don't need the `FIPSBouncyCastleSecurityProviderRegistar` anymore, and the `FIPSRegistarInitPlugin` can just do
```
System.setProperty("org.apache.sshd.security.fipsEnabled", "true");
System.setProperty("org.apache.sshd.security.defaultProvider", "BCFIPS");
```

If the flag is set, a number of non-FIPS-approved crypto-algorithms get disabled (ed25519, curve25519, curve448, sntrup761, bcrypt, chacha20-poly1305).

Disabling sntrup761 and curve25519 means that this FIPS mode is left without any post-quantum key exchange method.
[FIPS 203](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.pdf) has come out recently; it specifies the post-quantum ML-KEM key encapsulation method. There is a [draft RFC](https://datatracker.ietf.org/doc/html/draft-kampanakis-curdle-ssh-pq-ke-04) for incorporating this into SSH as key exchange algorithm. Once OpenSSH has support for this and Bouncy Castle also provides these ML-KEM instances, we can add them easily. At that point there would be mlkem1024nistp384-sha384 available as (I believe) FIPS-compliant post-quantum algorithm. When we do have that algorithm, I think we should keep it enabled even in FIPS mode.

Can you test that this works for your use case in Jenkins?

olamy:
@tomaswolf  Thanks a lot. I will test this.

olamy:
@tomaswolf I did some successful testing. 
But I will still need the registar because the environment where it's running is very restrictive and even have a SecurityManager installed to block registration of SecurityProvider. 
So without using the registar, the Apache Mina is trying to register the `BCFIPS` and `BC` :) 

```
Caused: java.lang.IllegalArgumentException: Failed to register BC as a JCE provider
  at PluginClassLoader for mina-sshd-api-common//org.apache.sshd.common.util.security.SecurityUtils.registerSecurityProvider(SecurityUtils.java:534)
  at PluginClassLoader for mina-sshd-api-common//org.apache.sshd.common.util.security.SecurityUtils.register(SecurityUtils.java:489)
  at PluginClassLoader for mina-sshd-api-common//org.apache.sshd.common.util.security.SecurityUtils.isBouncyCastleRegistered(SecurityUtils.java:438)
```

tomaswolf:
That's strange. I did test this with your change, and it worked like a charm.

Your plug-in sets FIPS mode only if the "BCFIPS" security provider is already registered. So this code should never try to add the BCFIPS provider at run-time. Somehow it looks as if the BCFIPS security provider is not present in your environment.

tomaswolf:
In other words: you must have run with the flag set to `true`, and the bc-fips libraries present, but without the BCFIPS `java.security.Provider` being installed. Only then would the code try to install the provider itself.

olamy:
Yes, correct. 
I probably mix up some local build of snapshots from various PR, branches...
Sorry for the wrong alert. I haven't been to reproduce this (and I really tried :))
All looks good to me to me to be merged.
And released?

tomaswolf:
Edited my comments. There is indeed a bug; I missed one place where getName() must be replaced by getProviderName().

olamy:
> Edited my comments. There is indeed a bug; I missed one place where getName() must be replaced by getProviderName().

ah so I need to test again because I'm confused :)

tomaswolf:
I'll push an update this evening.

olamy:
Thanks.
By the way, I confirm I can reproduce. Sorry I made other changes in my test and forgot to correctly setup `-Djava.security.manager=` which is obviously mandatory to reproduce that.

tomaswolf:
> I'll push an update this evening.

Done. The bug fix is https://github.com/apache/mina-sshd/compare/31234defad60f723ba5f3de4d2a7527784e46e49..8092f28b9f379e3652907a8cad3916e1163bc646#diff-b775cafcbee766e042f74d6790cbaa143b5f175756db29140865a41222e7b94aR310 .

Hope I didn't miss anything else.

olamy:
@tomaswolf All is working fine.
Thanks a lot for your help!

tomaswolf:
Thanks for testing. I'll finish this this evening (documentation needs some updates still), then merge.

> And released?

I have a few other things to fix or clean-up first. But after that I'll ask our release manager if we could do a release.

olamy (approved):

PR #617, opened 2024-09-29
Add interface to configure details of JGit's pack implementation

### LINKED ISSUE #616: Add support for git hooks on GitPackCommand
### Description

JGit supports setting post-receive and pre-receive hooks on `ReceivePack` instances. It should be possible to use them with `GitPackCommand`, e.g. by allowing users to set suppliers of `PostReceiveHook` and `PreReceiveHook` on `GitPackCommandFactory` objects.

### Motivation

There is currently no way to hook into the Git command handling, which means that what happens on remote updates of Git repos is completely opaque to users of mina-sshd. This is very limiting, since it makes it impossible to, for example, prevent certain refs from being pushed to, preventing force pushes (pre-receive) or kicking off something else if pushing succeeded (post-receive).

### Alternatives considered

From what I've read, JGit is supposed to be able to run hooks that are on disk as executable files in the `hooks` directory, but it seems like this support is missing from `ReceivePack`. I haven't managed to get JGit to run those hooks either.

### Additional context

/

JohnnyJayJay:
My proposal to close #616 .

I'm open for suggestions for changes to the interface or the naming, but in general I think this would be a very useful thing to have.

### DIFF
--- a/docs/git.md
+++ b/docs/git.md
@@ -39,6 +39,26 @@ sshd.setCommandFactory(new GitPackCommandFactory().withGitLocationResolver(resol
 
 ```
 
+To further control what happens in response to `git-receive-pack` and `git-upload-pack`, you can implement a `GitPackConfiguration` which will be used
+to configure JGit's `ReceivePack` and `UploadPack`, respectively:
+
+```java
+sshd.setCommandFactory(new GitPackCommandFactory()
+    .withGitPackConfiguration(new GitPackConfiguration() {
+        @Override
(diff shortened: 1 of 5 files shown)


tomaswolf (commented):
Looks good, just some minor details that will make the CI build fail.

Also: run this at least once through the maven build. The maven build enforces our formatting rules by re-formatting. If there are any changes after that, commit them: the CI build will fail if formatting is not clean.

Finally, please squash all these small commits into one.

tomaswolf:
Thank you!

PR #653, opened 2025-01-13
Abstract revoked key handling in KnownHostsServerKeyVerifier

### LINKED ISSUE #652: Add option to extend KnownHostServerKeyVerifier handling of revoked keys
### Description

Currently, KnownHostServerKeyVerifier does not have handle for revoked keys. It simply returns false. It would be good to abstract it to a separate function that you can override if you want to handle it in a specific way. Now you need to override acceptKnownHostEntries in order to add special handling. 

### Motivation

It would be a lot easier to add functionality when a revoked key matches. E.g. giving information to the user, etc. 

### Alternatives considered

A seperate function to log etc when a revoked key matches.
Currently overriding acceptKnownHostEntries.

### Additional context

_No response_

joeljohansson99:
Abstract handling of revoked key so extending classes has an easier way to add functionality when a matching host entry has a 'revoked' marker. 

Fixes #652.

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/client/keyverifier/KnownHostsServerKeyVerifier.java
+++ b/sshd-core/src/main/java/org/apache/sshd/client/keyverifier/KnownHostsServerKeyVerifier.java
@@ -285,8 +285,7 @@ public class KnownHostsServerKeyVerifier
 
         if (keyMatches.stream()
                 .anyMatch(k -> "revoked".equals(k.getHostEntry().getMarker()))) {
-            log.debug("acceptKnownHostEntry({})[{}] key={}-{} marked as revoked",
-                    clientSession, remoteAddress, KeyUtils.getKeyType(serverKey), KeyUtils.getFingerPrint(serverKey));
+            handleRevokedKey(clientSession, remoteAddress, serverKey);
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


tomaswolf:
The formatter is not happy. Run a local maven build; it'll reformat the file. Then commit (amend) the result and force push.

joeljohansson99:
Hi Thomas, thanks for the feedback. I ran the formatting, and committed. I did not get the amend to work, hope that is alright.

sshd-core/src/main/java/org/apache/sshd/client/keyverifier/KnownHostsServerKeyVerifier.java
```diff
@@ -536,6 +537,16 @@ protected List<HostEntryPair> findKnownHostEntries(
...
 
+    /**
+     * Invoked if any matching host entry has a 'revoked' marker
+     *
+     * @param clientSession The {@link ClientSession}
+     * @param remoteAddress The remote host address
+     * @param serverKey     The presented server {@link PublicKey}
+     */
+    protected void handleRevokedKey(ClientSession clientSession, SocketAddress remoteAddress, PublicKey serverKey) {
+    }
```
tomaswolf, line 548: Now PMD is unhappy about this empty block.

Either insert a comment `// By default do nothing`, or move the debug logging from before the call into this body. The latter might make sense; if someone overrides this, he can then decide whether to have that debug logging (call super) or not (don't call super, do something else entirely).

joeljohansson99:
I moved the debug logging. Think it makes sense too. Thanks

PR #660, opened 2025-01-31
bugfix: fix cert auth failed bug

### LINKED ISSUE #650: Supplied key (org.apache.sshd.common.config.keys.OpenSshCertificateImpl) is not a RSAPublicKey instance
### Version

2.8.0

### Bug description

I overwrote the authenticate function of the PublickeyAuthenticator.java and registered it with the ssh server.

```
sshServer.setPublickeyAuthenticator(publicKeyAuth);
```
But when the client provides the key and certificate, MINA SSHD displays an error:

```
2024-12-25 17:29:30.882 sshd-SshServer[53600a30](port=2222)-nio2-thread-5 WARN  org.apache.sshd.server.session.ServerUserAuthService.warn - handleUserAuthRequestMessage(ServerSessionImpl[null@/127.0.0.1:56562]) Failed (InvalidKeyException) to authenticate using factory method=publickey: Supplied key (org.apache.sshd.common.config.keys.OpenSshCertificateImpl) is not a RSAPublicKey instance
```

After debugging, I found that the problem is that when MINA SSHD initializes the signature verifier of the public key, the public key passed in is of type OpenSshCertificateImpl, while Signature expects an RSAPublicKey object. Why does this problem occur? How can I solve it?

### Actual behavior

I rewrote the initVerifier function in SignatureRSA.java and now I can login correctly.
Old initVerifier function:
```
public void initVerifier(SessionContext session, PublicKey key) throws Exception {
      super.initVerifier(session, key);
      RSAKey rsaKey = ValidateUtils.checkInstanceOf(key, RSAKey.class, "Not an RSA key");
      verifierSignatureSize = getVerifierSignatureSize(rsaKey);
}
```
I overwrote the initVerifier function:
```
public void initVerifier(SessionContext session, PublicKey key) throws Exception {
      if (key instanceof OpenSshCertificate){
          super.initVerifier(session, ((OpenSshCertificate) key).getCertPubKey());
          RSAKey rsaKey = ValidateUtils.checkInstanceOf(((OpenSshCertificate) key).getCertPubKey(), RSAKey.class, "Not an RSA key");
          verifierSignatureSize = getVerifierSignatureSize(rsaKey);
      }else {
          super.initVerifier(session, key);
          RSAKey rsaKey = ValidateUtils.checkInstanceOf(key, RSAKey.class, "Not an RSA key");
          verifierSignatureSize = getVerifierSignatureSize(rsaKey);
      }
  }
```
I am still curious as to why this happens？ And whether there are any hidden dangers in my modification.
`

### Relevant log output

_No response_

### Other information

_No response_

liuziyu1226:
Fixes #650

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/server/auth/pubkey/UserAuthPublicKey.java
+++ b/sshd-core/src/main/java/org/apache/sshd/server/auth/pubkey/UserAuthPublicKey.java
@@ -124,12 +124,16 @@ public class UserAuthPublicKey extends AbstractUserAuth implements SignatureFact
             log.debug("doAuth({}@{}) verify key type={}, factories={}, fingerprint={}",
                     username, session, alg, NamedResource.getNames(factories), KeyUtils.getFingerPrint(key));
         }
-
+        /*
+         * When users employ cert authentication, need to use the public key in the cert for signing
+         * and cannot use the cert itself directly for signing
+         */
+        PublicKey verifyKey = key instanceof OpenSshCertificate ? ((OpenSshCertificate) key).getCertPubKey() : key;
         Signature verifier = ValidateUtils.checkNotNull(
                 NamedFactory.create(factories, alg),
                 "No verifier located for algorithm=%s",
                 alg);
-        verifier.initVerifier(session, key);
+        verifier.initVerifier(session, verifyKey);
         buffer.wpos(oldLim);
 
         byte[] sig = hasSig ? buffer.getBytes() : null;
--- a/sshd-core/src/test/java/org/apache/sshd/common/auth/PublicKeyAuthenticationTest.java
+++ b/sshd-core/src/test/java/org/apache/sshd/common/auth/PublicKeyAuthenticationTest.java
@@ -32,8 +32,11 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Stream;
 
+import org.apache.sshd.certificate.OpenSshCertificateBuilder;
 import org.apache.sshd.client.SshClient;
 import org.apache.sshd.client.auth.keyboard.UserInteraction;
 import org.apache.sshd.client.auth.pubkey.PublicKeyAuthenticationReporter;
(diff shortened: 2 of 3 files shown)


sshd-core/src/test/java/org/apache/sshd/common/auth/PublicKeyAuthenticationTest.java
```diff
@@ -459,4 +460,78 @@ void rsaAuthenticationOldServer() throws Exception {
             }
         }
     }
+
```
tomaswolf, line 463: The test needs to be formatted. Run the maven build; it'll format the source. Then commit the formatting change.

sshd-core/src/main/java/org/apache/sshd/server/auth/pubkey/UserAuthPublicKey.java
```diff
@@ -176,7 +180,7 @@ public Boolean doAuth(Buffer buffer, boolean init) throws Exception {
                     username, session, alg, KeyUtils.getFingerPrint(key));
         }
 
-        return Boolean.TRUE;
+        return key instanceof OpenSshCertificate;
```
tomaswolf, line 183: This is wrong. Change back to `return Boolean.TRUE`.  The method is supposed to return TRUE if authentication was successful, FALSE if unsuccessful, and null if not complete yet.

sshd-core/src/test/java/org/apache/sshd/common/auth/PublicKeyAuthenticationTest.java
```diff
@@ -459,4 +460,78 @@ void rsaAuthenticationOldServer() throws Exception {
...
+                assertEquals(2, authAttempts.get(), "There should be two attempts to authenticate using the certificate");
+            } finally {
+                client.stop();
+            }
+        }
+    }
+
+    private static Stream<Arguments> certificateAlgorithms() {
+        return Stream.of(
+            // 算法名称、密钥大小、签名算法
```
tomaswolf, line 529: This project uses English throughout. Please remove this comment, or write it in English: key algorithm, key size, signature algorithm

sshd-core/src/test/java/org/apache/sshd/common/auth/PublicKeyAuthenticationTest.java
liuziyu1226: My apologies for the oversight - I neglected to revert the code to its original state after completing the testing phase. Additionally, I've now properly formatted the PR. Kindly review the updated version when convenient.

liuziyu1226: > The test needs to be formatted. Run the maven build; it'll format the source. Then commit the formatting change.

My apologies for the oversight - I neglected to revert the code to its original state after completing the testing phase. Additionally, I've now properly formatted the PR. Kindly review the updated version when convenient.

tomaswolf:
Thanks for this contribution. We're about to create a new Apache MINA sshd 2.15.0 release. I will defer merging this until after that (so it will be in 2.16.0 then). The problem is that the server side of certificate authentication has many other things missing. As far as I can see the server side doesn't even check the signature in the certificate.

PR #716, opened 2025-03-31
Implement client-side support of X11 forwarding

### LINKED ISSUE #712: SshClient supports X11 forwarding
### Description

Currently, `SshClient` does not support X11 forwarding, but `SshServer` seems to.

### Motivation

X11 forwarding

### Alternatives considered

_No response_

### Additional context

_No response_

hstyi:
Closes #712

## example

```java
// create client...
SshClient client;
// set X11ChannelFactory
client.setChannelFactories(Collections.singletonList(X11ChannelFactory.INSTANCE));

// create session...
ClientSession session = client.connect("host").verify(CONNECT_TIMEOUT).getSession();
// set X11 server
CoreModuleProperties.X11_BIND_HOST.set(session, "localhost");
CoreModuleProperties.X11_BASE_PORT.set(session, 6000);

// create channel...
ChannelShell channel = session.createShellChannel();
// enable X11 forwarding
channel.setXForwarding(true);


```

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/client/channel/ChannelX11.java
+++ b/sshd-core/src/main/java/org/apache/sshd/client/channel/ChannelX11.java
@@ -1,0 +1,166 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.sshd.client.channel;
+
+import org.apache.sshd.client.future.DefaultOpenFuture;
+import org.apache.sshd.client.future.OpenFuture;
+import org.apache.sshd.client.session.ClientConnectionService;
+import org.apache.sshd.client.x11.X11IoHandler;
+import org.apache.sshd.common.Property;
+import org.apache.sshd.common.SshConstants;
+import org.apache.sshd.common.channel.ChannelOutputStream;
+import org.apache.sshd.common.io.IoConnectFuture;
+import org.apache.sshd.common.io.IoConnector;
+import org.apache.sshd.common.io.IoHandler;
+import org.apache.sshd.common.io.IoSession;
+import org.apache.sshd.common.util.buffer.Buffer;
+import org.apache.sshd.common.util.buffer.ByteArrayBuffer;
+
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * @author <a href="mailto:dev@mina.apache.org">Apache MINA SSHD Project</a>
+ */
+public class ChannelX11 extends AbstractClientChannel {
+
+    public static final Property<Object> X11_COOKIE = Property.object("x11-cookie");
+    public static final Property<Object> X11_COOKIE_HEX = Property.object("x11-cookie-hex");
+
+    private final AtomicBoolean isInitialized = new AtomicBoolean(false);
+    private final String host;
+    private final int port;
+
+    private IoSession x11;
+
+    public ChannelX11(String host, int port) {
+        super("x11");
+        this.host = host;
+        this.port = port;
+    }
+
+    @Override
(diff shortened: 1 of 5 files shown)


sshd-core/src/main/java/org/apache/sshd/client/channel/PtyCapableChannelSession.java
```diff
@@ -271,4 +302,34 @@ protected void doOpenPty() throws IOException {
...
+            return (byte[]) xCookieHex;
+        }
+
+        synchronized (session) {
+            xCookieHex = ChannelX11.X11_COOKIE_HEX.getOrNull(session);
+            if (xCookieHex instanceof byte[]) {
+                return (byte[]) xCookieHex;
+            }
+
+            byte[] foo = new byte[16];
```
tomaswolf, line 320: Give variables meaningful names. "foo" and "bar" just won't do. Here "foo" could be named "fakeCookie".

sshd-core/src/main/java/org/apache/sshd/client/channel/ChannelX11.java
```diff
@@ -0,0 +1,166 @@
...
+
+            if (l < 12 + plen + ((-plen) & 3) + dlen) {
+                return;
+            }
+
+            byte[] bar = new byte[dlen];
+            System.arraycopy(foo, s + 12 + plen + ((-plen) & 3), bar, 0, dlen);
+
+            if (Objects.deepEquals(xCookie, bar)) {
+                x11.writeBuffer(new ByteArrayBuffer(foo, s, l));
```
tomaswolf, line 131: If the cookie is the one that was random-generated below, then you would have to replace it here with the real cookie. So there must be a way to set that real cookie on the channel. Client code could read it from the local .Xauthority file.

sshd-core/src/main/java/org/apache/sshd/client/channel/PtyCapableChannelSession.java
```diff
@@ -239,6 +255,21 @@ protected void doOpenPty() throws IOException {
             writePacket(buffer);
         }
 
+        if (xForwarding) {
```
tomaswolf, line 258: Check that the X11 host & port properties are set. If not, it's futile to request X11 forwarding.

Also, I think there will need to be a way to pass in the cookie (from the .Xauthority file).

sshd-core/src/main/java/org/apache/sshd/client/channel/PtyCapableChannelSession.java
```diff
@@ -34,6 +30,11 @@
 import org.apache.sshd.common.util.buffer.ByteArrayBuffer;
 import org.apache.sshd.core.CoreModuleProperties;
 
+import java.io.IOException;
+import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.ThreadLocalRandom;
+
```
tomaswolf, line 37: Keep the order of imports.

tomaswolf (commented):
Thanks for this effort. Some inline comments.

Please also give a meaningful commit message. Title line, blank line, then details.

I didn't have the time yet to download and actually run the test.

In general:
* I don't understand what the tests are testing. In the end, there must be a way to tell the channel the real cookie used by the local X11 server, and a full test would need to connect to a local X11 server. I understand that it might be hard to do so in an automated integration test, but I'd expect at least a test that can be run manually when on a Linux or BSD system that does have a running X11 server.
* Some of the code has been copied verbatim from JSch. This is a no-go for me for two reasons:
    * I don't want to include the ymnk license in Apache MINA ssh, even if it is a BSD 3-Clause license.
    * JSch code is written in a rather obfuscated style that is hard to maintain. (Apache MINA sshd style may also be perceived as "obfuscated" and also is not easy to maintain, but it's "obfuscated" in a different way :-).) The point is that the original JSch code was written by a single person who had all the details about the code in his head. That won't do for a maintainable open-source project. Code here should be reasonably clear and have reasonable variable and object names so that others have a chance to understand it.

On the client side, X11 forwarding is basically the same as remote port forwarding plus the handling of the initial X11 connection setup message with the cookie. I wonder if the support already there for remote port forwarding shouldn't be re-used.

The change should also include documentation updates about how to use this feature.

sshd-core/src/main/java/org/apache/sshd/client/channel/PtyCapableChannelSession.java
```diff
@@ -239,6 +255,21 @@ protected void doOpenPty() throws IOException {
...
+        if (xForwarding) {
+            if (debugEnabled) {
+                log.debug("doOpenPty({}) Send xForwarding request", this);
+            }
+            Buffer buffer = session.createBuffer(SshConstants.SSH_MSG_CHANNEL_REQUEST);
+            buffer.putInt(getRecipient());
+            buffer.putString("x11-req");
+            buffer.putBoolean(false); // want-reply
+            buffer.putBoolean(false);
+            buffer.putString("MIT-MAGIC-COOKIE-1");
```
tomaswolf, line 267: RFC 4254 does not require MIT-MAGIC-COOKIE-1 to be used. User code will need to be able to configure the authentication protocol name and "cookie" (actually: arbitrary authentication data), for instance after having read it from the .Xauthority file.

hstyi:
Thank you for reviewing the code, English is not my first language and I have a hard time reading the X11 documentation.

I don't think I have the ability to perfect this PR, I may implement it in the future, but not now.

Thanks again, @tomaswolf

Issue #725, opened 2025-04-11
Add maxCommandExecutionTimeout to session.executeRemoteCommand


raajeive:
### Description

Introduce a maxCommandExecutionTimeout parameter to the session.executeRemoteCommand method in Apache MINA SSHD. This is essential to prevent command executions from hanging indefinitely due to unresponsive commands or misbehaving remote servers.

### Motivation

In our environment, we need to SSH into a large number of servers and execute commands. Occasionally, some servers are in a bad state, causing the command execution to hang indefinitely. This not only blocks the current session but also delays subsequent SSH operations to other servers.

To work around this, we currently have to create a custom execution channel and manually manage timeouts. Providing native support for a maxCommandExecutionTimeout within executeRemoteCommand would greatly simplify this process and improve robustness when dealing with unreliable targets.

`            // TODO use a configurable timeout
            Collection<ClientChannelEvent> waitMask = channel.waitFor(REMOTE_COMMAND_WAIT_EVENTS, 0L);
            if (waitMask.contains(ClientChannelEvent.TIMEOUT)) {
                 throw new SocketTimeoutException("Failed to retrieve command result in time: " + command);
            }
`

raajeive:
I made the changes in the library to accept the commandTimeoutMillis in the executeRemoteCommand function.

https://github.com/apache/mina-sshd/pull/726

PR #726, opened 2025-04-11
[GH-725] Added commandTimeoutMillis in executeRemoteCommand

raajeive:
Added commandTimeoutMillis in executeRemoteCommand

Implemets https://github.com/apache/mina-sshd/issues/725

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/client/session/ClientSession.java
+++ b/sshd-core/src/main/java/org/apache/sshd/client/session/ClientSession.java
@@ -37,6 +37,7 @@ import java.util.EnumSet;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.sshd.client.ClientAuthenticationManager;
 import org.apache.sshd.client.ClientFactoryManager;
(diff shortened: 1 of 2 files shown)


sshd-core/src/main/java/org/apache/sshd/client/session/ClientSession.java
```diff
@@ -281,19 +282,25 @@ default String executeRemoteCommand(String command, OutputStream stderr, Charset
...
+     *                             channel opening and result waiting. A zero or negative value means no timeout.
+     * @throws IOException         If the command execution fails, times out, or returns a non-zero exit code.
+     *                             A {@link RemoteException} may be thrown if the remote side reports an error.
+     * @see                        ClientChannel#open()#verify(long, java.util.concurrent.TimeUnit)
+     * @see                        ClientChannel#waitFor(Collection, long)
+     * @see                        ClientChannel#validateCommandExitStatusCode(String, Integer) validateCommandExitStatusCode
      */
     default void executeRemoteCommand(
-            String command, OutputStream stdout, OutputStream stderr, Charset charset)
+            String command, OutputStream stdout, OutputStream stderr, Charset charset, long commandTimeoutMillis)
```
tomaswolf, line 303: This is an API change and is thus a no-go.

If you want this, you'll have to add a new method, and make the existing one call the new one.

sshd-core/src/main/java/org/apache/sshd/client/session/ClientSession.java
```diff
@@ -304,10 +311,10 @@ default void executeRemoteCommand(
              ClientChannel channel = createExecChannel(command, charset, null, Collections.emptyMap())) {
             channel.setOut(channelOut);
             channel.setErr(channelErr);
-            channel.open().await(); // TODO use verify and a configurable timeout
 
-            // TODO use a configurable timeout
-            Collection<ClientChannelEvent> waitMask = channel.waitFor(REMOTE_COMMAND_WAIT_EVENTS, 0L);
+            channel.open().verify(commandTimeoutMillis, TimeUnit.MICROSECONDS);
```
tomaswolf, line 315: Time unit mismatch (millis vs micros).

sshd-core/src/main/java/org/apache/sshd/client/session/ClientSession.java
```diff
@@ -304,10 +311,10 @@ default void executeRemoteCommand(
...
+
+            Collection<ClientChannelEvent> waitMask = channel.waitFor(REMOTE_COMMAND_WAIT_EVENTS, commandTimeoutMillis);
```
tomaswolf, line 317: Here we should probably wait only for commandTimeoutMillis minus the time we spent waiting for the channel to open.

tomaswolf (commented):
This needs a test, and must be done in a way that is backwards-compatible.

sshd-core/src/main/java/org/apache/sshd/client/session/ClientSession.java
raajeive: Done

sshd-core/src/main/java/org/apache/sshd/client/session/ClientSession.java
raajeive: Good catch, Thanks, Updated.

raajeive:
Hi @tomaswolf i have addressed all the comments provided, Can you please check this now?

garydgregory:
Hi all,

I think we should stop using ints and longs to express durations ASAP. Java has Duration class, let's use it. If a duration ends up as int seconds or long milliseconds or whatever scale a low level API in Java needs, that's fine, I just feel there is still room for confusion using primitives. A Duration also offers the best precision on Java 9 and up (nanosecond).

sshd-core/src/main/java/org/apache/sshd/client/session/ClientSession.java
```diff
@@ -95,6 +96,11 @@ enum ClientSessionEvent {
         AUTHED
     }
 
+    /**
+     * Minimum timeout value. When used in {@link #executeRemoteCommand}, the command execution will wait indefinitely.
+     */
+    Duration MIN_TIMEOUT = Duration.ofMillis(0L);
```
tomaswolf, line 102: Seems to me that we don't need this constant. Just use `Duration.ZERO`.

raajeive: Resolved

raajeive:
@tomaswolf @garydgregory addressed all the comments, can you please check this and provide the approvals?

tomaswolf:
Thank you!

raajeive:
@tomaswolf @garydgregory @osi @knalli 

When can we expect this to get released?

Issue #771, opened 2025-07-15
Using PuttyKeyUtils I got => NoClassDefFoundError: net/i2p/crypto/eddsa/EdDSAPublicKey


rde-infologic:
### Version

2.15

### Bug description

The request [SSHD-1260](https://issues.apache.org/jira/browse/SSHD-1260) made net.i2p.crypto dependency optional.

When loading PuttyKeyUtils : I got the following error  

> java.lang.NoClassDefFoundError: net/i2p/crypto/eddsa/EdDSAPublicKey
> 	at org.apache.sshd.putty.EdDSAPuttyKeyDecoder.<init>(EdDSAPuttyKeyDecoder.java:48)
> 	at org.apache.sshd.putty.EdDSAPuttyKeyDecoder.<clinit>(EdDSAPuttyKeyDecoder.java:45)
> 	at org.apache.sshd.putty.PuttyKeyUtils.<clinit>(PuttyKeyUtils.java:50)

I have BC dependency so SecurityUtils.isEDDSACurveSupported() return true.

I think that we can add SecurityUtils.isNetI2pCryptoEdDSARegistered() to load EdDSAPuttyKeyDecoder.INSTANCE until this decoder will use new EdDSA loading dependencies.

It's may be usefull to look around all SecurityUtils.isEDDSACurveSupported() call to check if there is no dependency to net.i2p.crypto

Like that :

```
public final class PuttyKeyUtils {
    public static final List<PuttyKeyPairResourceParser<?, ?>> DEFAULT_PARSERS;

    public static final NavigableMap<String, PuttyKeyPairResourceParser<?, ?>> BY_KEY_TYPE;

    public static final KeyPairResourceParser DEFAULT_INSTANCE;

    static {
        List<PuttyKeyPairResourceParser<?, ?>> parsers = new ArrayList<>();
        parsers.add(RSAPuttyKeyDecoder.INSTANCE);
        parsers.add(DSSPuttyKeyDecoder.INSTANCE);
        if (SecurityUtils.isECCSupported()) {
            parsers.add(ECDSAPuttyKeyDecoder.INSTANCE);
        }
        if (SecurityUtils.isEDDSACurveSupported() && SecurityUtils.isNetI2pCryptoEdDSARegistered()) {
            parsers.add(EdDSAPuttyKeyDecoder.INSTANCE);
        }
[...]

PR #773, opened 2025-07-16
#771 Avoid NoClassDefFoundError: net/i2p/crypto/eddsa/EdDSAPublicKey

rde-infologic:
Using PuttyKeyUtils we can get an error if we don't want dependency net.i2p.crypto but having BC dependency, I made some change using SecurityUtils.getEdDSASupport to remove net.i2p.crypto dependency


rde-infologic:
Sry I'm not maven expert. 

I Launch command and format file in default.

PR #790, opened 2025-07-27
Trigger ClientChannelEvent.Timeout and ClientSessionEvent.TIMEOUT independently to host's program cycle times

fersaru:
On some hosts, on which waiting for the `futureLock.wait(remWait)` is faster than 1 millisecond, `remWait` gets reduced by 123 milliseconds although not even one passed.
As each host has different cycle times, this happens an unknown amount of times.
This causes a `ClientChannelEvent.Timeout` after a much smaller time passed than the configured timeout.
With these changes, the `remWait` is calculated based on the milliseconds passed instead of transformed nanoseconds.

The same code is found in `ClientSessionImpl@waitFor`. This is also addressed.


tomaswolf:
Thank you for this fix. The old code looks like someone put in some placeholder code and forgot to clean up later.

PR #804, opened 2025-08-20
Client sent empty ssh package to keep connection issue

Main-Tomi:
I use sshd family libs(version: 2.14.0) to run a sftp server.
But many clients send some empty_ssh_packages to keep the connection.
Based on our business requirements, we do not wish to maintain such persistent connections. 
Therefore, I have made this modification, hoping to add an additional parameter to control this behavior.


tomaswolf:
This doesn't make sense. You can't stop clients from sending SSH_MSG_IGNORE packets. There are already mechanisms in place that enable you to control how your server handles these messages. And I think there are enough hooks in place that would enable you to implement some kind of global connection timeout.

PR #843, opened 2025-10-16
fix: negate followLinks flag for checkSymlinkState

bemoty:
I'm using sshd-sftp as a custom SFTP server for its great virtual file system capabilities (it allows me to easily create a virtual space for each user). Each user directory contains symlinks (essentially "shares") to folders outside of the virtual file system.

After upgrading to 2.16.0 I noticed that file deletion did not work anymore. It kept giving me a `NoSuchFileException` "Removal candidate not found".

In `doRemoveFile` it seems like `followLinks` is missing a `!` before being passed to `checkSymlinkState`. This was done correctly in other methods such as `doRemoveDirectory`.


tomaswolf:
Good catch!

Looking through the callers of `checkSymlinkState` it appears that it is also wrong at

https://github.com/apache/mina-sshd/blob/2bb67e61dd16479fc1e0a796914185b7c3ca464f/sshd-sftp/src/main/java/org/apache/sshd/sftp/server/SftpSubsystem.java#L848

Can we also add tests for this (at least on Unix-like platforms, where we have symlinks)?

bemoty:
I gave it a try! Not sure if this was the right place for them though... let me know if you need me to move something. I created one test case for each `checkSymlinkState` caller, and excluded Win32 from the tests.

tomaswolf:
Thanks a lot for the fixes and tests!

PR #849, opened 2025-10-27
Add Maven Wrapper

chrjohn:
... to eliminate the need to install a specific Maven version locally.

### DIFF
--- a/assembly/src/main/descriptors/unix-src.xml
+++ b/assembly/src/main/descriptors/unix-src.xml
@@ -40,7 +40,8 @@
                 <include>**/*.bin</include>
                 <include>**/*.jnilib</include>
                 <include>**/*.cert</include>
-                <include>**/.mvn/*</include>
+                <include>**/.mvn/**</include>
+                <include>mvnw.cmd</include>
 
                 <include>sshd-openpgp/src/test/resources/**/*</include>
             </includes>
@@ -92,6 +93,7 @@
 
                 <exclude>sshd-core/src/docs/*.txt</exclude>
                 <exclude>sshd-openpgp/src/test/resources/**/*</exclude>
+                <exclude>mvnw.cmd</exclude>
             </excludes>
             <lineEnding>unix</lineEnding>
         </fileSet>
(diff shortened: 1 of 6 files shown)


tomaswolf:
What about line endings? Force LF for mvnw and CR-LF for mvnw.cmd on checkout via .gitattributes? Also we should ensure in the assembly project that these two files get the right line endings when archived.

What line endings do the properties need? Will they work if it's LF-only on Windows?

chrjohn:
I'll look into the attributes stuff.

As for the properties: I cannot test this on Windows but I have a similar setup on the QuickFIX/J project where the wrapper works fine in github actions on Linux and Windows. https://github.com/quickfix-j/quickfixj/blob/master/.github/workflows/maven.yml

assembly/src/main/descriptors/unix-src.xml
chrjohn: I'm not sure I got the logic right on `unix-src.xml` and `windows-src.xml`.

tomaswolf: LGTM.

tomaswolf:
Works as intended on Windows in cmd.exe and in git bash; works well on Mac. Scripts have the correct line endings in the source archives produced in the assembly project.

This look good. Please squash the commits into one, and give it a good commit message. Then this is ready to go in.

Thanks a lot!

chrjohn:
@tomaswolf could you please check if it is OK. Probably my last merge messed things up?

tomaswolf:
> Probably my last merge messed things up?

Yes, it did :-) Just check-out your squash commit e82a9d3 and force-push it onto your chrjohn:mvn-wrapper branch (i.e., the mvn-wrapper branch in your fork). That should fix it: we should end up with this PR being updated and having only the single commit e82a9d3.

The squash itself worked fine.

chrjohn:
Hmm, force pushing does not work, says working directory is clean. Will check again tomorrow...

tomaswolf:
Shall I do it?

chrjohn:
That would be great, thanks. :)

tomaswolf:
Done :-)

chrjohn:
> Done :-)

Thank you. :)

PR #858, opened 2025-12-12
Replace "synchronized" blocks in AbstractSession with ReentrantLock

noxvost:
Synchronized blocks pin the virtual thread under JDK21+ to physical thread. It is better to avoid such pinning in I/O bound operations so VirtualThread planner can do its job


tomaswolf:
There are a _lot_ more synchronized blocks and methods in Apache MINA sshd. I'm fine with replacing these two; I would object to replacing all uses of `synchronized` by re-entrant locks.

If thread pinning is a problem, I think you could migrate to Java >= 24, which should have [JEP-491](https://openjdk.org/jeps/491) implemented: using `synchronized` together with virtual threads should not be a problem anymore.

PR #870, opened 2026-01-11
Fix duplicate character echoing in ProcessShell

### LINKED ISSUE #469: InteractiveProcessShellFactory .  Each character entered is repeated
### Version

2.12.1

### Bug description

Start a service with sshd, connect to the service with iterm, and repeat each time you enter the command, including carriage return

### Actual behavior

Using follow code to start one server .

```
        SshServer sshd = ServerBuilder.builder()
                .build();

        sshd.setShellFactory(InteractiveProcessShellFactory.INSTANCE);

        SimpleGeneratorHostKeyProvider keyPairProvider = new SimpleGeneratorHostKeyProvider(Paths.get("/tmp/key"));
        keyPairProvider.setOverwriteAllowed(false);
        sshd.setKeyPairProvider(keyPairProvider);

        sshd.setPasswordAuthenticator(AcceptAllPasswordAuthenticator.INSTANCE);
        sshd.setFileSystemFactory(new NativeFileSystemFactory());
        sshd.setCommandFactory(ProcessShellCommandFactory.INSTANCE);
        sshd.setPort(22);
        sshd.start();
        System.out.println("sshd.isStarted() = " + sshd.isStarted());
        TimeUnit.DAYS.sleep(1);
        sshd.stop();
```

Console log :
```
15:58:02.758 [main] INFO  org.apache.sshd.common.io.DefaultIoServiceFactoryFactory [getIoServiceProvider:68] - No detected/configured IoServiceFactoryFactory using Nio2ServiceFactoryFactory
sshd.isStarted() = true
```

Then use Iterm to connect to ssh server.
image

### Expected behavior

Expect to use commands like a normal shell after connecting to the sshd service

### Relevant log output

_No response_

### Other information

Thanks a lot

denyshorman:
### Summary
This PR fixes a bug where characters were duplicated when using an interactive shell (e.g., "ping" appearing as "ppiinngg"). The fix involves removing the redundant manual echoing logic from `TtyFilterOutputStream`.

### Problem Description
When using `InteractiveProcessShellFactory`, MINA SSHD starts a local process (like `/bin/sh` or `cmd.exe`) and bridges its I/O streams using `ProcessShell`. `ProcessShell` used `TtyFilterOutputStream` to wrap the process's input stream. 

`TtyFilterOutputStream` was configured to manually echo every character received from the client back to the client's output stream if `PtyMode.ECHO` was enabled (which is common for interactive sessions). However, interactive shells already perform their own echoing. Because `ProcessShell` operates over pipes rather than a real PTY, both the server-side filter and the shell process were echoing the input, leading to doubled characters.

Standard SSH implementations (like OpenSSH) do not perform manual echoing in the server for shell sessions; they rely on the TTY or the shell process to handle it.

### Solution
The manual echoing mechanism in `TtyFilterOutputStream` has been removed. 
1. `PtyMode.ECHO` is no longer handled by `TtyFilterOutputStream`.
2. The `echo` parameter (a `TtyFilterInputStream`) has been removed from `TtyFilterOutputStream` constructors.
3. `ProcessShell` has been updated to stop passing the echo stream.
4. Added a test `ProcessShellTest` to verify that input is no longer redundantly echoed to the error stream.

This ensures that only the underlying shell process handles echoing, which is the correct behavior for this type of shell bridging.

Fixes #469.

### DIFF
--- a/sshd-common/src/main/java/org/apache/sshd/server/shell/TtyFilterOutputStream.java
+++ b/sshd-common/src/main/java/org/apache/sshd/server/shell/TtyFilterOutputStream.java
@@ -25,33 +25,30 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 
 import org.apache.sshd.common.channel.PtyMode;
 import org.apache.sshd.common.util.GenericUtils;
 
 /**
- * Handles the output stream while taking care of the {@link PtyMode} for CR / LF and ECHO settings
+ * Handles the output stream while taking care of the {@link PtyMode} for CR / LF settings
  *
  * @author <a href="mailto:dev@mina.apache.org">Apache MINA SSHD Project</a>
  */
 public class TtyFilterOutputStream extends FilterOutputStream {
     public static final Set<PtyMode> OUTPUT_OPTIONS
-            = Collections.unmodifiableSet(EnumSet.of(PtyMode.ECHO, PtyMode.INLCR, PtyMode.ICRNL, PtyMode.IGNCR));
+            = Collections.unmodifiableSet(EnumSet.of(PtyMode.INLCR, PtyMode.ICRNL, PtyMode.IGNCR));
 
     private final Set<PtyMode> ttyOptions;
-    private final TtyFilterInputStream echo;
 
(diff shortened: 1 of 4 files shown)


tomaswolf:
Great job; thank you!

denyshorman:
Thanks for the quick review!

PR #882, opened 2026-02-12
Docs/Add Android instructions, demo app and example

Chiogros:
Hi 👋🏻 ,

I'm developing [Trante](https://github.com/Chiogros/Trante), an Android app to allow remote storage access from Android's file manager.
I used _mina-sshd_ for SFTP access to remote servers and I wanted to contribute back to help future developers to use it.

This PR brings:
- a complete Kotlin example
- a link to a class handling sshd-sftp calls
- Proguard instructions to help with app optimizations
- moving examples first and issues + next features then

I addition,
> we would appreciate feedback on problems and solutions our users have encountered when running on Android

so here it is.

1. Configuring filesystem.
The existing example didn't do the trick for me. I targeted Android SDK 35 and as far as I can remember, I couldn't get a `Path` to use in `Supplier` because some methods were not available anymore on the JDK. After some digging, [here](https://github.com/Chiogros/Trante/blob/main/app/src/main/java/chiogros/trante/protocols/sftp/data/network/SftpNetwork.kt)'s what worked:
```kotlin
// Set Android's filesystem path
val path: Supplier<Path> = Supplier { Paths.get("") }
setUserHomeFolderResolver(path)
val client: SshClient = SshClient.setUpDefaultClient()

client.start()
```

2. Release builds optimization
Only during release builds, the app crashed. During release builds, [R8](https://developer.android.com/topic/performance/app-optimization/enable-app-optimization) (an app optimizer) struggles to find any reference to some code pieces using [reflection](https://developer.android.com/topic/performance/app-optimization/library-optimization#use-codegen) in  the library and deletes those from the final build. Unfortunately at runtime, these deleted pieces of code were really needed for the app to work.
After hours of build & trial, I've figured out which classes need to be protected from R8 wiping. It led to this [`proguard-rules.pro`](https://github.com/Chiogros/Trante/blob/main/app/proguard-rules.pro) ruleset:
```
-keep,allowoptimization,allowobfuscation class org.apache.sshd.common.io.nio2.Nio2ServiceFactoryFactory { *; }
-keep,allowoptimization,allowobfuscation class org.apache.sshd.common.session.helpers.SessionHelper { *; }
-keep,allowoptimization class org.apache.sshd.common.util.security.bouncycastle.BouncyCastleSecurityProviderRegistrar { *; }
-keep,allowoptimization class org.apache.sshd.common.util.security.eddsa.EdDSASecurityProviderRegistrar { *; }
-keep,allowoptimization class org.apache.sshd.common.util.security.SunJCESecurityProviderRegistrar { *; }
-dontwarn org.apache.sshd.**
``` 

Thank you!

### DIFF
--- a/docs/android.md
+++ b/docs/android.md
@@ -74,6 +74,34 @@ public class MyApplication extends Application {
 }
 ```
 
+Another example (in Kotlin flavor):
+```kotlin
+suspend fun connect(host: String, port: Int, user: String, pwd: String): ClientSession {
+    // Set Android's filesystem path
+    val path: Supplier<Path> = Supplier { Paths.get("") }
+    setUserHomeFolderResolver(path)
+    val client: SshClient = SshClient.setUpDefaultClient()
+
+    client.start()
+
+    return withContext(Dispatchers.IO) {
+        // Connect to server
+        val session = client.connect(user, host, port).verify().clientSession
+        // Supply password
+        session.addPasswordIdentity(pwd)
+
+        // Start authentication
+        val authVerif = session.auth()
+        authVerif.verify()
+
+        if (authVerif.isSuccess) {
+            // Return session for further server calls
+            session
+        } else throw authVerif.exception
+    }
+}
+```
+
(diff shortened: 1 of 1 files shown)


PortaSFTPServer:
Thank you. I am developing a library that uses IKVM to make use of Apache Mina SSHD Java Library in .NET/.NET Core. This will attract more developer to use Apache Mina SSHD in the future. Sample demo for simple SFTP Server using C#  https://github.com/PortaSFTPServer/ApacheMinaSSHDNetServer

PR #902, opened 2026-06-19
Preserve security key signature blobs

nataphon-ktsystems:
Security-key (`sk-*`) signatures carry a structured OpenSSH signature blob: algorithm, raw signature, flags, and counter. The client-side auth path and OpenSSH agent proxy path were treating those like ordinary two-field signatures, which dropped or double-wrapped the trailing security-key fields.

This change preserves the full security-key signature blob when reading OpenSSH agent responses, forwarding local agent responses, and appending the final userauth signature. Ordinary key types keep the existing algorithm-plus-signature wrapping.

Tests cover both sides of the client path:

- `AgentUnitTest.securityKeySignatureBlob` verifies a security-key signature blob survives the agent protocol path with flags and counter intact.
- `UserAuthPublicKeySkTest.securityKeySignatureBlobIsNotWrapped` verifies the client userauth packet writes the security-key signature blob directly instead of wrapping it again.

Validation:

```bash
./mvnw -pl sshd-core -am -Dtest=AgentUnitTest,UserAuthPublicKeySkTest -Dsurefire.failIfNoSpecifiedTests=false test
```

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/agent/common/AbstractAgentProxy.java
+++ b/sshd-core/src/main/java/org/apache/sshd/agent/common/AbstractAgentProxy.java
@@ -32,9 +32,11 @@ import org.apache.sshd.agent.SshAgentConstants;
 import org.apache.sshd.agent.SshAgentKeyConstraint;
 import org.apache.sshd.common.SshException;
 import org.apache.sshd.common.config.keys.KeyUtils;
+import org.apache.sshd.common.config.keys.u2f.SecurityKeyPublicKey;
 import org.apache.sshd.common.session.SessionContext;
 import org.apache.sshd.common.util.GenericUtils;
 import org.apache.sshd.common.util.buffer.Buffer;
+import org.apache.sshd.common.util.buffer.BufferException;
 import org.apache.sshd.common.util.buffer.BufferUtils;
 import org.apache.sshd.common.util.buffer.ByteArrayBuffer;
 import org.apache.sshd.common.util.logging.AbstractLoggingBean;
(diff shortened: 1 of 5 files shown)


sshd-core/src/main/java/org/apache/sshd/agent/common/AbstractAgentProxy.java
```diff
@@ -175,7 +176,11 @@ public Map.Entry<String, byte[]> sign(SessionContext session, PublicKey key, Str
         } else {
             Buffer buf = new ByteArrayBuffer(signature);
             String algorithm = buf.getString();
-            signature = buf.getBytes();
+            if (key instanceof SecurityKeyPublicKey<?>) {
+                signature = signature.clone();
```
tomaswolf: I don't think this is correct. `signature` here should not contain the algorithm name. This should be `signature = buf.getCompactData()`, and then there is no change at all needed in `AbstractAgentClient` or in `UserAuthPublicKey`.

And in fact we should probably check that there is no trailing garbage:
```
if (key instanceof SecurityKeyPublicKey<?>) {
  signature = buf.getCompactData();
  long sigLength = 4 + buf.getUInt() + 5; // 4 for the just read UInt, 1 byte flag, 4 bytes counter
  if (signature.length > sigLength) {
    throw new SshException("Trailing garbage in signing response, expected " + sigLength + " but got " + signature.length + " bytes");
  } else if (signature.length < sigLength) {
    throw new SshException("Signing response too short");
  }
} else {
  signature = buf.getString();
  if (buf.avaliable() > 0) {
    throw new SshException("Trailing garbage in signing response, got" + buf.available() + " extra bytes");
  }
}
```

tomaswolf: Scratch that; it would still wrap once too often.

sshd-core/src/main/java/org/apache/sshd/client/auth/pubkey/UserAuthPublicKey.java
```diff
@@ -465,8 +466,12 @@ protected byte[] appendSignature(
         }
 
         bs.clear();
-        bs.putString(signatureAlgo);
-        bs.putBytes(sig);
+        if (key instanceof SecurityKeyPublicKey<?>) {
+            bs.putRawBytes(sig);
```
tomaswolf, line 471: Compare with what OpenSSH does in the server-side pubkey auth: https://github.com/openssh/openssh-portable/blob/0d156d385e1e9c31bd5fb3652cf620a1f51c9b19/auth2-pubkey.c#L186

It reads a "string", i.e. length and data, and later at https://github.com/openssh/openssh-portable/blob/0d156d385e1e9c31bd5fb3652cf620a1f51c9b19/auth2-pubkey.c#L224 is verifies the signature and extracts the flags and counter into `sig_details`.

So this `putRawBytes` is definitely not correct. "sig" must not contain the algorithm name here, and the previous code was correct.

tomaswolf: Ok, I see now where this all goes wrong. You're right, but I got confused by the fact that the way you do it, we suddenly have signatures that already include the algorithm name.

I'd find it somehow easier to understand if we kept these signatures without type so that the code here becomes
```
bs.putString(signatureAlgo);
if (key instanceof SecurityKeyPublicKey<?>) {
  bs.putRawBytes(sig);
} else {
  bs.putBytes(sig);
}
```
I find that would align better with the signer interface, which returns type and signature separately (line 448).

tomaswolf:
Thanks a lot. I've added an integration test in a follow-up commit.

nataphon-ktsystems:
Glad it helps! Thanks for such a responsive review. Looking forward to be the part of Apache community.

Issue #905, opened 2026-06-28
Implement more checks in AuthorizedKeyEntriesPublickeyAuthenticator


tomaswolf:
Resolve the TODOs: handle "from" and "expiry-time" entry attributes and provide unit tests for it.

Note that the "from" attribute may match host names, so it may involve a reverse DNS lookup. Use `InetAddress.getCanonicalHostName()` for that.

PR #907, opened 2026-07-05
GH-905 Implement authorized_keys from and expiry-time checks

Unbbal:
This PR implements the missing authorized_keys option checks tracked in GH-905.

Changes:
- Enforce the expiry-time option before accepting a matching authorized_keys entry.
- Enforce the from option using the client address, with support for address/hostname patterns, negation, and CIDR ranges.
- Prefer ServerSession#getClientAddress(), falling back to getRemoteAddress().
- Reject invalid from patterns and invalid expiry-time values fail-closed.
- Add unit tests for matching from, rejected from, CIDR, negation, invalid from, future expiry-time, and expired keys.

Tests:
- .\mvnw.cmd -pl sshd-core -am -Dtest=AuthorizedKeyEntriesPublickeyAuthenticatorTest '-Dsurefire.failIfNoSpecifiedTests=false' test

### DIFF
--- a/sshd-core/src/test/java/org/apache/sshd/server/auth/AuthorizedKeyEntriesPublickeyAuthenticatorTest.java
+++ b/sshd-core/src/test/java/org/apache/sshd/server/auth/AuthorizedKeyEntriesPublickeyAuthenticatorTest.java
@@ -18,6 +18,7 @@
  */
 package org.apache.sshd.server.auth;
 
+import java.net.InetSocketAddress;
 import java.security.KeyPair;
 import java.security.KeyPairGenerator;
 import java.security.interfaces.ECPublicKey;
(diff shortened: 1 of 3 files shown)


sshd-core/src/test/java/org/apache/sshd/server/auth/AuthorizedKeyEntriesPublickeyAuthenticatorTest.java
```diff
@@ -68,4 +75,133 @@ void testMatch(String first, String second) throws Exception {
...
+        return auth.authenticate("user", pair.getPublic(), session);
+    }
+
+    private static KeyPair generateKeyPair() throws Exception {
+        KeyPairGenerator gen = KeyPairGenerator.getInstance("EC");
+        gen.initialize(256);
+        return gen.generateKeyPair();
+    }
+
+    private static ServerSession session(InetSocketAddress remote) {
```
tomaswolf, line 148: Use Mockito for this. Simply
```
    private static ServerSession session(InetSocketAddress remote) {
        ServerSession session = Mockito.mock(ServerSession.class);
        Mockito.when(session.getClientAddress()).thenReturn(remote);
        Mockito.when(session.getRemoteAddress()).thenReturn(remote);
        return session;
    }
```

tomaswolf (commented):
Thanks a lot for this contribution! Looks good; the test could be simplified.

sshd-core/src/test/java/org/apache/sshd/server/auth/AuthorizedKeyEntriesPublickeyAuthenticatorTest.java
tomaswolf: Done.

PR #912, opened 2026-08-11
Fix SOCKS5 dynamic forwarding of fragmented and pipelined requests

### LINKED ISSUE #911: `SocksProxy` mishandles fragmented and pipelined CONNECT requests and corrupts the reply
### Version

2.15.0

### Bug description

When a client connects through the dynamic SOCKS5 proxy started with
`ClientSession.startDynamicPortForwarding(...)`, two things go wrong:

1. If the greeting or the CONNECT request arrives split across several
   reads, the proxy throws `BufferException: Underflow` and closes the
   connection.

2. If the client sends data right after the CONNECT request, in the same
   write, the proxy sends those bytes back to the client inside the CONNECT
   reply. The data never reaches the destination.

This breaks any protocol that starts sending immediately after CONNECT. 

Cause, in `SocksProxy.Socks5`: each read is parsed as if it were one
complete SOCKS5 message (no bytes are kept between reads), and the CONNECT
request buffer is reused as the reply buffer.

### Actual behavior

- Fragmented greeting/CONNECT: proxy throws and closes the connection.

- Pipelined CONNECT+data: client receives a reply containing request and/or
  application bytes; the payload is not forwarded (or not forwarded exactly
  once) to the destination.

In other words, the proxy corrupts the data instead of just passing it.

### Expected behavior

- A message split across several reads is parsed correctly: the proxy waits
  until it is complete.
- Data sent after the CONNECT request reaches the destination, in order,
  exactly once.

The proxy should pass bytes through untouched for any TCP payload, including TLS/mTLS.

### Relevant log output

```Shell
org.apache.sshd.common.util.buffer.BufferException: Underflow: no available bytes in buffer
    at org.apache.sshd.common.util.buffer.ByteArrayBuffer.getByte(ByteArrayBuffer.java:209)
    at org.apache.sshd.common.util.buffer.Buffer.getUByte(Buffer.java:284)
    at org.apache.sshd.common.forward.SocksProxy$Proxy.getUByte(SocksProxy.java:117)
    at org.apache.sshd.common.forward.SocksProxy$Socks5.onMessage(SocksProxy.java:258)
    at org.apache.sshd.common.forward.SocksProxy.messageReceived(SocksProxy.java:85)
    at org.apache.sshd.common.io.nio2.Nio2Session.handleReadCycleCompletion(Nio2Session.java:409)
```

### Other information

I will try to drop a PR with the fix soon.

mucinoab:
The SOCKS5 proxy used for dynamic port forwarding parsed each read as a complete message, throwing a buffer underflow when the greeting or the CONNECT request arrived split across several reads. It also reused the CONNECT request buffer as the reply, echoing request bytes (and any application data pipelined after the request) back to the client instead of forwarding them.

Rewrite Socks5 as a state machine with a retained buffer: parse only once a message is complete and queue bytes received before the channel opens, forwarding them once it is up.

The state is synchronized because channel open and write futures complete on other threads.

Fixes #911

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/common/forward/SocksProxy.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/forward/SocksProxy.java
@@ -201,12 +201,20 @@ public class SocksProxy extends AbstractCloseable implements IoHandler {
         }
     }
 
+    private enum Socks5State {
+        GREETING,
+        CONNECT_REQUEST,
+        OPENING_CHANNEL,
+        FORWARDING,
+        CLOSED
+    }
+
     /**
      * @see <A HREF="https://en.wikipedia.org/wiki/SOCKS#SOCKS5">SOCKS5</A>
      */
     public class Socks5 extends Proxy {
-        private byte[] authMethods;
-        private Buffer response;
+        private Socks5State state = Socks5State.GREETING;
+        private Buffer pending = new ByteArrayBuffer();
 
         public Socks5(IoSession session) {
             super(session);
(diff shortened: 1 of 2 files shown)


mucinoab:
@tomaswolf could you please trigger a pipeline run again. Thank you.

PR #913, opened 2026-08-24
GH-656: shrink ChannelPipedInputStream buffer after all data is read

### LINKED ISSUE #656: How Do I Clear the Buffer After a Cache Session Is Used to Avoid Memory Usage?
Default implementation. Each cache session occupies 512 KB.

Image

Image

arimu1:
Fixes #656

## Summary
- When all piped data has been consumed, replace the internal `ByteArrayBuffer` with a fresh default-sized buffer instead of calling `compact()`, which only resets positions and leaves the large backing array allocated
- On `eof()` with no unread data, apply the same shrink so cached SSH sessions do not retain ~512 KB per channel after command completion

## Test plan
- [x] Added `bufferShrinksAfterAllDataRead` regression test (64 KiB payload; asserts backing array returns to `ByteArrayBuffer.DEFAULT_SIZE` after drain + EOF)
- [x] Existing `ChannelPipedInputStreamTest` cases pass
- [x] Command: `mvn -pl sshd-core -am -DskipTests install -q && mvn -pl sshd-core -Dtest=ChannelPipedInputStreamTest test` (Java 21, Maven 3.9.16)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelPipedInputStream.java
+++ b/sshd-core/src/main/java/org/apache/sshd/common/channel/ChannelPipedInputStream.java
@@ -153,7 +153,9 @@ public class ChannelPipedInputStream extends InputStream implements ChannelPiped
                 len = buffer.available();
             }
             buffer.getRawBytes(b, off, len);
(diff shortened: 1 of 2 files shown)


tomaswolf:
Thank you!

